PR #9, opened 2017-09-07
Fixes #8 Create a mod friendly way to manipulate loot tables

### LINKED ISSUE #8: Create a mod friendly way to manipulate loot tables
Although loot tables are already deprivatized there is no common way to add/remove loot.
There are some caveats involved,e.g. loot chances must be recalculated per rollgroup when a loot table is edited.
A simple helper class with static methods would go a long way in providing a unified interface for mods to interact with loot tables without the hassle of ini entry manipulating.

For pre start state manipulation of loot tables use the static methods in `OnPostTemplatesCreated` like this
`class'X2LootTableManager'.static.AddEntryStatic(LootTableName, LootTableEntry);`

At runtime you can use the singleton instance like

`LootManager = class'X2LootTableManager'.static.GetLootTableManager();`
`LootManager.AddEntry(LootTableName, LootTableEntry);`
`LootManager.InitLootTables();`

this will also reinitialise the loot tables.

For bulk add to loot tables you can move the chance recalculation to the end:

`LootManager.AddEntry(TableName, EntryToAdd, false);`
`LootManager.AddEntry(TableName, EntryToAdd, false);`
`LootManager.AddEntry(TableName, EntryToAdd, false);`
`LootManager.RecalculateLootTableChance(TableName);`

This way your chance ratio is preserved.

Musashi1584:


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2LootTable.uc
```diff
@@ -0,0 +1,248 @@
...
+//             requires re-invoking InitLootTables again
+var config array<LootTable> LootTables; 
+var private native Map_Mirror       LootTablesMap{TMap<FName, INT>};        //  maps table name to index into LootTables array
+
+// bValidateItemNames when false, expected to be used for character template names.
+native function InitLootTables(bool bValidateItemNames=true);		//  validates loot tables and sets up LootTablesMap
+native function RollForLootTable(const out name LootTableName, out array<name> RolledLoot);
+
+// Start Issue #8 - Add a loot table interface
+public function AddLootTable(LootTable AddLootTable)
```
MalucoMarinero: Not keen on variable name AddLootTable being the same as the function name. Would `LootTableToAdd` be better?

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2LootTable.uc
```diff
@@ -0,0 +1,248 @@
...
+{
+	AddLootTableIntern(self, AddLootTable);
+}
+
+public function RemoveLootTable(LootTable LootTable)
```
MalucoMarinero: Variable being same name as the type doesn't seem like it should've compiled, change `LootTable` to `LootTableToRemove` or some such.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2LootTable.uc
```diff
@@ -0,0 +1,248 @@
...
+
+// **************************************************
+// Private function
+// **************************************************
+private static function X2LootTable GetX2LootTableCDO()
+{
+	return X2LootTable(class'Engine'.static.FindClassDefaultObject(string(default.Class.Name))); 
+}
+
+private static function AddLootTableIntern(X2LootTable LootTable, LootTable AddLootTable)
```
MalucoMarinero: Does `Intern` actually mean `Internal`? Should probably do that for clarity.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2LootTable.uc
Musashi1584: Actually that is valid and compiles, but i will change it to prevent confusion.

Issue #18, opened 2017-09-13
Modifying default spawn size


bountygiver:
Allow the modification of default spawn size to support squads larger than 9 units

PR #19, opened 2017-09-13
Larger spawn

bountygiver:
Issue #18

### DIFF
--- a/X2WOTCCommunityHighlander/Config/XComGame.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGame.ini
@@ -4,5 +4,5 @@ DLCIdentifier="X2WOTCCommunityHighlander"
 [X2WOTCCommunityHighlander.X2WOTCCH_UIScreenListener_ShellSplash]
 bEnableVersionDisplay=true
 
-[XComGame.XComGroupSpawn]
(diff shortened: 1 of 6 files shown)


X2WOTCCommunityHighlander/Config/XComGame.ini
```diff
@@ -3,3 +3,6 @@ DLCIdentifier="X2WOTCCommunityHighlander"
 
 [X2WOTCCommunityHighlander.X2WOTCCH_UIScreenListener_ShellSplash]
 bEnableVersionDisplay=true
+
+[XComGame.CHHelpers]
+SPAWN_EXTRA_TILE=0
```
robojumper: Remove or comment out so that a single other mod using this doesn't run into load order issues?

bountygiver: so you think it's better to let other mods modify this value rather than put it in one place for the user to modify it?

robojumper: Why not both? Comment it out and add a comment so that the user knows that they *may* edit this value.

MalucoMarinero:
This seems alright to merge, a few notes for tidy up but otherwise good.

Issue #6, opened 2017-09-07
Allow to use the Reaper UI without being super concealed.


Musashi1584:
By adding two events in `UITacticalHUD`, mods should able to use the Reaper UI without using the "SuperConcealed" mechanic,

PR #22, opened 2017-09-14
#6 Enhances Reaper UI Event to make the Shot HUD overridable

Musashi1584:


MalucoMarinero:
Slight addition to Issue #6, is backwards compatible as it only adds an item to the tuple.

Issue #21, opened 2017-09-13
Request for function allowing mods to add sockets from their DLCInfo


shiremct:
Something like the DLCAppendSockets functions in XComUnitPawn and XComHumanPawn from LW2/X2CommunityHighlander.

Requested in order to bring the models/animations/sockets of the secondary weapons added in LW2 over to WOTC. Thanks!

PR #28, opened 2017-09-16
Issue #21 - DLCAppendSockets

shiremct:


MalucoMarinero:
this looks alright at first glance @shiremct - have you run this without issue? just checking cause I'd rather know I'm starting from a stable state before trying to merge it into the release

shiremct:
Yea, I ran it with the LW2 secondaries and they showed up without issue. Right now it's almost a direct copy of these functions from the previous highlander. Thanks!

Issue #41, opened 2017-09-24
Ragdoll Collision is Disabled Non-Optionally 


PixalVortex:
[Code Changes to XComUnit_Pawn.txt](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/files/1327017/Code.Changes.to.XComUnit_Pawn.txt)

The Configs in there current form must be implemented in a helper class! 

As an added bonus, the code change also allows Chosen death animations to be interrupted optionally

PR #57, opened 2017-10-04
Issue #41 - Allow configuration of Ragdoll behaviour to be always on,

MalucoMarinero:
and also enable ragdolls for the chosen.


MalucoMarinero:
Would like a review of any performance concerns here, but seems pretty straightforward.

PixV:
I have been testing the performance with the Ragdoll collision on for over a week and a half, the performance isn’t really impacted except by massive amounts of bodies such as Lost missions

Issue #53, opened 2017-09-30
Max Tactical Autosaves


ADVENTAvenger:
@Maluco I'd like to request a very, very easy unprotecting/configurability of a variable.

class XComAutosaveMgr extends Actor	dependson(XComOnlineEventMgr);

const MAX_TACTICAL_AUTOSAVES = 3;

There's really no reason to limit the game to 3 saves (they are so small), and since sometimes saves get corrupted it would be nice to be able to rollback to a previous one that works during a tactical mission.

Thank you.

PR #58, opened 2017-10-04
Issue #53 - make tactical autosave count configurable

MalucoMarinero:


robojumper:
Hmmm. Rewriting a `const` to a `var const int` is questionable, it makes changes to the data layout. In this case, it probably doesn't cause any issues (no cooked class, no native class), and is unlikely to break mods that use it (because `const` literals are compiled into the script package).

MalucoMarinero:
@robojumper can `const config int` work? haven't even tried it. Alternatively if all result in a data layout change and we want to avoid it, we'd just have to rewrite all usages to use a variable from CHHelpers. Which isnt that hard.

PR #60, opened 2017-10-04
Fixes #55 UIArmory_Loadout should evaluate …

### LINKED ISSUE #55: UIArmory_Loadout should evaluate X2SoldierClassTemplate.bNoSecondaryWeapon.
Currently `UIArmory_Loadout ` tests `UpdatedUnit.GetRank() > 0` to check if it should display the secondary weapon slot. The right way should be to test against `UpdatedUnit.NeedsSecondaryWeapon()` which evaluates `X2SoldierClassTemplate.bNoSecondaryWeapon`.

Musashi1584:
X2SoldierClassTemplate.bNoSecondaryWeapon


MalucoMarinero:
@robojumper would changing this check have any downsides/bugs we should look out for? -- possibly affecting resistance soldier bonuses on mission?

robojumper:
By the header comment, UIArmory_Loadout only handles soldiers. I can't think of a circumstance where there are non-soldiers that should have a secondary weapon that can be selected using this screen.

Issue #30, opened 2017-09-17
Ensure that soldiers doesn't contain the abilities that can be added by training center


Brood:
Currently, the `X2SoldierClassTemplateManager.GetCrossClassAbilities(optional X2SoldierClassTemplate ExcludeClass)` doesn't check against abilities that the soldier class itself has.

This can happen for example when bAllowAWCAbilities is set to true for sparks, this allows both sparks and grenadiers the possibility to get "Shredder" as a trainable perk.

PR #63, opened 2017-10-06
Allow abilities to be added to the "XCOM" pool when they exist in the soldiers randomized ability pool

Brood:
So this fix enhanced what I did in #30, the problem with vanilla / #30 is that it adds all the perks not "used" by the soldier class to the "XCOM" pool. Where "used" in this case means all the potential perks the soldier could use, not the actual perks that are in the soldiers ability tree.
This means that for solder classes that use random decks, all the perks in a random deck are "reserved" and will not be added to the "XCOM" pool. 

This fix changes the behavior to allow all perks not currently in the tree to be added to the "XCOM" pool.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2SoldierClassTemplateManager.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2SoldierClassTemplateManager.uc
@@ -76,7 +76,7 @@ function array<SoldierClassAbilityType> GetCrossClassAbilities(optional X2Soldie
 
 	foreach arrClassTemplates(ClassTemplate)
 	{
-		if(ClassTemplate.DataName != DefaultSoldierClass && ClassTemplate != ExcludeClass && ClassTemplate.bAllowAWCAbilities)
+		if(ClassTemplate.DataName != DefaultSoldierClass && ClassTemplate.bAllowAWCAbilities)
 		{
 			AbilityTree = ClassTemplate.GetAllPossibleAbilities();
 			for(idx = 0; idx < AbilityTree.Length; idx++)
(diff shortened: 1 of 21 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2SoldierClassTemplateManager.uc
```diff
@@ -76,7 +76,7 @@ function array<SoldierClassAbilityType> GetCrossClassAbilities(optional X2Soldie
 
 	foreach arrClassTemplates(ClassTemplate)
 	{
-		if(ClassTemplate.DataName != DefaultSoldierClass && ClassTemplate != ExcludeClass && ClassTemplate.bAllowAWCAbilities)
+		if(ClassTemplate.DataName != DefaultSoldierClass && ClassTemplate.bAllowAWCAbilities)
```
robojumper: When someone calls GetCrossClassAbilities with an ExcludeClass, they should not get abilities from this class.

Brood: They don't since the function now takes a non optional array <SoldierRankAbilities> SoldierAbilityTree, and later on in the code we only add to the soldier if we don't find it in this list. ~10 rows down, start of #30.

MalucoMarinero:
So I'm just looking this over, might pass over it for this release (I'm hoping to get these faster out soon anyway), just hoping to get straight on it.

Essentially, GetCrossClassAbilities is being kept only as a compatibility aid for mods that target 'GetCrossClassAbilities' directly right? Otherwise, it's meant to be passed over entirely for vanilla code?

Also feels like GetCrossClassAbilities_CH should get some comments clearly stating it's approach as opposed to how the original version does it, but maybe I'm just tired and it's actually pretty self explanatory if I read it properly.

Brood:
I could definitely add comments to the code, didn't know that we should/was
allowed to do it.
But yes, GetCrossClassAbilities is a one use only, and I created a new
variant since the signature changed too much.

Now that I think about it, we should remove the old variant from the CH
since it doesn't do anything (and we therefore doesn't need to overwrite
it).



/Jonas

2017-10-18 12:29 GMT+02:00 MalucoMarinero <notifications@github.com>:

MalucoMarinero:
Would any mod have a reason for calling the original CrossClassAbilities function? Because if they did, and we changed the signature we'd guarantee a crash, meaning vanilla mods being incompatible with the highlander, an undesirable situation in my mind, we should always aim for compat with code targeting vanilla.

Brood:
Sorry maybe I wasn't clear :)

What I propose:

   1. Remove the previous fix for the CrossClassAbilities function, i.e.
   delete it from the highlander
   2. Add some comments into why the _CH version is needed and change the
   use in the calling function to the _CH

No mod should have any reason to call the original, but with this we are
safe.



/Jonas

2017-10-19 11:29 GMT+02:00 MalucoMarinero <notifications@github.com>:

MalucoMarinero:
@robojumper @Brood kinda want to close this one out, it's been running on for a while and I haven't got around to properly merging. Feels like all thats left in terms of old compatibility is to just ensure the old GetCrossClassAbilities function is there, operating exactly as it does in vanilla. It won't be used by the highlander code, but at least it's there just in case a mod calls it.

Brood:
I thought it was done and ready.

MalucoMarinero:
Actually reading it over I think we're good, just no response to your response to robojumper. All good, will merge.

Issue #64, opened 2017-10-07
Streamline code for number of utility item slots and add DLC Hook.


robojumper:
In the vanilla Highlander, there was a function `GetNumUtilitySlots()` (see X2CommunityCore/X2CommunityHighlander#114 ). It was called wherever some code would change / reset the number, and expected an override number. This made a lot of code more ugly than it already is.

In WotC, it gets worse. Consider this code that handles unequipping a piece of armor:

	case eInvSlot_Armor:
		if(!IsMPCharacter() && X2ArmorTemplate(Item.GetMyTemplate()).bAddsUtilitySlot)
		{
			if (!HasExtraUtilitySlotFromAbility())      //  don't lower the bonus if one is given via an ability
			{
				SetBaseMaxStat(eStat_UtilityItems, GetMyTemplate().GetCharacterBaseStat(eStat_UtilityItems));
				SetCurrentStat(eStat_UtilityItems, GetMyTemplate().GetCharacterBaseStat(eStat_UtilityItems));
			}
		}
		break;

This is bad. All code that handles the utility item stat needs to be aware of any other systems that touch the slot number.

I would use a slightly different approach: A function called `RealizeNumUtilitySlots()` that can be called whenever anything happens that would modify utility slot numbers, and an additional DLC hook that allows mods to modify / override this number. This function would always run all the checks: Character Base Stat, Bonus Armor Slot, Tactical Rigging, DLC Hooks, and then set the character stat. Sensible places to do this are equipping / unequipping items, buying abilities.

This function should then be the only code that sets this stat (except TQL / dev tools, cheats and MP).

Danice123:
I think LWOTC is gonna need this, I've already hit code dealing with this. I may try something.

bountygiver:
Closed with #171

PR #65, opened 2017-10-07
Streamline code for number of utility item slots and add DLC Hook

robojumper:
This addresses #64


robojumper:
Not yet tested, please don't merge yet

Issue #76, opened 2017-10-11
UIStrategyMap.uc does not show custom faction HQ icons


RealityMachina:
The specific offender is the function `UpdateMissions()`, as the if check for faction HQs assumes that any haven with a faction HQ attached it will be one of the vanilla three factions to init, as so:

``` 
               FactionHQ = XComGameState_Haven(ScanSite);
                if(FactionHQ != none)
                {
                    numHQ++;
                    switch (FactionHQ.GetResistanceFaction().GetChampionClassName())
                    {
                    case 'Skirmisher':
                        MissionItemUI.MissionIcons[1].SetScanSite(ScanSite);
                        break;
                    case 'Reaper':
                        MissionItemUI.MissionIcons[0].SetScanSite(ScanSite);
                        break;
                    case 'Templar':
                        MissionItemUI.MissionIcons[2].SetScanSite(ScanSite);
                        break;
                    }
                    
                }
```

This should be relatively easy to fix by altering this function to account for the possibility of mod-created factions, both for the actual initialization of the icon and for the NumHQ check so it doesn't think the center three icons are filled when they are not.

PR #77, opened 2017-10-11
Fix issue #76 via adding default fallthrough check for mod-added factions. Fix Issue #82 via filtering out custom faction templates. Fix Issue #85 via adding checks for positive traits and restoring a commented out function.

RealityMachina:


X2WOTCCommunityHighlander/Src/XComGame/Classes/UIStrategyMap.uc
```diff
@@ -0,0 +1,1953 @@
...
+{
+	local array<XComGameState_ScanningSite> arrScanSites;
+	local XComGameState_ScanningSite ScanSite;
+	local XComGameState_MissionSite MissionSite;
+	local XComGameState_Haven FactionHQ;
+	local int i, numScanSites, numMissions, numHQ;
+	local XComGameStateHistory History;
+	local bool bGuerillaAdded;
+	//issue #76, added variable for proper accounting of faction HQs
+	local bool VanillaHQ;
```
robojumper: Doesn't seem to be used

RealityMachina: Editing a script at 11PM: almost as bad as editing a script at midnight. :V

(On the plus side, midnight me probably would've went with the original workaround I had planned before I realized i could just move the NumHQ++ to being used in the vanilla HQ cases)

Anyway, fixed.

MalucoMarinero:
How are we looking with this one? I'll pass it over for the release tonight but I guess it's not too far off yah?

RealityMachina:
Should be all good now.

MalucoMarinero:
@robojumper regarding config, arrays like this should be fine yes? later mods can neg out these factions safely. It's a bit different compared to other config values where the default is just an unset variable.

robojumper:
@MalucoMarinero There's always the possibility to make it an exclusion list instead of an inclusion list, which would not require us to add anything to it.

I would argue that an exclusion list may even make more sense, because this patch requires all factions that are *not* added to that list to be manually set up (in InstallNewCampaign) if I understand it right. @RealityMachina is free to correct me on that if I'm wrong.

RealityMachina:
Yeah, my logic is that since the game is currently not really set up to handle more than three base game factions when assigning rival Chosen (especially since the Chosen missions as-is will fail on custom chosen), the presumption is that either the mod adding the custom faction as-is will have special handling for itself, or will be part of an overhaul that that just changes all that anyway.

RealityMachina:
Is there any more concerns?

Beyond the "learn to git" one, I know about that one. :v

MalucoMarinero:
@robojumper any objections to me merging. I'll do a little cleanup and documentation when I do but I can take it the last mile.

robojumper:
I would prefer a config array for factions that are excluded from being set up at the beginning of the campaign, since (as of now), this change would suddenly stop setting up mod-added factions at campaign start.

It makes more sense that a mod author adds their faction to the config array and then sets up their faction manually in InstallNewCampaign, because this retains behavior for mods that don't use the Highlander.

RealityMachina:
Alright, changed.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAvengerHud_SoldierStatusIndicatorContainer.uc
```diff
@@ -0,0 +1,306 @@
...
+					}
+				}
+			}
+			else if(Unit.AlertTraits.Length > 0)
+			{
+			//start issue #85: init variables here to check for trait template
+			EventTemplateManager = class'X2EventListenerTemplateManager'.static.GetEventListenerTemplateManager();
+			TraitTemplate = X2TraitTemplate(EventTemplateManager.FindEventListenerTemplate(AlertTraits[0])); //while it's an array, we go with the base game's assumption of having only one trait to worry about when it changes				
+				
+			if(!TraitTemplate.bPositiveTrait)
```
robojumper: Why not `AddStatusIndicator(Unit, CreateTraitIcon(Unit, TraitTemplate.bPositiveTrait));`

Issue #80, opened 2017-10-17
Prevent certain classes from rolling AWC abillities


Musashi1584:
I have two classes which do not have AWC abilities. The Samurai and the SpecOps class.

The samurai wouldnt even benefit from 90% of the AWC abilities because his primary weapon is the sword.

Either way i want these classes to behave like normal classes and don't want them to be Hero classes or get AWC abilities. I just just want to allow them to participate in the Combat Intelligence / Ability Points System.

This can be achieved by a config array of class names that prevent certain classes from rolling awc abilities in `XComGameState_Unit.RollForTrainingCenterAbilities()` as @robojumper suggested.

This way you could also give Sparks access to the Combat Intelligence / Ability Points System without having them AWC abilities as an example.

Musashi1584:
Example usage:

In XComGame.ini 
```
[XComGame.CHHelpers]
+ClassesExcludedFromAWCRoll=ExampleClassName
```

PR #81, opened 2017-10-17
Solves Issue #80

Musashi1584:
Let certian soldier classes allow to get AP regardless of `bAllowAWCAbilities`

These classes are defined in the config array `class'CHHelpers'.default.AlwaysAllowAbilityPointsClasses`.


MalucoMarinero:
I feel like I have to push back on this one or at least get more of an explanation for why it exists. Currently it seems like you could do what it does simply by modifying templates OnPostTemplatesCreated to get the job done.

What situation arises where a template with bAllowAWCAbilities is false - yet the name of that template is on a separated list of inclusions - is desirable? It doesn't seem like it serves a unique purpose, at least from where I'm looking at it.

What's the specific scenario it's looking to fix?

robojumper:
bAllowAWCAbilities controls whether:

* This soldier gets an additional row with random abilities 
* This soldier can purchase Abilities with ability points 
* This soldier gains Ability Points on rank up 

And related. In general, `IsResistanceHero() || bAllowAWCAbilities` is equivalent to *this soldier participates in the combat-intelligence-abilitiy-points system*.

What @Musashi1584 is suggesting is adding more fine control to that system. I'm also not necessarily happy with how it's implemented, I'd at least add a few accessor functions, and maybe approach it the other way around (so bAllowAWCAbilities still means "participates in the ComIng points system).

Musashi1584:
I have two classes which do not have AWC abilities. The Samurai and the SpecOps classes.
The samurai wouldnt even benefit from 90% of the AWC abilities because his primary weapon is the sword.
Either way i want these classes to behave like normal classes and don't want them to be Hero classes just to allow them to get and spend Ability Points.

PR #84, opened 2017-10-18
Fixes Issue #80 Prevent certain classes from rolling AWC abillities

Musashi1584:


MalucoMarinero:
Yeah, I like this way better in terms of commenting and code style, much clearer what the role of the config is and it's a very light touch in terms of added code.

Issue #89, opened 2017-10-21
Poison Bypasses Shields


ADVENTAvenger:
I'd like to add 1 line to the CreatePoisonedStatusEffect function, Poison Damage Over Time to bypass Energy Shields.

	DamageEffect = new class'X2Effect_ApplyWeaponDamage';
	DamageEffect.EffectDamageValue.Damage = default.POISONED_DAMAGE;
	DamageEffect.EffectDamageValue.DamageType = 'Poison';
	DamageEffect.bIgnoreBaseDamage = true;
	DamageEffect.DamageTypes.AddItem('Poison');
	DamageEffect.bAllowFreeKill = false;
	DamageEffect.bIgnoreArmor = true;
        DamageEffect.bBypassShields = true; <=======
	PersistentStatChangeEffect.ApplyOnTick.AddItem(DamageEffect);

I classify this as a bugfix.

PR #94, opened 2017-10-24
Issue #89 - Tick damage ignores shields

TNychka:
See issue #89. Acid/Burning/Poison now ignore shields. Currently, this is assumed to be a bug fix: It may be intentionally designed behavior.

Second question, when submitting this pull request I had to manually change the EOL to unix style from windows. Is there something configurable in modbuddy so that I don't need to do this? Thanks.


robojumper:
It's a gameplay change, so I would want it guarded behind a config option.

TNychka:
Can do

TNychka:
Do not merge this pull request, still doing some testing

TNychka:
Added config variables that default to false. Tested in game as functional. @robojumper PTAL

robojumper:
Can you comment out the entries in `XComGameCore.ini`? This way, mods that enable this behavior don't need to care about config load order.

TNychka:
Fixed it!

Issue #114, opened 2017-11-28
Pass in optional itemstate to CanAddItemToInventory hooks


RealityMachina:
This should be doable without breaking any mods reliant on it by adding `optional XComGameState_Item Item` to the end of the relevant functions in `X2DownloadableContentInfo`, and having`XComGameState_Unit` and `UIArmory_Loadout` pass along their item states.

Require this for my Item Repair mod to function at 100%. (Since I am looking for something not available in the item's template, but is contained in the item state itself.)

EDIT:
Scratch editing XComGameState_Unit: it has the same issue with how it does CanAddItemToInventory, and editing what a vanilla function takes in can be...problematic.

That said this is still solvable for me: the nature of how the vanilla game works means it doesn't automatically use non-infinite items: I only need to affect the hook that prevents the player from equipping the item (the one that can give out a disabled reason), not the one used by the game in various places automatically.

EDIT2:

Robojumper has informed me that certain functions can be altered even if the class uses native code in some capacity; `CanAddItemToInventory` is one of these functions.

Iridar:
The recent discovery of a rather major bug caused by misuse of this DLC hook (Issue #1056) makes it crystal clear that this hook desperately needs proper docs. I'm reopening the issue so that I don't forget to handle this.

Xymanek:
Documentation does not require an updated build, hence there is no need to assign such cases to milestones

PR #115, opened 2017-11-28
Edited X2DLCInfo hook to accommodate ItemState for Issue #114 and Issue #127

RealityMachina:
Lets mods look at the gamestate of the Item when deciding whether to disable it from being equipped by the player. For Issue #114


robojumper:
Have you tested this? What if a mod's subclass implements the function without the parameter?

RealityMachina:
...

Right, that completely slipped my mind.

Ok, on to take two: make a wrapper function for the wrapper function. :v

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -273,14 +273,26 @@ static function bool GetValidFloorSpawnLocations(out array<Vector> FloorPoints,
...
+/// start Issue #114: added XComGameState_Item as something that can be passed down for disabled reason purposes
+/// basically the inventory hook wtih an added paramter to pass through
+/// we leave the old one alone for compatibility reasons, as we call it through here for those mods.
+///
+static function bool CanAddItemToInventory_CH_Improved(out int bCanAddItem, const EInventorySlot Slot, const X2ItemTemplate ItemTemplate, int Quantity, XComGameState_Unit UnitState, optional XComGameState CheckGameState, optional out string DisabledReason, optional XComGameState_Item ItemState)
+{
+	if(CheckGameState == none)
+		return CanAddItemToInventory_CH(bCanAddItem, Slot, ItemTemplate, Quantity, UnitState, CheckGameState, DisabledReason) //this can be handled by this function by newer mods
+
+	return CanAddItemToInventory(bCanAddItem, Slot, ItemTemplate, Quantity, UnitState, CheckGameState);
```
robojumper: Any reason to add the check and both functions here? If `CheckGameState == none`, `CanAddItemToInventory_CH` defaults to false, and if it's not none, it calls `CanAddItemToInventory` anyway, so you should be fine replacing the function body with a call to `CanAddItemToInventory_CH`

MalucoMarinero:
@RealityMachina just confirming this is good to go before I merge it in? Would like to cut a release in the next couple days.

RealityMachina:
Yeah, it should be good to go.

Issue #128, opened 2017-12-06
Allow to define mutually exclusive abilities


Musashi1584:
Allow mutually exclusive abilities by adding 

`Template.PrerequisiteAbilities.AddItem('NOT_SomeAbility');`

to an ability template.

This will be evaluated in XCGS_Unit.MeetsAbilityPrerequisites and the Hero Promotion Screen.

PR #129, opened 2017-12-06
Issue #128 Allow to define mutually exclusive abilities 

Musashi1584:
by 'NOT_' prefix in PrerequisiteAbilities


MalucoMarinero:
@Musashi1584 just to be clear on its purpose, the idea is the skills will be mutually exclusive even within the Training Centre, where players can pursue every tree, right?

PR #130, opened 2017-12-07
Update CHHelpers with Strip Config Vars

john-luke-pack-hard:
Added config vars from PI's highlander to configure stripping behavior on soldiers.


Musashi1584:
Shouldnt this included the whole extended strip functionality? Only the config vars won't bring any benefit.

john-luke-pack-hard:
Give me some time to do a full investigation. I'll update it with the full functionality.

PR #132, opened 2017-12-10
Add UISquadSelect.uc

john-luke-pack-hard:
Contains the equipment strip config vars and all of its co-requisite behavior.


john-luke-pack-hard:
Don't add it yet, I forgot to do a diff check to see where WotC changed the code. The file is straight from the LW2 source code.

Issue #163, opened 2018-01-27
Add hooks to support new menus in the Avenger


Danice123:
Specifically:
LW2 Squad management menu

PR #164, opened 2018-01-27
Issue #163 Avenger Menu Hook

Danice123:
Not a lot of changes for this, other than the new file. I'm hoping to test this in a bit.


MalucoMarinero:
Will leave this PR open until you've tested it, rather than merge into the mainline (as I won't be able to build or deploy over the next two weeks, don't want to risk untested changes)

Danice123:
Thats fine, I have tested it, but with unstable code. I'll confim when I get some time

Issue #159, opened 2018-01-24
Add hook to IsUnitValidForPsiChamberSoldierSlot in X2StrategyElement_DefaultStaffSlots to allow mod interaction with psi operative training


RealityMachina:
Just so multiple mods can interact with it without needing to override it for the same reason of "I need to add my own check here for my own purposes."

Format should ideally look something like:

```
if(Unit.IsSoldier() && Unit.IsActive() && Unit.CanBeStaffed() //original basic check
{
     ModHookHere(bool_override, bool_allowed); //event trigger or DLC hook

     if(bool_override) //if a mod says it wants to override the result, the hook stops and returns that mod's results
         return bool_allowed; //and we go with whatever it says

        //original psi check of "is it a rookie" and "is it a psi operative with abilities remaining" can continue from here if there's been no override of the result
}

```

PR #176, opened 2018-02-03
Pull request for issue #159

RealityMachina:


MalucoMarinero:
Just looking at it, this hook is possibly better as an event? ala: https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/173/files

That way you can override to true OR false, rather than only true using 1 or above as the return.

Issue #171, opened 2018-02-03
Allow more customization to armory loadout screen for mods.


bountygiver:
This is to roll in features from TONS of Utility Items to community highlander that allows wider customization of armory loadout screen for mods. The main difference of the highlander version and the standalone mods will be highlander will be configured through X2DLCInfo instead of an ini config to allow an even wider customization.

Changes already implemented in highlander:
- Custom restriction on specific items

Changes that needs to be implemented:
- Customization of number of utility item slots (#64 )
- Overriding images of items conditionally
- Overriding whether items should be equipped automatically for base game item slots (particularly 1st utility item and grenade slots, probably still use ini for this)
- Allow multiple heavy weapons
- Handling of ammo pockets in ValidateLoadout

Danice123:
+1 <3

PR #177, opened 2018-02-04
Added changes for Issue #171

bountygiver:
Added enhancements to increase customization of armory loadout behaviour. 

- Added hooks to change maximum number of utility items and heavy weapons
- Changed `CHItemSlot.SlotGetUnequipBehavior` to accept empty ItemState, returning `eCHSUB_AttemptReEquip` as default, overwritable by mods. `XCOmGameState_Uni.ValidateLoadout` will also now respect the return value of this function for heavy weapons, utility items and grenade slots.
- Added hooks to override image icons for specific items on specific slots (List of image paths can be manipulated directly, allowing you to add a layer of icon on items improved by a perk).
- Handle Ammo Slot properly, removing incompatible ammos from ammo slots as if they are in utility slots, also forcing only 1 ammo in either ammo slot or utility slot, not both.


bountygiver:
the code duplication is because i forgot native class just cannot change the variables, it's still good to add more functions. I can fix that. For multi heavy weapon slots, I have been using a hacky way for my mod and it seems to interact well with the rest of the game, as long as other perks don't use it as source weapon but heavy weapons are optional items anyways. One thing with equipping on pawn is that for now all the heavy weapons are the same for now, but we can tackle that problem separately by maybe keeping track of identical archetypes and only render different ones?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -13847,9 +13846,9 @@ function String GetSoldierClassSummary()
 
 // Start Issue #171
 // Sets the eStat_UtilityItems of the unit and returns the number of heavy weapons
-function int RealizeItemSlotsCount(out int NumUtility, out int NumHeavy, XComGameState_Item ArmorItem, XComGameState CheckGameState, optional bool bSetStat=true)
+function int RealizeItemSlotsCount(XComGameState_Item ArmorItem, XComGameState CheckGameState, optional bool bSetStat=true)
```
robojumper: Comment is wrong.  
Is there any reason bSetStat would ever be false?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -10528,13 +10566,33 @@ function ValidateLoadout(XComGameState NewGameState)
...
+		}
+	}
 	EquippedUtilityItems = GetAllItemsInSlot(eInvSlot_Utility, NewGameState, ,true);
 	for(idx = 0; idx < EquippedUtilityItems.Length; idx++)
 	{
-		if(X2AmmoTemplate(EquippedUtilityItems[idx].GetMyTemplate()) != none && 
-		   !X2AmmoTemplate(EquippedUtilityItems[idx].GetMyTemplate()).IsWeaponValidForAmmo(X2WeaponTemplate(EquippedPrimaryWeapon.GetMyTemplate())))
+		if (X2AmmoTemplate(EquippedUtilityItems[idx].GetMyTemplate()) != none && 
+		   (!X2AmmoTemplate(EquippedUtilityItems[idx].GetMyTemplate()).IsWeaponValidForAmmo(X2WeaponTemplate(EquippedPrimaryWeapon.GetMyTemplate())) ||
+		   AmmoPocketFilled))
```
robojumper: This is a behavioral change from the base game, no? A better way to realize the behavior of "only one ammo allowed" would be to modify the unique rule depending on some config flag.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
bountygiver: ArmorItem is just there to provide convenience. I suppose bSetStat will never be false if you want people to use only eStat_UtilityItem as getter

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
bountygiver: so do you want the ammo to check for unique rule and make changes in UIArmory_Loadout instead?

robojumper: UIArmory_Loadout already handles the Unique rule correctly, the only change would be to make https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/47a00149c83d86b9826da64866313192cb41b18c/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc#L1120-L1122 configurable.

bountygiver: it is one sided though, like it will restrict utility slot when ammo slot is occupied but not the other way round. https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIArmory_Loadout.uc#L820

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
robojumper: I would take the ArmorItem out because it's not immediately clear what it is and how to use it. Case in point, you are calling it from here https://github.com/bountygiver/X2WOTCCommunityHighlander/blob/1aff6aa5afea7139eaeafbb0755a8da8f881307d/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc#L7512-L7516 with the armor item you just removed. Make the function more robust by removing both of those parameters.

Side note, X2Ability_DefaultAbilitySet::TacticalRiggingPurchased also sets the stat, consider changing that to call this function as well.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -10528,13 +10537,32 @@ function ValidateLoadout(XComGameState NewGameState)
...
+		{
+			EquippedAmmo = XComGameState_Item(NewGameState.ModifyStateObject(class'XComGameState_Item', EquippedAmmo.ObjectID));
+			RemoveItemFromInventory(EquippedAmmo, NewGameState);
+			XComHQ.PutItemInInventory(NewGameState, EquippedAmmo);
+			EquippedAmmo = none;
+		}
+		else
+		{
+			AmmoPocketFilled = true;
+		}
```
robojumper: This bit isn't needed anymore, right?

robojumper:
`X2Ability_DefaultAbilitySet::TacticalRiggingPurchased` still directly sets the stat. Won't be a huge problem but better be thorough.

bountygiver:
`TacticalRiggingPurchased` will only be a problem if something else checks for utility stats when it is not in loadout armory, and are we just going to fix `HotLoadAmmo` with this PR? The only drawback of merging that fix into this PR is it become slightly harder to find the fix being documented but it will be easier for me to push them.

robojumper:
Yeah, the HotLoadAmmo fix has its place here. It's code that every mod using the ammo slot would need to include otherwise.

bountygiver:
Come to think of it, should I make it iterate through whole inventory instead of just ammo pocket and utility when looking for ammo? For purposes of reading custom slots.

robojumper:
Yeah, the whole inventory would be considered.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -10646,11 +10680,20 @@ function ValidateLoadout(XComGameState NewGameState)
...
+	NumUtility = GetCurrentStat(eStat_UtilityItems);
+	for (idx = 0; idx < NumUtility; idx++)
 	{
-		UtilityItem = GetBestUtilityItem(NewGameState);
-		AddItemToInventory(UtilityItem, eInvSlot_Utility, NewGameState);
-		EquippedUtilityItems.AddItem(UtilityItem);
+		if (idx >= EquippedUtilityItems.Length && (idx < NumMinEquip || NumMinEquip == -1))
+		{
+			UtilityItem = GetBestUtilityItem(NewGameState);
+			if (AddItemToInventory(UtilityItem, eInvSlot_Utility, NewGameState))
```
robojumper: This is a bit dubious, but I don't know how to make it better: Since `GetBestUtilityItem` is randomized, we often end up hitting the unique rule, preventing us from equipping a specific item, leaving not all slots filled. Obviously, there may not be enough distinct categories to fill all slots, but if `NumMinEquip` is set to `2`, for example, we may roll two grenades, leaving only one slot filled.

Issue #183, opened 2018-02-10
Add in event hook to HasBeenModified() in XComGameState_Item


RealityMachina:
Due to the problematic issues that can arise from mods attaching gamestate components to vanilla gamestates, an alternate way for mods to signal that an item shouldn't be stacked is required.

It should check for a mod override before it defaults to the behaviour of returning whether the item has a nickname, been modified, or has a component.

Issue #160, opened 2018-01-26
Let part pack sliders in UIOptionsPCScreen be scrollable


RealityMachina:
Currently, the game just renders the first 11 valid part packs it can find with the recent highlander additions.

Fine for relatively small time mod users, but depending on mod pack (some may include "multiple" DLC packs) and # of mods installed, this may be very insufficient to fully customizing one's experience.

Hopefully one of the people actually competent with UI related things will be able to fix this for the rest of us. :v

MalucoMarinero:
UIList actually has scrolling behaviour built in so it may very well be as easy as updating line number 211 
```
const NUM_LISTITEMS = 16;
```
to something higher. That's what they use to limit the number of possible options on a given tab to 16. Whether or not something in the flash disables the scrolling remains to be seen but it could work out pretty simply.

PR #184, opened 2018-02-10
Pull request for issues #183 and #160

RealityMachina:
I will clean up the commit history after this one, honest. :v


robojumper:
Change for #160 does not work correctly: By default, some items are already scrolled out, and the scrollbar is shown and scrollable even when not needed.

grafik

Issue #188, opened 2018-02-16
Convert the MP teams into usable teams for modders


RealityMachina:
So, one of the fun things Firaxis never fixed in War of the Chosen is the silliness of how teams are implemented in XCOM 2, specifically as enums used as bitfields in the native end of the game:

```
// FIRAXIS begin -tsmith
// NOTE: need the eTeam_DONT_USE_X because the enum values are used as bitfields in C++ -tsmith 
enum ETeam
{
	eTeam_None,		// = 0
	eTeam_Neutral,	// = 1
	eTeam_One,		// = 2
	eTeam_DONT_USE_3,
	eTeam_Two,		// = 4
	eTeam_DONT_USE_5,
	eTeam_DONT_USE_6,
	eTeam_DONT_USE_7,
	eTeam_XCom,		// = 8
	eTeam_DONT_USE_9,
	eTeam_DONT_USE_10,
	eTeam_DONT_USE_11,
	eTeam_DONT_USE_12,
	eTeam_DONT_USE_13,
	eTeam_DONT_USE_14,
	eTeam_DONT_USE_15,
	eTeam_Alien,		// = 16
	eTeam_DONT_USE_17,
	eTeam_DONT_USE_18,
	eTeam_DONT_USE_19,
	eTeam_DONT_USE_20,
	eTeam_DONT_USE_21,
	eTeam_DONT_USE_22,
	eTeam_DONT_USE_23,
	eTeam_DONT_USE_24,
	eTeam_DONT_USE_25,
	eTeam_DONT_USE_26,
	eTeam_DONT_USE_27,
	eTeam_DONT_USE_28,
	eTeam_DONT_USE_29,
	eTeam_DONT_USE_30,
	eTeam_DONT_USE_31,
	eTeam_TheLost,		// = 32
	eTeam_DONT_USE_33,
	eTeam_DONT_USE_34,
	eTeam_DONT_USE_35,
	eTeam_DONT_USE_36,
	eTeam_DONT_USE_37,
	eTeam_DONT_USE_38,
	eTeam_DONT_USE_39,
	eTeam_DONT_USE_40,
	eTeam_DONT_USE_41,
	eTeam_DONT_USE_42,
	eTeam_DONT_USE_43,
	eTeam_DONT_USE_44,
	eTeam_DONT_USE_45,
	eTeam_DONT_USE_46,
	eTeam_DONT_USE_47,
	eTeam_DONT_USE_48,
[...]

PR #195, opened 2018-02-21
Provides basic implementation for issue #188

RealityMachina:
By basic I mean it hardcodes some things (colour of enemy units on eTeam_One and eTeam_Two, what the game does when it's their turn UI wise, etc.). They should probably be modularized at a later date for customizability reasons.

Decided to go with an implementation that provided its own basic AI Player for the teams: this way mods that don't need to do anything fancy don't need to worry about providing it on their own, they just need to provide the relevant enemies and a way to add them to missions on the right team.


X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
```diff
@@ -79,6 +90,68 @@ simulated static function RebuildPerkContentCache() {
...
+
+	foreach `XCOMHISTORY.IterateByClassType(class'XComGameState_Player', PlayerState, eReturnType_Reference)
+	{
+		if( PlayerState.TeamFlag == eTeam_Two)
+		{
+			break;
+		}
+	}
+
+	return PlayerState != none ? XGAIPlayer(PlayerState.GetVisualizer()) : none;
```
robojumper: What are these functions supposed to do? I don't think `foreach` expressions explicitly clear the iterator property upon leaving the loop normally, so what this does is actually "return the requested player if it exists, return an arbitrary player if the requested one doesn't exist".

X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
```diff
@@ -79,6 +90,68 @@ simulated static function RebuildPerkContentCache() {
...
+
+	foreach `XCOMHISTORY.IterateByClassType(class'XComGameState_Player', PlayerState, eReturnType_Reference)
+	{
+		if( PlayerState.TeamFlag == eTeam_Two)
+		{
+			break;
+		}
+	}
+
+	return PlayerState == eTeam_Two ? XGAIPlayer(PlayerState.GetVisualizer()) : none;
```
robojumper: `PlayerState.TeamFlag == eTeam_Two`

PR #199, opened 2018-02-25
Provide an input hook for custom keybinds

### LINKED ISSUE #198: Provide an input hook for custom keybinds
## The Problem
The vanilla game does not provide a simple way for mods to add custom keybinds that trigger a function. 

## The Solution
A highlander hook in **UIScreenStack:OnInput()** should enable easy custom keybinds, and allow them to be created without class overrides.

### Working Example
Here's an example of a custom function using this event hook to run a function when the '8' keybind is pressed:

```
class UIFacilityGrid_Listener_Custom extends UIScreenListener;

//When player is in facility grid view, subscribe to input events
function OnReceiveFocus(UIScreen screen)
{
	HandleInputDelegates(true);
}

//Remove input event subscription, when player leaves facility grid view
function OnLoseFocus(UIScreen screen)
{
	HandleInputDelegates(false);
}

function HandleInputDelegates(bool isSubscribing)
{
	local delegate<UIScreenStack.CHOnInputDelegate> inputDelegate;
	inputDelegate = WriteToLog;

	if(isSubscribing)
	{
		`SCREENSTACK.SubscribeToOnInput(inputDelegate);
	}
	else
	{
		`SCREENSTACK.UnsubscribeFromOnInput(inputDelegate);
	}
}

function bool WriteToLog(int iInput, int ActionMask)
{
	//write to log when '8' key is released
	if(iInput == class'UIUtilities_Input'.const.FXS_KEY_8 && ActionMask == class'UIUtilities_Input'.const.FXS_ACTION_RELEASE)
	{
		`log("Custom keybind invoked!!!");
		return true;
	}
	return false;
}

defaultproperties
{
	ScreenClass = UIFacilityGrid
}
```

KongMD-Steam:
Fix #198.

Add UIScreenStack class. Enable custom keybinds for mods, and allow them to be created without class overrides. See #198 for a complete example of how to use this input hook.

The majority of the new input code was provided by robojumper.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
@@ -733,7 +733,7 @@ function UnsubscribeFromOnInput(delegate<CHOnInputDelegate> callback)
     }
 }
 
-/* helper function to OnInput() that returns true if any mods are subscribing to the key that was pressed
+/* helper function to OnInput() that returns true if any mods handled this input event
 	Parameters:
 		iInput is the int of the most recent input event. See UIUtilities_Input class for values
 		ActionMask is the action bitmask associated with the input event. Search for 'Actions - bitmasks' 
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
@@ -1,0 +1,765 @@
+//----------------------------------------------------------------------------
+//  *********   FIRAXIS SOURCE CODE   ******************
+//  FILE:    UIScreen.uc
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -0,0 +1,765 @@
...
+function UnsubscribeFromOnInput(delegate<CHOnInputDelegate> callback)
+{
+	// remove the delegate from the array, if it exists
+    if (OnInputSubscribers.Find(callback) != INDEX_NONE)
+    {
+        OnInputSubscribers.RemoveItem(callback);
+    }
+}
+
+/* helper function to OnInput() that returns true if any mods are subscribing to the key that was pressed
```
robojumper: ~~if any mods are subscribing to the key that was pressed~~ if any mods handled this input event

PR #207, opened 2018-03-12
Add VS Code build scripts

robojumper:
This PR adds build scripts that can be picked up by Visual Studio Code IDE. They require minimal initial setup, the main setup consisting of adding the SDK and game paths in the Visual Studio Code configuration like this:

    "xcom.highlander.gameroot": "E:\\SteamLibrary\\steamapps\\common\\XCOM 2\\XCom2-WarOfTheChosen",
    "xcom.highlander.sdkroot": "E:\\SteamLibrary\\steamapps\\common\\XCOM 2 War of the Chosen SDK"

Most of the code comes from [jammerware/x2mods-dev-scripts](https://github.com/jammerware/x2mods-dev-scripts) by [@jammerware](https://github.com/jammerware) and [@tracktwo](https://github.com/tracktwo).

Compared to the batch file, this requires less `make` invocations and formats error/warning messages in a way that makes them clickable.

### DIFF
--- a/.scripts/build.ps1
+++ b/.scripts/build.ps1
@@ -189,7 +189,7 @@ $modSrcRoot = "$srcDirectory\$modNameCanonical"
 ValidateProjectFiles $modSrcRoot $modNameCanonical
 
 # clean
-$stagingPath = "{0}\XComGame\Mods\{1}\" -f $sdkPath, $modNameCanonical
+$stagingPath = "{0}\XComGame\Mods\{1}" -f $sdkPath, $modNameCanonical
 Write-Host "Cleaning mod project at $stagingPath...";
 if (Test-Path $stagingPath) {
     Remove-Item $stagingPath -Recurse -WarningAction SilentlyContinue;
@@ -286,7 +286,7 @@ for ($i=0; $i -lt $files.length; $i++) {
 }
 
 # Ideally, the cooking process wouldn't modify the big *.tfc files, but it does, so we don't overwrite existing ones (/XC /XN /XO)
-# In order to "reset" the cooking direcory, just delete it and let the game recreate them
+# In order to "reset" the cooking direcory, just delete it and let the script recreate them
 Write-Host "Copying Texture File Caches"
 Robocopy.exe "$cookedpcconsoledir" "$modcookdir" *.tfc /NJH /XC /XN /XO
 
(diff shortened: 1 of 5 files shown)


KongMD-Steam:
The -file parameter in tasks.json doesn't work when ${workspaceRoot} has spaces in the path. I put my project folder as a subfolder of the 'XCOM 2 SDK' folder. Here's the error output:

```
Processing` -File ''G:\SteamLibrary\steamapps\common\XCOM' failed: The given path
's format is not supported. Specify a valid path for the -File parameter.
```

.scripts/build.ps1
```diff
@@ -0,0 +1,309 @@
...
+}
+
+# This doesn't work yet, but it might at some point
+Clear-Host
+
+Write-Host "SDK Path: $sdkPath"
+Write-Host "Game Path: $gamePath"
+
+# Check if the user config is set up correctly
+if ([string]::IsNullOrEmpty($sdkPath) -or [string]::IsNullOrEmpty($gamePath))
```
KongMD-Steam: This existence check doesn't work, because the parameter will be passed as a literal. Script output without user config variables present:
```
SDK Path: ${config:xcom.highlander.sdkroot}
Game Path: ${config:xcom.highlander.gameroot}
```
This check worked when I tested it:
```
if ([string]::IsNullOrEmpty($sdkPath) -or $sdkPath -eq '${config:xcom.highlander.sdkroot}' -or [string]::IsNullOrEmpty($gamePath)  `
        -or $gamePath -eq '${config:xcom.highlander.gameroot}' )
```

KongMD-Steam:
Here's a sanity check we can use. Would go right after the '# Check if the user config is set up correctly' code:
```
# Verify the SDK and game paths exist before proceeding
elseif (!(Test-Path $sdkPath)) 
{
    throw "The path '$sdkPath' doesn't exist. Please adjust the xcom.highlander.sdkroot variable in your user config and retry."
}
elseif (!(Test-Path $gamePath)) 
{
    throw "The path '$gamePath' doesn't exist. Please adjust the xcom.highlander.gameroot variable in your user config and retry."
}
```

robojumper:
One unresolved "problem": Contrary to the ModBuddy MSBuild setup, this VS Code script does not require any file paths to be present in the `.x2proj` file. This means that people who use VS Code for this need to make sure that they add any new files to the `.x2proj` file manually.

KongMD-Steam:
Perhaps a new ValidateProjectFiles() function in build.ps1 would be in order? It would iterate through the files in the 'Src' subdirectory and compare them with .x2proj file entries in the XML file. If there's a mismatch, error out.

.vscode/tasks.json
```diff
@@ -0,0 +1,34 @@
...
+        },
+        {
+            "label": "debug",
+            "type": "shell",
+            "command": "powershell.exe -file '${workspaceRoot}\\.scripts\\run.ps1' -gamePath '${config:xcom.highlander.gameroot}'"
+        },
+        {
+            "label": "runUnrealEditor",
+            "type": "shell",
+            "command": "powershell.exe -file '${workspaceRoot}\\.scripts\\runUnrealEditor.ps1' -gamePath '${config:xcom.highlander.sdkroot}'",
```
KongMD-Steam: Bug: the variable name here should be -sdkPath

Issue #213, opened 2018-03-30
GetValidWeaponRange in XComGameState_Ability is broken, leading to wonky AI behavior


LeaderEnemyBoss:
[Example 1](https://media.giphy.com/media/tocvxJJrkz6pt801NE/giphy.gif)
[Example 2](https://media.giphy.com/media/5pUuPuebjtzEr3U5L2/giphy.gif)

We all know the XCom 2 AI can act goofy at times. Sometimes it seems as if it got even worse in WOTC. I think I have found the reason for at least some of these issues. The event GetValidWeaponRange is used to determine the min and max weapon range for AI purposes. However through some investigation and logging* I discovered that there are many cases, where this function just returns the wrong values. This can lead to issues like they can be observed in my example gifs. 

In Example 1, GetValidWeaponRange just returns the sight range instead of the actual range of the Andromedon's melee attack. This can under specific circumstances result in it dancing around the intended target, but never coming close enough to actually attack it.

A similar thing happens in Example 2. The Gatekeeper wants to use Consume, thinks it is in range (GetValidWeaponRange again returns the sight range) opens up. Now it cant use Consume, since nobody is near it. So it jumps to the next valid leafnode instead (MoveNoCover) and tries the same thing again.

I have a fix ready and will open a pull request soonish.

*Thanks again to robojumper for showing me how to get X2DebugBehavior to work

PR #214, opened 2018-03-30
Added XComGameState_Ability, fixed Range Calculation for AI purposes

LeaderEnemyBoss:
Handles Issue #213

I made two changes in the GetValidWeaponRange event:

1. Added an extra check for Cursor-targeted Cone abilities. The event returned the max Cursor Range (the whole map!), which could lead to strange behavior of enemies like the faceless and the purifier.

2. Removed "if( MaxRange == -1 )" before checking for AbilityTargetConditions. From what I could see MaxRange is always set by this point, which led to these conditions being ignored. They are used for abilities like Andromedons BigDamnPunch and Gatekeepers Consume. 

Note: I know changing elementary functions/events can have unintended side effects. However I think its fine in this case. GetValidWeaponRange is only used for AI purposes, so there should be not negative effects in other parts of the game. I also tested these changes for quite a few missions in my current playthrough aswell as under laboratory conditions and didnt notice any new bugs. On the contrary, the AI gets ever so slightly more challenging since its less likely to just wander around.


LeaderEnemyBoss:
An alternative way to do this would be to include a hook in the function GetValidWeaponRange and otherwise leave it as it is. I could than release a small fix mod.

Issue #251, opened 2018-07-25
Fix how XComHumanPawn handles underlays


RealityMachina:
SoldierAppearances have these wonderful slots for storing a custom set of underlay arms, legs, and torsos.

That are completely unused by the game in favour of forcing clerk assets.

There are two alterations in XComHumanPawn required to fix this;

`RequestFullPawnContent` : 

The sections for request arms and legs content need to be modified so they're set up like the torso: to still be in use if there are underlay cosmetics it can use instead of normal arms and legs.

`GetUnderlayName`

This should be altered so it actually checks the given kAppearance to see if it has any custom torso underlay to use, instead of resorting to the monolithic clerk torso content.

Since other functions in this class have already been altered, I am hopeful this is something that can be altered by the highlander successfully.

PR #252, opened 2018-07-27
Modification of XComHumanPawn for issue #251

RealityMachina:
Initial testing seems to have normal pawns work as before (since they were made without any sort of character template filtering: they would have selected the right underlay cosmetics), while custom units intending to use the feature were...

Well, I'll need to fix how they work in the code in the mod they're from so they get actual uniform underlays, but the highlander code is working fine. :v


RealityMachina:
OK lemme try something to fix my dumb history

PR #253, opened 2018-07-27
Modification of XComHumanPawn for issue #251

RealityMachina:


RealityMachina:
So further testing reveals a bit of a thorny issue: it seems like the original split underlay assets were never completed for soldiers.

Might need an added configurable .ini so it works on a whitelist system: allow only character templates that will be using proper underlays and otherwise resort to the standard clerk outfits.

robojumper:
`XComGame\Classes\XComHumanPawn.uc:163 : Error, You can only access default values of variables here`:

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/b3efe40498f98b23e443e535937060d8e4f9e9d3/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHumanPawn.uc#L163

`GetUnderlayName` is static.

Issue #260, opened 2018-08-01
Add support for letting weapons decide which upgrades can apply to them


GingerAvalanche:
Currently, X2WeaponUpgradeTemplate has CanApplyUpgradeToWeaponFn which is used to determine whether or not an upgrade will let itself be attached to a weapon. However, there is no logic which does the opposite: allows a weapon to determine whether or not it will let an upgrade be attached to it. With mods adding more weapon categories and blurring the lines between what can and cannot have an upgrade, there are times when an upgrade believes it should be allowed to apply itself to a weapon via one mod, but another mod that adds a weapon may not want that upgrade. For example, a primary sword, by default, would be a suitable match for most mod-added scopes, which is of course silly.

Propose to add a CanWeaponApplyUpgradeFn and CanWeaponApplyUpgradeDelegate to X2WeaponTemplate, as well as a wrapper function CanWeaponApplyUpgrade which would either run the CanWeaponApplyUpgradeFn if present or otherwise simply return true, which UIArmory_WeaponUpgrade would call alongside CanApplyUpgradeToWeaponFn to allow both the upgrade and the weapon to deny the attachment.

For simplicity's sake, also propose to add an array of names UpgradeCats to X2WeaponUpgradeTemplate which would store the categories of weapon that an upgrade should be attached to by default. As most mod-added upgrades that limit themselves do so on the basis of the WeaponCat of the weapon to be attached to, I foresee the most common use of CanWeaponApplyUpgradeFn to also be checking categories, which this change enables the possibility of. The desired functionality is certainly possible without this change, but it would generally be unnecessarily complicated compared to what could be achieved with this change. It also allows future mods to compare categories from either direction: in CanApplyUpgradeToWeaponFn or CanWeaponApplyUpgradeFn.

GingerAvalanche:
This issue was dead before it began. I didn't realize X2WeaponTemplate is native. Variables and delegates may not be added to it. Will probably revisit this later via a different method.

GingerAvalanche:
Mod Everything is a mod that enables upgrades to be added for any XComGameState_Item, not just those made from an X2WeaponTemplate. As such, I'm going to remove that safety check, so as not to break mods like that one and anything that may rely on them. If necessary, any modder that wishes to restrict upgrades being applied to their items may check it within the DLCInfo function.

GingerAvalanche:
[...]

PR #261, opened 2018-08-02
Reimplementation of Issue #260 changes

GingerAvalanche:
Ignore my messy commits, I had a hell of a time with GitHub merging my testing branch back into my dev branch every time I merged the dev branch into the testing branch.


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
```diff
@@ -2044,7 +2044,27 @@ function bool IsMissionObjectiveItem()
 	return false;
 }
 
+// Issue #260 start
+// This function will be used to cycle through DLCInfos that will allow mods to check generally whether or not a weapon is compatible with an upgrade
+// X2WeaponUpgradeTemplate::CanApplyUpgradeToWeapon still exists as the "can this upgrade be applied to this weapon RIGHT NOW?"
+function bool CanWeaponApplyUpgrade(X2WeaponUpgradeTemplate UpgradeTemplate)
```
robojumper: Either include the call to the delegate in UpgradeTemplate here, or name this function in a way that makes it clear this only considers the DLC hook (e.g. `DLCCanWeaponApplyUpgrade`)

GingerAvalanche: XComGameState_Item::CanWeaponApplyUpgrade is a mirror of X2WeaponUpgradeTemplate::CanWeaponApplyUpgrade which does nothing but call the CanApplyUpgradeToWeaponFn. The purpose of these wrapper functions is simply to ensure that an upgrade is allowed if there is no explicit reason to deny it, instead of the default behavior that would occur if calli he delegate directly (denying the upgrade if there is no explicit reason to allow it).

Since XComGameState_Item::CanWeaponApplyUpgrade is called alongside X2WeaponUpgradeTemplate::CanApplyWeaponToUpgrade wherever it’s used, calling the upgrade delegate in CanWeaponApplyUpgrade is unnecessary.

I can still change the name if you want me to, it’s like 3 letters and isn’t worth an extended discussion. I just don’t know if that explanation changes your concern at all.

robojumper: I see. Probably fine as is, then.

Issue #237, opened 2018-06-30
Revamp weapon upgrades to be more universal


GingerAvalanche:
Currently, weapon upgrades can modify several aspects of a weapon's abilities, but damage and critical damage are not among them. Modders can use X2Effects to get around this, but it leads to quite a bit of wonkiness at the UI level.

Propose to change items and upgrade templates in order to add base-level compatibility for increasing the damage, critical damage, shred, and pierce of a weapon, directly onto the upgrade template.

To begin with, X2WeaponUpgradeTemplate would need four new int properties and four new delegates to govern their interaction with the rest of the code:
`DmgBonus`
`CritDmgBonus`
`PierceBonus`
`ShredBonus`
`AddDamageModifierFn`
`AddCritDamageModifierFn`
`AddPierceModifierFn`
`AddShredModifierFn`

There are also several parts of XComGameState_Item that are coded under the assumption that weapon upgrades cannot modify damage or critical damage:
`GetWeaponStatsForUI`
`GetUpgradeEffectForUI`
`GetUpgradeModifiersForUI`
`GetUISummary_WeaponStats`
These would need to be modified in order to accept the new properties from the template.

Lastly, and the most problematic, is that the interface for passing around this information, EUISummary_WeaponStats, is a struct that does not contain properties governing critical damage, pierce, or shred (though it does contain a property for damage). This would need to be changed, but as it is used in XComGameState_Item, a native class, this may not be possible. Will update.

GingerAvalanche:
Several parts of the code included unfinished pieces that were clearly meant to function towards allowing weapon upgrades to modify more things, but at the last minute they seem to have scrapped those ideas and hardcoded in limitations. For example, when calculating damage, `X2Effect_ApplyWeaponDamageToUnit` assumes that there can be more than one upgrade affecting damage (even though that part of the code was behind a conditional that could never be `true`), while `GetUpgradeModifiersForUI` assumes that anytime an upgrade changes a stat, it will be the only one to do so.

In order to make this universal, and to provide more leeway for mods, I went ahead and changed all instances of "set bonus value to X" to "add bonus value to X."

GingerAvalanche:
Pitfall: any damage increases from an attached upgrade will not show on the UI wing in the tactical HUD.

[...]

PR #262, opened 2018-08-02
Implementation of Issue #237 changes

GingerAvalanche:
The branch is fixed. It doesn't contain the changes from issue 235 anymore. Hopefully it stays that way.


X2WOTCCommunityHighlander/Src/XComGame/Classes/XLocalizedData.uc
```diff
@@ -0,0 +1,239 @@
...
+
+// Item/weapon stats
+var localized string CriticalDamageLabel;
+var localized string CriticalChanceLabel;
+var localized string CriticalChanceBonusLabel;
+var localized string PoisonLabel;
+var localized string BurningLabel;
+var localized string ArmorChanceLabel;
+var localized string PierceLabel;
+var localized string RuptureLabel; // Issue #237
```
robojumper: We don't have any localization for this, do we?

Issue #278, opened 2018-09-18
Extend PostEncounterCreation hook to include reinforcement spawns


GingerAvalanche:
#136 adds the capability to modify encounters (pods) before a mission starts, but any reinforcement pods are generated dynamically within the mission by XComGameState_AIReinforcementSpawner.

Propose to add a call to PostEncounterCreation in XComGameState_AIReinforcementSpawner::OnReinforcementSpawnerCreated, immediately after the SpawnInfo is filled out by the SpawnManager. OnReinforcementSpawnerCreated already has context to all the appropriate variables.

Also propose to add an additional optional variable to PostEncounterCreation, a second XComGameState_BaseObject, which will carry the state corresponding to the calling instance, whose existence can be used to differentiate between mission generation and reinforcement generation, and whose properties can be used to restrict pod manipulation based on usage scenario, e.g. if a modder wants to target/ignore Chosen reinforcement spawns.

When XComGameState_MissionSite calls PostEncounterCreation, it passes itself as the first XComGameState_BaseObject which is useful because it contains things such as the mission schedule, which contains the default spawn lists for the mission. It may be a good idea to instead pass the BattleData as the first XComGameState_BaseObject so that current battle conditions can be taken into account when determining reinforcement units. The Mission Site state can be accessed through the BattleData, so nothing will be lost by this switch.

Due to the fact that OnReinforcementSpawnerCreated later validates the circumstances of the spawn data, placing the call immediately after the SpawnInfo is filled out allows modders to arbitrarily assign units to the reinforcement pods without fear of glitches such as matinees, for example if a large unit or alien were to try to jump out of an ATT.

GingerAvalanche:
Changing the definition of PostEncounterCreation would break mods that currently utilize it, but I really think it would be beneficial to have the added properties.

Instead, will create a new hook, PostReinforcementCreation, that contains all the aforementioned changes to properties and which will, by default, serve as a wrapper for the old hook.

PR #285, opened 2018-09-29
Reinforcement changer

GingerAvalanche:
Implements changes for issue #278

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
@@ -366,6 +366,19 @@ static function PostEncounterCreation(out name EncounterName, out PodSpawnInfo E
 }
 // End Issue #136
 
+// Start Issue #278
+/// <summary>
+/// Called from XComGameState_AIReinforcementSpawner:OnReinforcementSpawnerCreated
+/// SourceObject is the calling function's BattleData, as opposed to the original hook, which passes MissionSiteState. BattleData contains MissionSiteState
+/// Added optional ReinforcementState to modify reinforcement conditions
+/// Encounter Data is modified immediately after being generated, before validation is performed on spawn visualization based on pod conditions
+/// </summary>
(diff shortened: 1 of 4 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -366,6 +366,19 @@ static function PostEncounterCreation(out name EncounterName, out PodSpawnInfo E
...
+// Start Issue #278
+/// <summary>
+/// Called from XComGameState_AIReinforcementSpawner:OnReinforcementSpawnerCreated
+/// SourceObject is the calling function's BattleData, as opposed to the original hook, which passes MissionSiteState. BattleData contains MissionSiteState
+/// Added optional ReinforcementState to modify reinforcement conditions
+/// Encounter Data is modified immediately after being generated, before validation is performed on spawn visualization based on pod conditions
+/// </summary>
+static function PostReinforcementCreation(out name EncounterName, out PodSpawnInfo Encounter, int ForceLevel, int AlertLevel, optional XComGameState_BaseObject SourceObject, optional XComGameState_BaseObject ReinforcementState)
+{
+	PostEncounterCreation(EncounterName, Encounter, ForceLevel, AlertLevel, `XCOMHISTORY.GetGameStateForObjectID(XComGameState_BattleData(SourceObject).m_iMissionID));
```
robojumper: Why are the parameters `optional` and `XComGameState_BaseObject`? Can we not rely on those objects always having the correct type? When are they `none`?

GingerAvalanche: The first parameter is `optional` and `XComGameState_BaseObject` because that’s how PostEncounterCreation did it and I was mimicking the behavior. We can rely on them always having the correct type. They are never `none`.

I can change that if it’s important. I just tend to assume things were done a certain way for a reason and copy behavior when I haven’t figured out a limiting reason not to.

robojumper: Ah. I think we copied over that hook from the LW2 Highlander. Not sure if we want to do it better?

GingerAvalanche: Can’t change the original hook at this point without breaking anything that uses it, but I can change this one when I get the time.

PR #288, opened 2018-10-04
Improve Log message accuracy

robojumper:
The old log messages are kind of inaccurate and don't give a full picture. In particular:

* `present and correct` is wrong. That message only checks for the companion script package.
* `X2CH SCREEN LISTENER ON SPLASH` doesn't give any warning about a missing core package replacement.
* `Creating CHXCOMGameVersionTemplate` is pretty bad too -- we want to show a version here as well in case the companion script package fails.

This PR addresses these issues and adds the `X2WOTCCommunityHighlander` log tag to all log messages.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_UIScreenListener_ShellSplash.uc
+++ b/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_UIScreenListener_ShellSplash.uc
@@ -3,6 +3,7 @@ class X2WOTCCH_UIScreenListener_ShellSplash extends UIScreenListener config(Game
 var config bool bEnableVersionDisplay;
 
 var localized string strCHLVersion;
+var string HelpLink;
 
 event OnInit(UIScreen Screen)
 {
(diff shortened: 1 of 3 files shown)


X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_UIScreenListener_ShellSplash.uc
```diff
@@ -62,5 +67,6 @@ function OnTextSizeRealized()
 
 defaultProperties
 {
-    ScreenClass = none
+	ScreenClass = none
+	HelpLink = "https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/issues/228#issuecomment-422218639"
```
Xymanek: Maybe a separate page is better rather than a single comment somewhere mid-discussion? For example, you can use github wiki. Alos, updating a wiki page will be much easier than a hardcoded link

robojumper:
Good idea. Added https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/wiki/Troubleshooting

Xymanek:
That page is missing info about family sharing and no net connection, which both are know to cause problem with loading HL from workshop mods

robojumper:
Updated the wiki page.

Issue #299, opened 2018-10-12
Add an Event that allow override if Appearance should be randomized


Musashi1584:
Add an event in `XComGameState_Unit.AddItemToInventory`

PR #300, opened 2018-10-12
Issue #299 Add an Event that allow override if appearance should be randomized

Musashi1584:


robojumper:
Please mention this in the changelog.

Issue #307, opened 2018-10-15
Make "Legacy Operations" respect Squad/Soldier/Class changes


MrNiceUK:
Except for a new random Operation, where the squad must be generated from scratch, for the first mission in a Legacy Ops soldiers are used directly from the save for the first mission, without referencing the relevant squad/soldier progressions and soldier class templates. For later missions, changes to squad progression made by mods is respected. To open this design space for mods, I propose that the CHL should recreate the squad (similar to what happens when you create a new random operation), and so pick up changes to relevant  configs, soldier classes, etc.

Xymanek:
> To open this design space for mods

Do you mean making new legacy ops? Or changing the existing

MrNiceUK:
@Xymanek I mean changing existing. Basically, although it appears you can change some aspects of existing Legacy Ops through ini edits and similar, most changes will be ignored for the first mission, while taking effect in later missions.

MrNiceUK:
Ok, changed the Issue Title, since changing the soldier class and/or soldier class templates doesn't work for any missions, it's not just a "first mission" issue. Same for CharacterTemplate and changing squad size, which are related issues.

PR #308, opened 2018-10-16
Issue #307 LoadOut, Abilities and SoldierClass Changes

MrNiceUK:
#307 As in title, now loadout and ability changes are respected for the first mission not just later, and soldier class changes now work at all, for all missions.


MrNiceUK:
Ok, going to re do commits so there's an initial add of the srcorig original, so that the actual changes can be shown as a diff!

Issue #313, opened 2018-10-21
XCGS_Unit:GetStatModifiers reports bogus modifiers for multipliers


robojumper:
https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/8954b83b97d6296e26003dea22ea32da2af76c4f/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc#L6514

Instead of returning the correct contribution `(mult - 1) * current_value`, it seems to return ~~the value after multiplication `mult * current_value`, at least for `MODOP_Multiplication` (and probably `MODOP_PostMultiplication` too).~~ `mult * base_stat` for both `MODOP_Multiplication` and `MODOP_PostMultiplication` as @MrNiceUK has pointed out to me.

The only user of this is `X2AbilityToHitCalc_StandardAim`, which computes a bogus to-hit chance:
https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/8954b83b97d6296e26003dea22ea32da2af76c4f/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityToHitCalc_StandardAim.uc#L364-L369

and a bogus to-crit chance:

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/8954b83b97d6296e26003dea22ea32da2af76c4f/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityToHitCalc_StandardAim.uc#L523-L528

where it gives you a positive bonus for multipliers between 0 and 1.

We should fix (or work around) the problem.

RedDoby:
I tested this with Simple Red Fog and it is still giving a positive bonus modifier.

RedDoby:
Are you sure that this is working as intended?  I tested it last month with the Simple Red Fog mod and the Beta Highlander, and it was still giving the positive offensive modifiers for values 0-1.

RedDoby:
I double checked the classes XComGameState_unit and X2AbilityToHitCalc_StandardAim for the CHE that I was using at the time of testing.  These changes are in the code.  I will try it again with the debugger and see what is going on.

RedDoby:
[...]

PR #314, opened 2018-10-22
Fix for #313, broken GetStatModifiers()

MrNiceUK:
#313


Musashi1584:
Can this be potentially merged? If so why hasn't it already been merged?

MrNiceUK:
Reason why it hasn't already? Closest to an answer:
https://discordapp.com/channels/165245941664710656/273238884433788928/506778534929367043

Xymanek:
The repercussions of possible/accidental mis-implementation in this will disastrous. I would argue that this needs very through testing, much more thorough than most others CHL features

MrNiceUK:
Well it's had 6 months for someone other than me to actually test it in....

Issue #137, opened 2017-12-13
[Discussion] Inventory Slot uses record (for #118)


robojumper:
This issue is a discussion thread for requesting / adding / using new slot enum values for #118.
A slot can have several entries (mods that use them), in case compatibility between these mods is not desirable. In this sense, this record helps mod authors to make an informed decision about whether there is already a slot that they can  or should use.

Mod authors who add a new slot or (plan to) use an existing slot *should* add a comment here.

Enum | Requesting User(s) | Purpose | Workshop Link
---- | ---- | ---- | ----
eInvSlot_Vest | @robojumper | A dedicated slot for vests similar to Grenade/Ammo pockets. | \<none\>

*Last Updated 2017-12-13T16:00:53+00:00*

loadhwa:
Enum | Requesting User(s) | Purpose | Workshop Link
------------ | ------------- | ------------- | -------------
eInvSlot_AugmentationHead | @Musashi1584 | A custom slot for soldier head augmentations. | \<none yet>
eInvSlot_AugmentationTorso | @Musashi1584 |  A custom slot for soldier torso augmentations. | \<none yet>
eInvSlot_AugmentationArms | @Musashi1584 | A custom slot for soldier arms augmentations. | \<none yet>
eInvSlot_AugmentationLegs | @Musashi1584 | A custom slot for soldier legs augmentations. | \<none yet>

Musashi1584:
Enum | Requesting User(s) | Purpose | Workshop Link
------------ | ------------- | ------------- | -------------
eInvSlot_AugmentationHead | @Musashi1584 | A custom slot for soldier head augmentations. | \<none yet>
eInvSlot_AugmentationTorso | @Musashi1584 |  A custom slot for soldier torso augmentations. | \<none yet>
eInvSlot_AugmentationArms | @Musashi1584 | A custom slot for soldier arms augmentations. | \<none yet>
eInvSlot_AugmentationLegs | @Musashi1584 | A custom slot for soldier legs augmentations. | \<none yet>

robojumper:
Enum | Requesting User(s) | Purpose | Workshop Link
------------ | ------------- | ------------- | -------------
eInvSlot_CombatDrugs | @robojumper | An inventory slot for consumable combat drugs like Overdrive Serum | \<none yet>

Hotl3looded:
| Enum  | Requesting<br>User(s) | Purpose | Workshop<br>Link | 
[...]

PR #316, opened 2018-10-24
Photobooth pose filters for soldier class mods

robojumper:
Allows mods to claim "this class behaves like a Ranger" for purposes of pose filtering... or just implement their own filter like `ePAFT_FieldMedic`. 

TODO: Set up a thread for requests à la https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/issues/137

Proper compare view: 9d2446d...f7f351c


bountygiver:
can you also change the spark poses to match with unit template instead of class templates?

robojumper:
@bountygiver Done!

robojumper:
Tracking issue opened and mentioned in the code, I think we're all set.

Issue #328, opened 2018-11-07
Allow Torsos, Arms, Legs to have localized names


robojumper:
X2BodyPartTemplateManager has this code:

		// HAX: Load 'Torso', 'Arms' and 'Legs' templates the old way to prevent template name collisions:
		// TODO: @rmcfall - make all body part template names unique so they can be individually localized
		switch(PartInfo.PartType)
		{
		case "Torso":
		case "Arms":
		case "Legs":
			Template = new class'X2BodyPartTemplate';
			break;
		default:
			// This is the "new", proper way of instantiating a template (requires unique TemplateName)
			Template = new(None, string(PartInfo.TemplateName)) class'X2BodyPartTemplate';
		}

While a mod [exists on the Workshop](https://steamcommunity.com/sharedfiles/filedetails/?id=878158600) to work around the problem, it's still a dependency, is awkward to set up, and has some sort of race condition with Unrestricted Customization[1]. Some fix in here shouldn't cause too many problems.

--------------------
[1] I still have no idea what the hell that mod is doing and why it breaks Add Part Names, but I'm like 50% sure that an HL fix wins over Unrestricted Customization.

MrNiceUK:
The "plan" for the enhancement is to concatenate the template name and part type to create the unique instance name. Probably also requires changes to the Armoury code to actually check for the DisplayName of the 3 body types. On a related note, apparently Shin decos don't show the display name for equipped parts, only for the selection list. Other less used deco slots may also have that issue.

MrNiceUK:
Do we take it upon ourselves to give reasonable display names to vanilla/WotC/Anarchy's Children parts?

robojumper:
> Do we take it upon ourselves to give reasonable display names to vanilla/WotC/Anarchy's Children parts?

Some people already did, but I'm not sure whether we want to ship it with the HL.

https://steamcommunity.com/workshop/filedetails/discussion/878158600/135511455870610385/

Dragon-32:
Well I prefer mine ;)

I'd say not to include names in the HL, chances are people will release mods making use of this just like Vortex Pixalation and Musashi have released mods changing a value or two in the HL .ini files.

PinkysBrain:
Unrestricted Customization seems to build display names for the menus on the fly from the template names even when it doesn't have localization.

[...]

PR #332, opened 2018-11-09
Issue #328 Torso/Arms/Legs Localization

MrNiceUK:
#328 
Uses '_*PartType*' as the suffix for Torso/Arms/Legs object names. Also fixes Shins not having their display name shown in the overview of selected parts. Due to allowing for Torso/Arms/Legs not necessarily having a DisplayName set, showing the Index number in the overview is used as a fallback (which is what Vanilla shows for these parts). As a consequence of the implementation, this fall back is used for all parts, not just these 3 (and Shins). Essentially, the Vanilla code assumed that all other part types would have correctly set localized DisplayNames. This will only affect mod parts, which before would have shown nothing at all in the overview if selected.


Musashi1584:
Can this be potentially merged? If so why hasn't it already be merged?

PR #340, opened 2018-11-15
DLC script replacements, component version checks

robojumper:
This PR adds a better versioning mechanism to the Highlander, providing more feedback about whether the Highlander was installed correctly with all its components. It also lays the groundwork for making separate mods that replace DLC scripts, while interfacing with the Highlander for purposes of version checks.
Still needed:

* [x] A way to show details about installed/enabled/expected components on the shell screen. custom tooltip? How about controllers?
* [ ] An updated troubleshooting page to direct users to
* [x] An update to the PowerShell build script for the DLC mods

grafik


robojumper:
Note: The DLC_2 (and DLC_3) checks are disabled, as we currently don't plan on shipping a DLC Highlander with 1.17.0. We'll re-enable them when needed.

PR #344, opened 2018-11-17
BugFix#307

### LINKED ISSUE #307: Make "Legacy Operations" respect Squad/Soldier/Class changes

MrNiceUK:
Fixes issue with starting with the last mission, not the first, when restarting an Op which was last completed, not abandoned. Also fixed that the greyed out difficulty selection for random Ops did not reflect the locked difficulty. For technical reasons, this fix was required for the first fix. Lastly, a crash condition from Vanilla when backing out from a random op restart, then immediately re-entering and confirming. This was caused by not resetting an index correctly on backing out. #307 #308 

Closes #307 -- @robojumper

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITLE_LadderModeMenu.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITLE_LadderModeMenu.uc
@@ -385,6 +385,12 @@ simulated function OnDifficultySelectionCallback(name eAction, int selectedDiffi
 
 	if (eAction != 'eUIAction_Accept')
 	{
+		// Start Issue #307
+		if (ClickedList == List)
+		{
+			clickedIndex -= NarrativeList.GetItemCount() - 1;
+		}
+		// End Issue #307
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UITLE_LadderModeMenu.uc
```diff
@@ -385,6 +385,12 @@ simulated function OnDifficultySelectionCallback(name eAction, int selectedDiffi
 
 	if (eAction != 'eUIAction_Accept')
 	{
+		// Start Issue #307
+if (ClickedList == List)
+{
+	clickedIndex += NarrativeList.GetItemCount() - 1;
+}
```
robojumper: Indentation!

MrNiceUK:
As per the commit name. Related to that, Lazarus Operation gives Central bladestorm from mission one, the only operation which does. Since he also gets no new abilities in mission two, that may have just been a Firaxis copypasta error. Since in vanilla, there is no "progression" before the first mission, that this would be blocked because his custom soldier class has it as a Corporal, not Squaddie, ability doesn't come up. The CHL change, which does do pre first mission progression to allow config changes to take effect, did block this first mission bladestorm. Also widened so that Faction Hero hero classes can given abilities from their random decks on progression, and also cross class eligible abilities are also allowed. For maximum flexibility, this is done regardless of the classes "awceligible" status, or the classes AWC exclusion list.

PR #381, opened 2019-01-08
#380 Armoury Gremlin Customization Fix

### LINKED ISSUE #380: Gremlin Customization Update Errors
The Gremlin pawn in the Armoury is not correctly updated when customization options are changed for the weapon and/or armour. When the weapon is customized via the "upgrade weapon" menu, then the Gremlin is not updated at all. When the weapon is customized via the Unit customization menu, then the secondary color is erroneously set to iWeaponDeco, which is not actually used at all in the game. When the secondary armor color is customized, then the Gremlin is again not updated at all.

MrNiceUK:
Closes #380 -- @robojumper


Musashi1584:
Can this be potentially merged? If so why hasn't it already been merged?

PR #389, opened 2019-01-13
Allow mods to override the transition map and mission intro

### LINKED ISSUE #388: Allow mods to override the transition map
Functionality should be identical to the hook from the original LW2 Highlander.

### LINKED ISSUE #395: Allow mods to override the mission intro
Similar to the LW2 Highlander:

```
// Start Issue #395
/// <summary>
/// Called from XComTacticalMissionManager:GetActiveMissionIntroDefinition before it returns the Default.
/// Notable changes from LW2: Called even if the mission/plot/plot type has an override.
/// OverrideType is -1 for default, 0 for Mission override, 1 for Plot override, 2 for Plot Type override.
/// OverrideTag contains the Mission name / Plot name / Plot type, respectively
/// Return true to use.
/// </summary>
static function bool UseAlternateMissionIntroDefinition(MissionDefinition ActiveMission, int OverrideType, string OverrideTag, out MissionIntroDefinition MissionIntro)
{
	return false;
}
```

The override logic is more complex, so the hook gets a bit more information for WotC!

robojumper:
Closes #388.
Closes #395.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,7 @@ All notable changes to Vanilla 'War Of The Chosen' Behaviour will be documented 
 - `WeaponInitialized` added to conditionally change the weapon archetype on initialization (#245)
 - `UpdateWeaponMaterial` added to conditionally change the weapon materials(#246)
 - `DLCAppendWeaponSockets` allows adding new sockets to weapons(#281)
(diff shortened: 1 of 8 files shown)


robojumper:
Shamelessly decided to tack #395 on.

Issue #400, opened 2019-01-25
Support dynamic colours for tactical HUD ability icons?


pledbrook:
The LW2 highlander modifies `UITacticalHUD_Ability.UpdateData()` (if I remember correctly — I don't have access to the source right now) so that it handles ability colours of "Variable". When it encounters a colour with that value, it triggers an event with an `XComLWTuple`. This allows the mod to use different colours for the same ability depending on various factors.

I'm wondering whether it makes sense to add the same or a similar feature to X2WOTCCommunityHighlander, or whether this should be handled in a completely different way.

robojumper:
This is definitely something we'd accept in the Highlander.

Iridar:
The PR that was supposed to close this issue has been merged to the live version of the Highlander. It remains open because it needs docs. Please add some :)

Iridar:
@pledbrook I wanted to make docs for this issue, but while I was testing it I noticed the event only ever triggers for abilities that 
`if (IsObjectiveAbility || AbilityTemplate.AbilityIconColor != "")`

Can you please explain what is the reason for this restriction?

robojumper:
Also see #750 for an attempt to rework this.

Iridar:
> Also see #750 for an attempt to rework this.

Ok, thanks, I'll take that over.

PR #401, opened 2019-01-25
Implements solution for issue #400

pledbrook:
This change adds a copy of WOTC's `UITacticalHUD_Ability` class with slightly
modified code for rendering the colour. Specifically, the new code (based on
the old LW2 changes) fires one of two events that allow listeners to override
the ability colour in the Tactical HUD using the provided XComLWTuple.

From a mod's perspective, it can now register listeners for the following
events:

 * 'OverrideObjectiveAbilityIconColor'

   Allows the listener to override the colour of an ability that is tied to a
   mission objective. The normal colour is orange.

   The event includes a 2-element `XComLWTuple` with an ID matching the event
   name.

   The first element of the tuple is a boolean for which `true` means the
   listener wants to override the ability colour.

   The second element of the tuple is a string representing the desired colour.
   See `UIUtlities_Colors` for examples of this string's format.

 * 'OverrideAbilityIconColor'

   Allows the listener to override the colour of any ability whose initial
   colour is the string "Variable".

   The event includes a single-element `XComLWTuple` containing the desired
   colour as a string

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,9 @@ All notable changes to Vanilla 'War Of The Chosen' Behaviour will be documented 
   so that players don't know exactly where reinforcements will be arriving (#448)
 - `OverrideReinforcementsAlert` allows mods to force the display of the reinforcements
   alert panel and also change its text and color (#449)
+- `OverrideAbilityIconColor` provides a tuple with the same ID as the
+  event and data of the form `[bool IsObjective, string Color]` that allows
+  mods to override the color of soldier abilities in the tactical HUD (#400)
 
 ### Configuration
 - Added ability to modify default spawn size (#18)
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Ability.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Ability.uc
@@ -1,0 +1,365 @@
+//----------------------------------------------------------------------------
+//  *********   FIRAXIS SOURCE CODE   ******************
+//  FILE:    UITacticalHUD_AbilityContainer.uc
+//  AUTHOR:  Brit Steiner, Sam Batista
+//  PURPOSE: Containers holding current soldiers ability icons.
+//----------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


robojumper:
I wonder if limiting it to abilities whose color is set to "Variable" and abilities tied to mission objectives is really necessary. If we just triggered a single kind of event for every ability, one could even implement something like "Cost-Based Ability Colors" without overrides. There might be performance concerns, though I don't think we have ever done any profiling on the costs of event triggers/listeners.

pledbrook:
I'll leave profiling to those with more experience with these systems. My educated guess based on LW2 using a lot of events is that the overhead is pretty darn small as long as listeners don't block it.

I think it's reasonable to send a generic event. All one would need to do is pass in the "normal" colour. Listeners could identify objective abilities via that value, i.e. whether it's orange or not. And variable colours would have the value "Variable".

pledbrook:
@robojumper Shall I make the change to a general ability colour event? Or wait until someone knows what the performance cost might be?

pledbrook:
I would hold off on putting this into the next release. I want to think about implementation. I think it might make sense to turn it into something more generally applicable. Plus there's an oddity with objective colouring that I need to investigate.

pledbrook:
I wouldn't mind redoing this. I think a single event that triggers for all abilities is better than the current solution, with the tuple looking something like `{ID: "OverrideAbilityIconColor", Data: [in bool IsObjective, out string Color]}`, and the standard colour being sent in the tuple as the "default" value.

MrNiceUK:
Yeah, single event makes way more sense. Then a mod can change the colour for whatever reason it likes between the ability state and the existing colour.
Assume the reference to objective abilities being orange is a LWotC thing? It's Green for standard Xcom2/WotC?

pledbrook:
Don't know what the colours are for vanilla. But orange is definitely the LW2 colour for objectives. Except it's applying to Haywire for example.

pledbrook:
@robojumper This is now reimplemented. See the commit message for an overview of the new behaviour. It's the same as discussed, i.e. fire an event always for ability icons.

It basically works the same in LW2. The mod overwrites the default colours of a bunch of abilities to "Variable" and only changes the colour of ability icons that have that colour string. Well, those and the objective abilities.

Issue #409, opened 2019-02-02
Allow mods to add parallel soldier abilities, such as officer abilities, and have them displayed


pledbrook:
Much of XCOM 2's UI uses `XComGameState_Unit.GetEarnedSoldierAbilities()` to gain access to a soldier's current abilities, but it's hard-coded to only include class and AWC abilities. The officer ability tree is a parallel set of abilities, so soldiers' officer abilities aren't displayed in the main game.

The highlander should allow mods to add their own abilities. This would enable other systems, like allowing normal classes to also have psi powers, just as in XCOM 1.

pledbrook:
I've realised there's a significant issue here: `XComGameState_Unit.HasSoldierAbility()` is native and appears to ignore what's in earned abilities. Long War 2 replaced the native function with an Unreal Script version to overcome the issue. Are there any other solutions? At the moment, `HasSoldierAbility()` and `GetEarnedSoldierAbilities()` are inconsistent with one another.

Iridar:
Not sure why this was reopened, closing.

PR #411, opened 2019-02-03
Add support for adding earned soldier abilities

pledbrook:
This change introduces a new DLC hook, `ModifyEarnedSoldierAbilities`,
that allows mods to add to or modify a soldier's abilities. For example,
you could add officer or psi abilities to existing soldier classes.


pledbrook:
Implements #409

Issue #408, opened 2019-02-02
Provide a way to allow mods to override the displayed rank for soldiers


pledbrook:
This is similar to #106 and should probably be implemented in the same way. This feature will allow soldiers to be displayed with their officer ranks rather than their normal ones.

Issue #106, opened 2017-11-22
Allow mods to change class icons and class display names dynamicly


Musashi1584:
In order to allow mods to change class name, summary and icon dynamcly i propose to add two functions in `XComGameState_Unit`

```
GetSoldierClassIcon()
GetSoldierClassDisplayName()
GetSoldierClassSummary()
```

and change all old references to the the soldier class template to the new functions.

The new functions trigger events that allow to override the default.

Musashi1584:
Whats better  DLCInfo hook or event for this one?

pledbrook:
Could this solution be extended to displayed ranks? The Long War 2 officer changes mean that soldiers can have two ranks and LW2 prefers to display the officer rank. The LW2 solution involved changing a lot of places in XComBase that fetched the soldier rank and replaced those with its own method.

PR #416, opened 2019-02-08
Allow mods to override displayed ranks (#408)

pledbrook:
(**Note** This pull request also adds some implementation for #106 in places that seem to have been overlooked, such as in `UIChosenMissionSummary`)

Resolves issue #408. This pull request introduces the following methods to
`XComGameState_Unit` that trigger events that listeners can use to
override rank names and icons:

 - GetSoldierRankName()
 - GetSoldierShortRankName()
 - GetSoldierRankIcon()

The commit also updates all the bits of code (outside of TLE and
multiplayer) that display soldier ranks and their icons so that they use
the above methods.

Here are the events that are fired:

 - `{ID: SoldierRankName, Data: [in int Rank, out bool override, out
 string RankName]}`
 - `{ID: SoldierShortRankName, Data: [in int Rank, out bool override,
 out string RankName]}`
 - `{ID: SoldierRankIcon, Data: [in int Rank, out bool override, out
 string IconPath]}`

Note that the rank sent in the event may be -1, in which case the unit's
current rank is being requested.

The overall implementation is very similar to that of #106.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -162,6 +162,9 @@ All notable changes to Vanilla 'War Of The Chosen' Behaviour will be documented 
 - `OverrideItemMinEquipped` added to allow mods to override the min number of equipped items in a slot (#171)
 - `AddConversation` added to allow mods to change narrative behavior before they are played (#204)
 - `OverrideRandomizeAppearance` added to allow mods to block updating appearance when switching armors (#299)
+- `XComGameState_Unit` triggers `SoldierRankName`, `SoldierShortRankName` and
+  `SoldierRankIcon` events that allow listeners to override the those particular
+  properties of a soldier's rank, i.e. rank name, short name and icon.
 
 ### Configuration
 - Able to list classes as excluded from AWC Skill Rolling, so they can still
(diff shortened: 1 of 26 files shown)


robojumper:
This looks great. There is one thing were I'd like some changes, and that's the "format" of the event tuples. I'm not a huge fan of the `bool override` parameter, and #106 demonstrates how it can work without an override boolean -- simply set the initial value of the `out string` to the default value. I realize we haven't been 100% consistent here, but since the hooks are so similar to the hooks in #106, the implementation should be similar too.

pledbrook:
The boolean parameter has the advantage that it avoids potentially expensive calculations of the default value. Not saying that's the case here and I'm happy to make the change. Just pointing out that it can be justified.

pledbrook:
I've updated `XComGame_Unit` so that it just sends the default value through to listeners. No boolean.

CHANGELOG.md
```diff
@@ -162,6 +162,9 @@ All notable changes to Vanilla 'War Of The Chosen' Behaviour will be documented
 - `OverrideItemMinEquipped` added to allow mods to override the min number of equipped items in a slot (#171)
 - `AddConversation` added to allow mods to change narrative behavior before they are played (#204)
 - `OverrideRandomizeAppearance` added to allow mods to block updating appearance when switching armors (#299)
+- `XComGameState_Unit` triggers `SoldierRankName`, `SoldierShortRankName` and
+  `SoldierRankIcon` events that allow listeners to override the those particular
+  properties of a soldier's rank, i.e. rank name, short name and icon.
```
robojumper: Style nit: Can you link to #408? Just to be consistent.

Issue #426, opened 2019-02-20
Aim Assist blocking ChangeHitResultForX() in X2AbilitytoHitCalc_StandardAim


MrNiceUK:
`InternalRollForAbilityHit()` applies Aim Assist (`GetModifiedHitChanceForCurrentDifficulty()`) *after* `ChangeHitResultForAttacker()` and `ChangeHitResultForTarget()` have been applied. Due to the mechanics of how aim assist works, this means that if XCom rolled hit results are changed to miss results, or enemy miss results turned into hits, aim assist will revert the change. Hit/Miss being in the sense of what `IsHitResultMiss/Hit()` returns rather than `eHit_Miss` and `eHit_Success` specifically. This is true regardless of the actual aim assist modifiers for the current difficulty (including Legend where they are all zero).
The base game only uses `ChangeHitResultForX()` for Ranger's 'Reaper' ability and Templar 'Deflect/Reflect', which change XCom misses into hits or enemy hits into misses,  except with MC Corner cases. So they are unaffected by the issue. 'LightningReflexes' and 'CounterAttack' do not use these methods. Notably however, the hard coded `LightningReflexes' check in `InternalRollForAbilityHit()` is handled just *after* Aim Assist.
Moving the aim assist earlier will have no effect on Reaper, only a minor feedback change to Templar's Deflect, where attacks which might previous have shown as deflected may now be shown as misses, and a very minor change to 'Reflect' since rolled hits where are in the ranger to be changed to misses by Aim Assist, will not have a chance to be reflected.

PR #427, opened 2019-02-20
#426 Aim Assist / ChangeHitResultForAttacker() conflict resolution

MrNiceUK:
#426


robojumper:
Moving the `ChangeHitResultForX()` block below the Lightning Reflexes check does actually change the behavior: If I understand correctly, Effects can now bypass Lightning Reflexes whereas previously, Lightning Reflexes would win. Is this an intended change? Can we justify changing this?

Issue #419, opened 2019-02-10
Add New AbilityTagExpandHandler To Allow Reflection


VampireMonkey:
Currently **AbilityTagExpandHandler** in **X2DownloadableContentInfo** doesn't allow much dynamic handling and trying to expand the existing AbilityTag class without overriding it or getting deadlocked by its need for native(core) or how doing :
```
XComEngine(class'GameEngine'.static.GetEngine()).LocalizeContext.LocalizeTags.AddItem(new class'X2AbilityTag_Custom');
```
Does nothing.

I propose changing(in **X2AbilityTag.uc**) : 

```
//  allow DLC info to handle the tag
	DLCInfos = `ONLINEEVENTMGR.GetDLCInfos(false);
	foreach DLCInfos(DLCInfo)
	{
		if (DLCInfo.AbilityTagExpandHandler(InString, OutString))
			return;
	}

`RedScreenOnce("Unhandled localization tag: '"$Tag$":"$InString$"'");
OutString = "<Ability:"$InString$"/>";
```
to
```
//  allow DLC info to handle the tag
	DLCInfos = `ONLINEEVENTMGR.GetDLCInfos(false);
	foreach DLCInfos(DLCInfo)
	{
		if (DLCInfo.AbilityTagExpandHandler(InString, OutString))
		{
			return;
		}

		if (DLCInfo.AbilityTagExpandHandler_CH(InString, OutString, ParseObj, StrategyParseObj, GameState))
		{
			return;
		}
	}

`RedScreenOnce("Unhandled localization tag: '"$Tag$":"$InString$"'");
OutString = "<Ability:"$InString$"/>";
```
and adding another static function to **X2DownloadableContentInfo**

```
static function bool AbilityTagExpandHandler_CH(string InString, out string OutString, Object ParseObj, Object StrategyParseOb, XComGameState GameState)
```

Will preserve compatibility with existing mods and will allow new mods to use it.

Allowing this level of reflection will help make neat code that doesn't use a switch monster of the likes in vanilla **X2AbilityTag.uc**

Would also mean that you could dynamically change the display text of abilities in the tactical game.

_Ofcourse would do it myself but currently too deep in on my current mod to try to fix the highlander, can't run the game without it just crashing when changing anything._

robojumper:
[...]

PR #429, opened 2019-02-22
#419 Add New AbilityTagExpandHandler To Allow Reflection

VampireMonkey:
Can't seem to figure this out...


Musashi1584:
The pull request is also broken. Please rebase. Not going into 1.19.

PR #436, opened 2019-02-27
Improve covertactionrisk flexibility

### LINKED ISSUE #434: Covert action risks - improve flexiblity
https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/ba117b928f4acae9bac89d90da5abd3fe3d88f57/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc#L892-L903

* Change arguments to pass the risk name (make the first argument into full `CovertActionRisk` struct?) - it's a private function anyway
* Dispatch `CovertActionRisk_AlterChanceModifier` event to allow mods to change the modifier
* Add `XCGS_CA::RecalculateRiskChanceToOccurModifiers()` which will call `CalculateRiskChanceToOccurModifiers` on all currently existing risks

```
function EnableDarkEventRisk(name DarkEventRiskName)
{
	// ...

	// Only add or modify risks which are available
	if (RiskTemplate.IsRiskAvailableFn == none || RiskTemplate.IsRiskAvailableFn(GetFaction()))
	{
```
Add a `AllowDarkEventRisk` event to allow mods to overwrite whether the risk is applied or not

statusNone:
closes #434

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
@@ -985,7 +985,7 @@ function EnableDarkEventRisk(name DarkEventRiskName)
 		// If the Risk is not a part of the default template, add it
 		if (RiskNames.Find(DarkEventRiskName) == INDEX_NONE)
 		{
-			AddRisk(X2CovertActionRiskTemplate(Tuple.Data[0].o), bChosenIncreaseRisks, true);
+			AddRisk(RiskTemplate, bChosenIncreaseRisks, true);
 		}
 		else // Otherwise search through the existing risks to modify the chance to occur
 		{
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
@@ -889,6 +889,9 @@ private function AddRisk(X2CovertActionRiskTemplate RiskTemplate, bool bChosenIn
 	Risks.AddItem(NewRisk);
 }
 
+// start CHL issue #434
+// CHL function modified: first parameter changed from "int ChanceToOccur" to "CovertActionRisk ActionRisk"
+// also added event 'CovertActionRisk_AlterChanceModifier' to allow mod interation
 private function int CalculateRiskChanceToOccurModifiers(CovertActionRisk ActionRisk, bool bChosenIncreaseRisks, bool bDarkEventRisk)
 {
 	local int ChanceToOccurModifier;
@@ -919,6 +922,7 @@ private function int CalculateRiskChanceToOccurModifiers(CovertActionRisk Action
 	return Tuple.Data[4].i;
 }
 
+// CHL fuction added: to calculate risk chance on all risks for a given covert action
 function RecalculateRiskChanceToOccurModifiers()
 {
 	local XComGameState_HeadquartersResistance ResHQ;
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
```diff
@@ -883,47 +883,109 @@ private function AddRisk(X2CovertActionRiskTemplate RiskTemplate, bool bChosenIn
...
+	if (Tuple.Data[1].b)
 	{
 		RiskNames = GetMyTemplate().Risks;
 		bChosenIncreaseRisks = GetFaction().GetRivalChosen().ShouldIncreaseCovertActionRisks();
 
 		// If the Risk is not a part of the default template, add it
 		if (RiskNames.Find(DarkEventRiskName) == INDEX_NONE)
 		{
-			AddRisk(RiskTemplate, bChosenIncreaseRisks, true);
+			AddRisk(X2CovertActionRiskTemplate(Tuple.Data[0].o), bChosenIncreaseRisks, true);
```
Xymanek: I don't think it's a good idea to allow mods to replace the template - it's better to keep this line as it was

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
```diff
@@ -883,39 +883,101 @@ private function AddRisk(X2CovertActionRiskTemplate RiskTemplate, bool bChosenIn
...
+		if (RiskTemplate.IsRiskAvailableFn == none || RiskTemplate.IsRiskAvailableFn(GetFaction()))
+        {
+            bDarkEventRisk = false;
+            if (ResHQ.CovertActionDarkEventRisks.Find(Risk.RiskTemplateName) != INDEX_NONE)
+            {
+                bDarkEventRisk = true;
+            }
+		}
+
+		Risk.ChanceToOccurModifier = CalculateRiskChanceToOccurModifiers(Risk, bChosenIncreaseRisks, bDarkEventRisk);
```
Xymanek:
```
foreach Risks(Risk)
```
Foreach doesn't pass items by reference, so your changes to `Risk.ChanceToOccurModifier` get lost - you need a `for` loop

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
```diff
@@ -883,39 +883,101 @@ private function AddRisk(X2CovertActionRiskTemplate RiskTemplate, bool bChosenIn
...
+	Tuple.Data[4].kind = XComLWTVInt;
+	Tuple.Data[4].i = ChanceToOccurModifier;
+
+	`XEVENTMGR.TriggerEvent('CovertActionRisk_AlterChanceModifier', Tuple, self);
+
+	return Tuple.Data[4].i;
 }
 
+// CHL fuction added: to calculate risk chance on all risks for a given covert action
+function RecalculateRiskChanceToOccurModifiers()
```
robojumper: Since this is a public function (not used in XComGame at least), can you comment when mods are required/expected to call it?

robojumper:
Can you point out the places where changes occured in a more fine-grained manner, as is customary for CHL changes? Also, for the new public function, I was more thinking of something like 

    // Due to how this class internally caches chances, when anything happens that 
    // would cause a change in the return value for your listener for 
    // `CovertActionRisk_AlterChanceModifier`, you must call this function to make
    // the game aware of this and trigger the event to update the chances.

PR #446, opened 2019-03-11
improve X2SitRepEffect_GrantAbilities by adding an array<ETeam> Teams

### LINKED ISSUE #445: General improvement to X2SitRep_GrantAbilities Template
add the ability to decide which team(s) are granted the given ability

statusNone:
closes #445

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,7 @@ All notable changes to Vanilla 'War Of The Chosen' Behaviour will be documented 
 - Make suppression work with units that don't have a suppression specific
   idle animation animation set on them (#74)
 - Gremlins (and other Cosmetic Units) are now correctly tinted and patterned (#376)
+- Allow mods to decide which team(s) are granted an ability via X2SitRepEffect_GrantAbilities (#445)
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


robojumper:
I'm not opposed to these changes, but the function specifically says

    // If you need different/more precise control of which units get which abilities, you can override
    // this function with a subclass

What is the reason for still going with a HL PR?

Xymanek:
Copy paste from discord:

I do not see any benefit in multitude of `X2SitRepEffect_GrantAbilities` versions floating around
I think it makes more sense to expand the effect's options, especially for common flags (like `X2Condition_UnitProperty` does)
an additional benefit of this approach is that mods can change config of effect used by other mods - if each package has their own version this becomes impossible (except for compile dependency)
if we were to follow all suggestions  for mods in game's source, then i guess we should start MCO'ing XCGS classes https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/ba117b928f4acae9bac89d90da5abd3fe3d88f57/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc#L2092 => the suggestions are clearly not made with highlander availability in mid

PR #447, opened 2019-03-11
Allow custom status from mods for units

### LINKED ISSUE #322: Allow custom statuses from mods for units
The specific function to change would be `GetStatusStringsSeparate` in `XComGameState_Unit`, which conveniently is the one stop place the game has for knowing what kind of status string to deliver.

By default it just checks projects and covert actions, but an event trigger for a tuple to come back with a string when applicable would allow for mods with unique strategy functions for units to give proper statuses of units interacting with them.

statusNone:
closes #322

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -5137,11 +5137,36 @@ function string GetStatusString()
 	return FormattedStatus;
 }
 
+// start CHL issue #322
+// CHL function modified: added event 'CustomizeStatusStringSeparate'
 function GetStatusStringsSeparate(out string Status, out string TimeLabel, out int TimeValue)
 {
 	local bool bProjectExists;
 	local int iHours, iDays;
-	
+	local XComLWTuple Tuple;
+
+	Tuple = new class'XComLWTuple';
+	Tuple.Id = 'CustomizeStatusStringsSeparate';
+	Tuple.Data.Add(4);
+	Tuple.Data[0].kind = XComLWTVBool;
+	Tuple.Data[0].b = false;
+	Tuple.Data[1].kind = XComLWTVString;
+	Tuple.Data[1].s = Status;
+	Tuple.Data[2].kind = XComLWTVString;
+	Tuple.Data[2].s = TimeLabel;
+	Tuple.Data[3].kind = XComLWTVInt;
+	Tuple.Data[3].i = TimeValue;
+
+	`XEVENTMGR.TriggerEvent('CustomizeStatusStringsSeparate', Tuple, self);
+
+	if (Tuple.Data[0].b)
+	{
+		Status = Tuple.Data[1].s;
+		TimeLabel = Tuple.Data[2].s;
+		TimeValue = Tuple.Data[3].i;
+		return;
+	}
+
 	if( IsInjured() )
 	{
 		Status = GetWoundStatus(iHours);
@@ -5160,7 +5185,7 @@ function GetStatusStringsSeparate(out string Status, out string TimeLabel, out i
 		if (Status != "")
 			bProjectExists = true;
 	}
-	else if( IsDead() )
+	else if(IsDead())
 	{
 		Status = "KIA";
 	}
@@ -5182,6 +5207,7 @@ function GetStatusStringsSeparate(out string Status, out string TimeLabel, out i
 		TimeLabel = class'UIUtilities_Text'.static.GetDaysString(iDays);
 	}
 }
+// end CHL issue #322
 //-------------------------------------------------------------------------
 // Returns a UI state (color) that matches the soldier's status
 function int GetStatusUIState()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ All notable changes to Vanilla 'War Of The Chosen' Behaviour will be documented 
 - Triggers the event `Geoscape_ResInfoButtonVisible` to allow to override the visibility of resistance orders button in `UIStrategyMap_HUD` (#365)
 - Triggers the event `NumCovertActionsToAdd` to allow mods to modfiy number of Covert Actions (#373)
 - Triggers the event `CompleteRespecSoldier` when a training center soldier respec was completed. (#339)
+- Triggers the event `CustomizeStatusStringsSeparate` in XComGameState_Unit::GetStatusStringsSeparate (#322)
 
 ### Modding Exposures
 - Allows mods to add custom items to the Avenger Shortcuts (#163)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -5137,11 +5137,36 @@ function string GetStatusString()
 	return FormattedStatus;
```
pledbrook: Could we not also modify `GetStatusString()` above so that mods have some control over whether the time value and label are included?

pledbrook: Ignore the above comment as `GetStatusString()\ doesn't impact soldier lists.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -5137,11 +5137,36 @@ function string GetStatusString()
...
 }
 
+// start CHL issue #322
+// CHL function modified: added event 'CustomizeStatusStringSeparate'
 function GetStatusStringsSeparate(out string Status, out string TimeLabel, out int TimeValue)
```
pledbrook: Could we not update this method to include an extra, optional `out` parameter that indicates whether the time portion should be displayed or not? It would be quite an extensive change (some callers of the method would need to be updated), but I don't see what use customising the status string is if a listener can't also disable the display of the time.

pledbrook: Also note that LW2's highlander allowed listeners to override the colour for the status strings as well, although that's less important.

pledbrook: One more point: `UIUtilities_Strategy.GetPersonnelStatusSeparate()` only uses `GetStatusStringsSeparate()` for certain conditions. This leads to several situations in which listeners _can't_ override the status string. This is noticeable when selecting a soldier as a haven adviser. The listener attempts to set the status of rookies to "Rank Too Low", but the listener is never called in this instance because `GetPersonnelStatusSeparate()` just sets the status string to "Ready AVAILABLE".

I think `GetPersonnelStatusSeparate()` should probably _always_ go through `GetStatusStringsSeparate()`.

Xymanek:
> Could we not update this method to include an extra, optional out parameter that indicates whether the time portion should be displayed or not? It would be quite an extensive change (some callers of the method would need to be updated)

The issue is that any mod that uses that function will need to be updated, or the game will crash. In fact, because LW2 did something along those lines (afaik also in soldier status stuff), [ShowMeThePersonel](https://steamcommunity.com/sharedfiles/filedetails/?id=638033313) was incompatible with vanilla CHL

> One more point: `UIUtilities_Strategy.GetPersonnelStatusSeparate()` only uses `GetStatusStringsSeparate()` for certain conditions

Yes, that seems to be an issue for a few select cases

> I think `GetPersonnelStatusSeparate()` should probably always go through `GetStatusStringsSeparate()`

The issue here is that the 2 have different use cases. For example, `GetPersonnelStatusSeparate` handles the following case (while `GetStatusStringsSeparate` doesn't):

```uc
		if(`HQPRES != none &&  `HQPRES.ScreenStack.IsInStack(class'UISquadSelect') && GetXComHQ().IsUnitInSquad(Unit.GetReference()) )
		{
			Status = default.m_strOnMissionStatus;
			eState = eUIState_Highlight;
		}
```

I think inserting another event with override option into `GetPersonnelStatusSeparate` is a good way to do this, unless we want to re-architect the whole status reporting logic. We could also insert another event before 

```uc
	Status = class'UIUtilities_Text'.static.GetColoredText(Status, eState, FontSize);
	TimeLabel = class'UIUtilities_Text'.static.GetColoredText(TimeLabel, eState, FontSize);
	if( TimeNum == 0 && bHideZeroDays )
		TimeValue = "";
	else
		TimeValue = class'UIUtilities_Text'.static.GetColoredText(string(TimeNum), eState, FontSize);
```
to allow mods to modify colours and time display/hiding

Xymanek:
Also, I took a look at DSL and figured out why the hook was working for me - I had the mod disabled 🤦‍♂️  
Anyway, it seems that the only reason it has its own version of `GetPersonnelStatusSeparate` and `GetPersonnelStatusSeparate` is

> Show status in hours instead of days when below certain threshold, customizable in XComGame.ini

I feel like at this point this was done in so many mods (we even have that logic in stratoverhaul - [1](https://github.com/WOTCStrategyOverhaul/CovertInfiltration/blob/3ecd3b0705ed049edc30f4bf794d21f901af453b/CovertInfiltration/Src/CovertInfiltration/Classes/UIEventQueue_CovertActionListItem_CI.uc#L27-L28), [2](https://github.com/WOTCStrategyOverhaul/CovertInfiltration/blob/3ecd3b0705ed049edc30f4bf794d21f901af453b/CovertInfiltration/Src/CovertInfiltration/Classes/UIEventQueue_ListItem_CI.uc#L29-L30), [3](https://github.com/WOTCStrategyOverhaul/CovertInfiltration/blob/3ecd3b0705ed049edc30f4bf794d21f901af453b/CovertInfiltration/Src/CovertInfiltration/Classes/UIEventQueue_MultipleActions.uc#L85-L119)) that it would make sense to implement that in the CHL and just switch individual mods to the centralized interface

cc @bountygiver

pledbrook:
I'll need to look at the two in more depth, but at first glance there appears to be a lot of overlap between `GetPersonnelStatusSeparate()` and `GetStatusStringsSeparate()` when it comes to the `if` conditions.

I was really only thinking of the `else` part going through `GetStatusStringsSeparate()`. The `ON MISSION` is fine as it is, although one might want to override that as well (just LW2 doesn't need to).

Anyway, as I said, I need to look at those more closely to get a sense of how they really relate to one another. Then I'll be able to comment on your proposed new events.

MrNiceUK:
> The issue is that any mod that uses that function will need to be updated, or the game will crash.

IIRC what Robo said about it last time it came up, you can add optional parameters to the end, and mods which call it will still be fine (since the extra parameters are optional), *but* if the class is derived in a mod and implements the function, that leads to crashes due to the function def mismatch.

bountygiver:
I can update my mod to call the default function if highlander is detected, as we have a way to safely detect highlander now

robojumper:
@statusNone What's the status on this? @pledbrook pointed out in https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/447#discussion_r271194797 that there are still some issues with code duplication and that this doesn't fix some of the use cases.

Xymanek:
I've suggested possible solution in https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/447#issuecomment-478956276 and https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/447#issuecomment-478963714 - it's up to @pledbrook to decide what approach suits his needs (the current implementation works for stratoverhaul)

pledbrook:
Ugh. Requires too much thought. I'm sad that function overloading isn't supported as that would solve the backwards compatibility issue nicely.

Anyway, I think much though will be needed to devise enhancements that are backward compatible and provide comprehensive customisation options. I don't think this should PR should be held up.

Xymanek:
> I don't think this should PR should be held up by that.

I assume this means that you think that this PR should be merged. Imo, if we are to replace this functionality with a better implementation, then there is no reason to merge this PR

pledbrook:
I think my concerns would be handled by adding a new event to `GetPersonnelStatusSeparate()` as @Xymanek suggested. That's the method that adds colour status, deals with mental states, and determines whether time left should be displayed or not. That work should not block this PR.

Having said that, the only reason to override `GetStatusStringsSeparate()` (via an event) appears to be so that one can affect the status displayed in staff reassignment dialogs (when switching them to new staff slots I guess). I think everything else goes through `GetPersonnelStatusSeparate()`, although I can only search through CHL source at the moment, not the full SDK.

Xymanek:
We will still need to somehow account for days-to-hours, which is the reason that DSL used custom versions of those functions

pledbrook:
You mean allow overriding of when hours are shown instead of days? I was just thinking of a config variable for that.

Xymanek:
Ok, so i digged around for a bit and this is really convoluted but I think managed to find the most optimal solution - this pr/event stays as-is and in addition, the following events are going to be needed:

* `OverrideUnitStatusTime` in `GetStatusStringsSeparate` (a) right before the end of the function and (b) right before `return` inside `if (Tuple.Data[0].b)` to account for mod-set statuses. This is needed for days-to-hours functionality
*  `OverridePersonnelStatusSeparate` in `GetPersonnelStatusSeparate`, at the end. I think that the tuple should contain also non-coloured `Status`, `TimeLabel` and `TimeNum` so that it's easier for mods to do their changes
* Optionally, we could also add an event to `GetPersonnelStatus` so that mods can change the status for cases when it's used as a single string (eg. armory header, sorting in `UIPresonnel`, etc)

pledbrook:
I'm happy for this PR to be merged and the other events added in a separate PR.

On the `OverrideUnitStatusTime` event, that should probably be added to `GetMentalStateStringsSeparate()` as well.

pledbrook:
Out of interest, why was the event added to `GetStatusStringsSeparate()` rather than `GetPersonnelStatusSeparate()`? The latter covers more status strings and includes mental states.

PR #451, opened 2019-03-15
add a function to X2TacticalGameRuleset::CreateTacticalGame to allow …

### LINKED ISSUE #450: Give SitReps access to Tactical StartState
add a function to `X2TacticalGameRuleset::CreateTacticalGame` to allow SitReps to modify the StartState in the same way that Strategy Cards do

statusNone:
closes #450

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2SitRepEffect_ModifyTacticalStartState.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2SitRepEffect_ModifyTacticalStartState.uc
@@ -1,3 +1,13 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    XComGameState_ModifyTacticalStartState
+//  AUTHOR:  statusNone
+//           
+//  Class created to give SITREPs access to the Tactical XCGS StartState
+//
+//---------------------------------------------------------------------------------------
+// File added to the CHL this is a non-base game class
+//---------------------------------------------------------------------------------------
+
 class X2SitRepEffect_ModifyTacticalStartState extends X2SitRepEffectTemplate;
 
 var Delegate<ModifyTacticalStartStateDelegate> ModifyTacticalStartStateFn;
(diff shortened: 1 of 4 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2SitRepEffect_ModifyTacticalStartState.uc
```diff
@@ -0,0 +1,5 @@
+class X2SitRepEffect_ModifyTacticalStartState extends X2SitRepEffectTemplate;
```
robojumper: Maybe a comment or class name that indicates that this class is new?

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
```diff
@@ -2890,8 +2923,9 @@ Begin:
 	WorldInfo.MyLocalEnvMapManager.SetEnableCaptures(TRUE);
 
 	ReleaseScriptLog("X2TacticalGameRuleset: Finished Generating Map");
-
+	
 	ApplyResistancePoliciesToStartState( );
+	ApplySitRepEffectsToStartState();  // CHL issue #450
 	ApplyDarkEventsToStartState( );
```
robojumper: 1. There's a stray whitespace.
2. Order? Is there a reason this is between Resistance Policies and Dark Events?

Xymanek: The implementation was based on bringing functionality available for resistance orders but not for sitreps to latter. Considering that DEs/orders are already working and are done here i do not see a need to change the positioning/order.

If such need ever arises, we can just add another delegate to the template

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2SitRepEffect_ModifyTacticalStartState.uc
```diff
@@ -0,0 +1,15 @@
...
+//
+//---------------------------------------------------------------------------------------
+// File added to the CHL this is a non-base game class
+//---------------------------------------------------------------------------------------
+
+class X2SitRepEffect_ModifyTacticalStartState extends X2SitRepEffectTemplate;
+
+var Delegate<ModifyTacticalStartStateDelegate> ModifyTacticalStartStateFn;
+
+delegate ModifyTacticalStartStateDelegate(XComGameState StartState);
```
Xymanek: I think it makes sense to override the validate function to check that the delegate has been set

PR #453, opened 2019-03-16
Add task and scripts to automatically update version and include commit info in debug builds

### LINKED ISSUE #452: Enhance version display to show more info about dev builds
Right now, the version display always shows whatever version is in the script files and nothing more. There is no way to identify dev builds and no way to see which commit the HL was built from. Adding an optional `commit` field to all version info that's optionally added when not compiling with `-final_release` would help.

Problems: We have multiple supported build processes. 

* The PowerShell script is the most configurable and should be able to handle this easily. We could also extend it to keep all the version fields in sync.
* There also is the standard MSBuild process that uses the `.x2proj` file. The best course of action there is to add a pre-build step that's shared with the PS script. However, when cooking, we simply recommend invoking the compiler manually or with a `.cmd` script, so even `-final_release` builds might end up with that info -- consider adding a function to `CHHelpers` that returns whether XComGame is compiled with `-final_release` and only show the commit info if this is not the case.

robojumper:
Fixes #452. There is no good way to make this process automatic with MSBuild.

### DIFF
--- a/.scripts/build.ps1
+++ b/.scripts/build.ps1
@@ -234,6 +234,16 @@ Write-Host "Copying the mod's scripts to Src..."
 Copy-Item "$stagingPath\Src\*" "$sdkPath\Development\Src\" -Force -Recurse -WarningAction SilentlyContinue
 Write-Host "Copied."
 
+if ($final_release) {
+    Write-Host "Updating version..."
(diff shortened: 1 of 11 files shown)


Xymanek:
> There is no good way to make this process automatic with MSBuild

😭

Issue #321, opened 2018-10-30
Suggestion: Add an ini config option to disable the minimum 1 dmg through armor


Hannibalian:
I inquired about this on the subreddit awhile back and was told that the Highlander is probably the safest way to make adjustments to `X2Effect_ApplyWeaponDamage.uc`. The code to reference:

```
if (ArmorMitigation >= WeaponDamage)
ArmorMitigation = WeaponDamage - 1;
```

Ideally we could set this to 0 from the config, or to any value we wished.

/u/septichem proposed a potential solution [here](https://www.reddit.com/r/xcom2mods/comments/9iebo3/anybody_have_information_on_how_the_minimum_1/e6nk6il/).

```
// somewhere at the top of the file (probably next to the other config values)
// can config values have a default value in code?
var config bool ALWAYS_APPLY_MINIMUM_DAMAGE;

// ...

if ((ArmorMitigation >= WeaponDamage) && ALWAYS_APPLY_MINIMUM_DAMAGE)
  ArmorMitigation = WeaponDamage - 1;
```

Iridar:
Closing this because it was implemented in https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/747

PR #461, opened 2019-03-29
Add config option to disable minimum damage through armor

cannonfodder1:
Adds the config option ALWAYS_APPLY_MINIMUM_DAMAGE and a single change to an if-statement inside X2Effect_ApplyWeaponDamage. Tested in skirmish and works with guns, grenades, and shredding.
Closes issue #321

### DIFF
--- a/X2WOTCCommunityHighlander/Config/XComGameCore.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGameCore.ini
@@ -18,7 +18,7 @@
 ;+CritUpgradesThatDontUseEffects="MyFancyTemplateName" ;For modders to use to specify upgrades in their mods that use Issue #237 functionality
 
 ; Issue 321
-; Set to false (vanilla behaviour) if you want to still damage the target by 1 even if armor is more than incoming damage
-; Set to true if you want damage that is less than target armor to be completely neutralized
+; Set to false/commented out (vanilla behaviour) if you want to still damage the target by 1 even if armor is more than incoming damage
+; Set to true/uncomment it if you want damage that is less than target armor to be completely neutralized
 [XComGame.X2Effect_ApplyWeaponDamage]
-+NO_MINIMUM_DAMAGE=false
+;+NO_MINIMUM_DAMAGE=true
--- a/X2WOTCCommunityHighlander/Config/XComGameCore.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGameCore.ini
@@ -16,3 +16,9 @@
 
 [XComGame.X2AbilityToHitCalc_StandardAim]
 ;+CritUpgradesThatDontUseEffects="MyFancyTemplateName" ;For modders to use to specify upgrades in their mods that use Issue #237 functionality
+
+; Issue 321
+; Set to true (vanilla behaviour) if you want to still damage the target by 1 even if armor is more than incoming damage
+; Set to false if you want damage that is less than target armor to be completely neutralized
+[XComGame.X2Effect_ApplyWeaponDamage]
++ALWAYS_APPLY_MINIMUM_DAMAGE=true
--- a/X2WOTCCommunityHighlander/Config/XComGameCore.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGameCore.ini
@@ -18,7 +18,7 @@
 ;+CritUpgradesThatDontUseEffects="MyFancyTemplateName" ;For modders to use to specify upgrades in their mods that use Issue #237 functionality
 
 ; Issue 321
-; Set to true (vanilla behaviour) if you want to still damage the target by 1 even if armor is more than incoming damage
-; Set to false if you want damage that is less than target armor to be completely neutralized
+; Set to false (vanilla behaviour) if you want to still damage the target by 1 even if armor is more than incoming damage
+; Set to true if you want damage that is less than target armor to be completely neutralized
 [XComGame.X2Effect_ApplyWeaponDamage]
-+ALWAYS_APPLY_MINIMUM_DAMAGE=true
++NO_MINIMUM_DAMAGE=false
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
@@ -19,6 +19,9 @@ var bool    bIgnoreArmor;
 var bool	bBypassSustainEffects;
 var array<name> HideVisualizationOfResultsAdditional;
 
+// Issue #321
+var config bool ALWAYS_APPLY_MINIMUM_DAMAGE;
+
 // These values are extra amount an ability may add or apply directly
 var WeaponDamageValue EffectDamageValue;
 var int EnvironmentalDamageAmount;
@@ -1036,7 +1039,8 @@ simulated function int CalculateDamageAmount(const out EffectAppliedData ApplyEf
 				ArmorMitigation -= ArmorPiercing;				
 				if (ArmorMitigation < 0)
 					ArmorMitigation = 0;
-				if (ArmorMitigation >= WeaponDamage)
+				// Issue #321
+				if (ArmorMitigation >= WeaponDamage && default.ALWAYS_APPLY_MINIMUM_DAMAGE)
 					ArmorMitigation = WeaponDamage - 1;
 				if (ArmorMitigation < 0)    //  WeaponDamage could have been 0
 					ArmorMitigation = 0;    
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
@@ -20,7 +20,7 @@ var bool	bBypassSustainEffects;
 var array<name> HideVisualizationOfResultsAdditional;
 
 // Issue #321
-var config bool ALWAYS_APPLY_MINIMUM_DAMAGE;
+var config bool NO_MINIMUM_DAMAGE;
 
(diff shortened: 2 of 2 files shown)


X2WOTCCommunityHighlander/Config/XComGameCore.ini
```diff
@@ -15,4 +15,10 @@
...
 [XComGame.X2AbilityToHitCalc_StandardAim]
-;+CritUpgradesThatDontUseEffects="MyFancyTemplateName" ;For modders to use to specify upgrades in their mods that use Issue #237 functionality
\ No newline at end of file
+;+CritUpgradesThatDontUseEffects="MyFancyTemplateName" ;For modders to use to specify upgrades in their mods that use Issue #237 functionality
+
+; Issue 321
+; Set to true (vanilla behaviour) if you want to still damage the target by 1 even if armor is more than incoming damage
+; Set to false if you want damage that is less than target armor to be completely neutralized
+[XComGame.X2Effect_ApplyWeaponDamage]
++ALWAYS_APPLY_MINIMUM_DAMAGE=true
```
robojumper: Adding a by-default-true flag makes it harder for mods to opt out due to load order. Consider changing to something like `NO_MINIMUM_DAMAGE`, inverting the logic to allow retaining vanila behavior while not setting anything in config?

X2WOTCCommunityHighlander/Config/XComGameCore.ini
```diff
@@ -15,4 +15,10 @@
...
 [XComGame.X2AbilityToHitCalc_StandardAim]
-;+CritUpgradesThatDontUseEffects="MyFancyTemplateName" ;For modders to use to specify upgrades in their mods that use Issue #237 functionality
\ No newline at end of file
+;+CritUpgradesThatDontUseEffects="MyFancyTemplateName" ;For modders to use to specify upgrades in their mods that use Issue #237 functionality
+
+; Issue 321
+; Set to false (vanilla behaviour) if you want to still damage the target by 1 even if armor is more than incoming damage
+; Set to true if you want damage that is less than target armor to be completely neutralized
+[XComGame.X2Effect_ApplyWeaponDamage]
++NO_MINIMUM_DAMAGE=false
```
robojumper: Comment this out; it's false by default.

looplick:
I've confirmed that primary weapons still cause 1 damage regardless.  Total mitigation worked for melee, pistols, and grenades, but not for any weapon using the "StandardShot" ability (e.g. Assault Rifles).

Further troubleshooting found that the minimum 1 damage is being applied by abilities that use Template.AddTargetEffect(class'X2Ability_GrenadierAbilitySet'.static.ShredderDamageEffect());

I can't quite decipher yet how this is applying the minimum damage, but it looks like any weapon that can shred armor - even if the player does not have the Shredder ability - will cause minimum damage on a hit.

Any ideas?

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
```diff
@@ -1036,7 +1039,8 @@ simulated function int CalculateDamageAmount(const out EffectAppliedData ApplyEf
 				ArmorMitigation -= ArmorPiercing;				
 				if (ArmorMitigation < 0)
 					ArmorMitigation = 0;
-				if (ArmorMitigation >= WeaponDamage)
+				// Issue #321
+				if (ArmorMitigation >= WeaponDamage && !default.NO_MINIMUM_DAMAGE)
 					ArmorMitigation = WeaponDamage - 1;
```
MrNiceUK: The original code here actually served two purposes, it both allowed the minimum damage of 1, and stopped "healing" because armour exceeded weapon damage (since actual damage taken by the unit is WeaponDamage - ArmorMitigation, which should never be negative). In otherwords, just the -1 should be conditional on default.NO_MINIMUM_DAMAGE

looplick: Good point.  The code for ArmorMitigation >= WeaponDamage is bypassed completely when NO_MINIMUM_DAMAGE=true.  However, earlier in the function ArmorMitigation is defined as 

ArmorMitigation=kTarget.GetArmorMitigation(ApplyEffectParameters.AbilityResultContext.ArmorMitigation);

So it shouldn't matter.

MrNiceUK: Hmmm, a user on the Discord channel has said they've seen "Healing" from using the config (and not just a random user, but a "trustworthy" source!). I'll have a read through the code for how the values for `WeaponDamage` and `TotalDamage` are arrived at by that point in the code, see if there's an edge case which can cause healing...
***************** And hey, I saw you edit away that code shippet!

looplick: Lol.  Yeah I kind of hate how code looks so I deleted it...

looplick:
I can confirm that the ShredderDamageEffect() function in X2Ability_GrenadierAbilitySet is what's causing the minimum 1 damage from primary weapons 'StandardShot' ability.  How, I don't know yet.

MrNiceUK:
You tried it with *just* the CHL, no other mods? Since `X2Effect_Shredder` just changes one method from it's parent `X2Effect_ApplyWeaponDamage.uc`, and that *isn't* the method changed in this PR...

looplick:
Thanks for the tips! 

I've created a custom ability that is a one-for-one copy of 'StandardShot' and it replaces the normal 'StandardShot'.  It causes the minimum damage.  

When I remove the offending code from the custom ability (i.e. `Template.AddTargetEffect(class'X2Ability_GrenadierAbilitySet'.static.ShredderDamageEffect());`) 

and replace it with 
```
WeaponDamageEffect = new class'X2Effect_ApplyWeaponDamage';
Template.AddTargetEffect(WeaponDamageEffect);
```

The minimum damage is gone. I can still shred if I also add
```
ShredderEffect = new class'X2Effect_Shredder';
Template.AddTargetEffect(ShredderEffect);
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
MrNiceUK: Ok, found the flaw in the check:
```
TotalDamage = WeaponDamage - ArmorMitigation;

if ((WeaponDamage > 0 && TotalDamage < 0) || (WeaponDamage < 0 && TotalDamage > 0))
{
	// Do not allow the damage reduction to invert the damage amount (i.e. heal instead of hurt, or vice-versa).
	TotalDamage = 0;
}
```
alluded to by @looplick above. Basically, if `TotalDamage < 0` is only clamped to zero if `WeaponDamage > 0`. In practice, this can only happen if `WeaponDamage == 0`, and the change made by this PR means that `ArmorMitigation` wasn't clamped by WeaponDamage/0, and so `TotalDamage` is negative. Arguably the best fix would just be to check `TotalDamage < 0` with no condition on `WeaponDamage` at all, although the CHL principle of no unnecessary change to vanilla behaviour would point to what I referred to earlier, ie clamp `ArmorMitigation` by `WeaponDamage`, rather than `WeaponDamage -1`, if the `NO_MINIMUM_DAMAGE` flag is set.

looplick:
Thanks for the tips! 

I've created a custom ability that is a one-for-one copy of 'StandardShot' and it replaces the normal 'StandardShot'.  It causes the minimum damage.  

When I remove the offending code from the custom ability (i.e. `Template.AddTargetEffect(class'X2Ability_GrenadierAbilitySet'.static.ShredderDamageEffect());`) 

and replace it with 
```
WeaponDamageEffect = new class'X2Effect_ApplyWeaponDamage';
Template.AddTargetEffect(WeaponDamageEffect);
```

The minimum damage is gone. I can still shred if I also add
```
ShredderEffect = new class'X2Effect_Shredder';
Template.AddTargetEffect(ShredderEffect);
```

MrNiceUK:
`class'X2Ability_GrenadierAbilitySet'.static.ShredderDamageEffect())` is *exactly* the same as just `new class'X2Effect_Shredder'`. If you have that *as well* as `X2Effect_ApplyWeaponDamage`, then weapon damage will be applied twice! I say again, test with no mods other than CHL, since I think another mod you have active is replacing `X2Effect_Shredder`...

looplick:
I'm only running CHL and my testing mod.  Melee, pistols, grenades... any ability that uses  
```
WeaponDamageEffect = new class'X2Effect_ApplyWeaponDamage';
Template.AddTargetEffect(WeaponDamageEffect);
```
works properly.  Weapons that use `class'X2Ability_GrenadierAbilitySet'.static.ShredderDamageEffect())` 

don't.  That's what's confounding me.

looplick:
If I create a copy of X2Effect_Shredder (I called it X2Effect_MyShredder), that's a one-for-one copy, and replace ONLY
`Template.AddTargetEffect(class'X2Ability_GrenadierAbilitySet'.static.ShredderDamageEffect());` 

with 
```
	ShredderEffect = new class'X2Effect_MyShredder';
	Template.AddTargetEffect(ShredderEffect);
```

Everything works great!  

So, for some reason X2Effect_Shredder is not respecting the NO_MINIMUM_DAMAGE.

MrNiceUK:
Deeply mysterious. Just tested myself (just edited the config so conventional rifle and pistol deal just one damage, with 0 crit, no spread, and no plusone). The damage preview with `NO_MINIMUM_DAMAGE` set doesn't show the one damage through armour flashing on the unit HP bar(as it does without the flag), but the actual shot *does* deal one damage with the rifle (and none with the pistol). So why is there a discrepancy between the preview and the actual shot? I'll investigate more....

looplick:
It looks like this has been resolved :)

Thanks for that!

Issue #449, opened 2019-03-15
Allow mods to override the reinforcements alert and force its appearance


pledbrook:
Long War 2 uses the reinforcements alert to warn multiple turns in advance that they are on their way. It also uses colouring to indicate how long before those reinforcements arrive.

To support this, the highlander should allow mods to force the appearance of the alert, change its text and change its colour.

Iridar:
The PR that was supposed to close this issue has been merged to the live version of the Highlander. It remains open because it needs docs. Please add some :)

Issue #448, opened 2019-03-15
Allow mods to disable reinforcement flares and other RNF visualisations


pledbrook:
The visuals that show where reinforcements will spawn can make handling them too easy for players. So allow mods to disable those visuals.

Iridar:
The PR that was supposed to close this issue has been merged to the live version of the Highlander. It remains open because it needs docs. Please add some :)

PR #462, opened 2019-03-30
Allow mods greater control over reinforcement visuals

pledbrook:
This pull request implements issues #448 and #449, allowing mods control of whether reinforcement flares are visible, when the reinforcements alert is displayed, and what the alert looks like (its text and colour at least).

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIObjectiveList.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIObjectiveList.uc
@@ -1,0 +1,612 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    UIObjectiveList.uc
+//  AUTHOR:  Brit Steiner --  11/13/2014
+//  PURPOSE: This is an autoformatting list of objectives. . 
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+
+class UIObjectiveList extends UIPanel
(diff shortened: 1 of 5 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UIObjectiveList.uc
```diff
@@ -0,0 +1,612 @@
...
+	foreach AssociatedGameState.IterateByClassType(class'XComGameState_AIReinforcementSpawner', AISpawnerState)
+	{
+		break;
+	}
+
+	// Start Issue #449
+	ForceShowReinforcementsAlert = IsReinforcementsAlertForced();
+
+	// if this state has nothing for us to update, then just return
+	if (ObjectiveList == none && AISpawnerState == none && !ForceShowReinforcementsAlert)
```
robojumper: Small question: This is an optimization to skip updating if nothing of relevance has changed in the game state. But `ForceShowReinforcementsAlert` can go from false to true or from true to false. Only the former case will cause an update if I understand this right. How does LWOTC deal with this? Is there never a situation where `ForceShowReinforcementsAlert` can turn off without the other conditions changing?

pledbrook: I don't think this is an issue. If `objectiveList` is `none`, then it doesn't need repositioning, so it doesn't matter if `ForceShowReinforcements` is now `false` after earlier being `true`. As soon as the objective list exists again, it gets repositioned based on whether `ForceShowReinforcements` is `true` or `false`.

Issue #463, opened 2019-04-03
Allow mods to control how many objectives are spawned for a mission


pledbrook:
`XComTacticalMissionManager.SelectObjectiveSpawns()` chooses a random number between the mission's defined min and max number of objectives. But there may be situations in which a mod needs to specify a precise number. An example is Jailbreak in LW2, which needs to know exactly how many rebels and soldiers to spawn.

This would basically involve a port of an old LW2 highlander change.

Iridar:
I see this issue mentioned in comments in Highlander's code, though not sure which PR added it. In any case, please add docs so this issue can be closed.

Xymanek:
The PR is https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/464

PR #464, opened 2019-04-03
Allow mods to control number of objective spawns (#463)

pledbrook:
This change adds an event trigger to `XComTacticalMissionManager` that allows
listeners to override the number of objective spawns for a mission. This is
important if a mod needs to sync the number of objective spawns with some
other data, such as the rewards for the mission.

The event takes the form of:

```
{
  ID: OverrideObjectiveSpawnCount,
  Data: [ in BattleData BattleData, out int SpawnCount]
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,7 +86,9 @@ All notable changes to Vanilla 'War Of The Chosen' Behaviour will be documented 
 - `OverrideVictoriousPlayer` to allow override whether a mission was successful or not (#266)
 - `OverrideItemSoundRange` to allow overriding an item's sound range (#363)
 - `OverrideHackingScreenType` and `HackIn2D` to allow hacking using a 2D movie and using
-  the Skulljack / ADVENT screen arbitrarily (#330)
+- Allow mods to override the number of objectives spawned for a mission via the new event
+  `OverrideObjectiveSpawnCount`, which is triggered as the objective spawns are being
+  selected by `XComTacticalMissionManager`. (#463)
 
 ### Configuration
 - Added ability to modify default spawn size (#18)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,7 @@ All notable changes to Vanilla 'War Of The Chosen' Behaviour will be documented 
 - `OverrideVictoriousPlayer` to allow override whether a mission was successful or not (#266)
 - `OverrideItemSoundRange` to allow overriding an item's sound range (#363)
 - `OverrideHackingScreenType` and `HackIn2D` to allow hacking using a 2D movie and using
+  the Skulljack / ADVENT screen arbitrarily (#330)
 - Allow mods to override the number of objectives spawned for a mission via the new event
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -86,7 +86,9 @@ All notable changes to Vanilla 'War Of The Chosen' Behaviour will be documented
 - `OverrideVictoriousPlayer` to allow override whether a mission was successful or not (#266)
 - `OverrideItemSoundRange` to allow overriding an item's sound range (#363)
 - `OverrideHackingScreenType` and `HackIn2D` to allow hacking using a 2D movie and using
-  the Skulljack / ADVENT screen arbitrarily (#330)
+- Allow mods to override the number of objectives spawned for a mission via the new event
+  `OverrideObjectiveSpawnCount`, which is triggered as the objective spawns are being
+  selected by `XComTacticalMissionManager`. (#463)
```
robojumper: Not sure you meant to axe half of the other line... 😄

pledbrook: That's my elite Vim ninja skills at work

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComTacticalMissionManager.uc
```diff
@@ -1090,8 +1090,42 @@ private function array<ObjectiveSpawnPossibility> SelectObjectiveSpawns(Objectiv
...
+    // If we have battle info with a mission ID, see if any mods want
+	// to specify the number of objectives to use by passing this
+	// BattleData to them via an event.
+	//
+	// The event takes the form:
+	//   {
+	//       ID: OverrideObjectiveSpawnCount,
+	//       Data: [ in BattleData BattleData, out int SpawnCount ]
+	//   }
+    if (BattleData != none && BattleData.m_iMissionID > 0)
```
robojumper: I assume that this check is for distinguishing strategy/tql? Is there a specific reason to have the check here instead of in listener code?

pledbrook: That I do not know. I was wondering the same thing. I don't know enough about the code to decide one way or another. That's just what's in the existing LW2 implementation. It's perhaps just a convenience for listeners that they don't have to perform the check themselves.

pledbrook: @tracktwo is probably the person you should be asking :smile:

tracktwo: It was to distinguish strategy from TQL yes. It could easily be moved to the listener and just have those handle receiving `none` if that's the preferred approach in CHL.

robojumper: I don't have any strong feelings about this but people always get grumpy when things work in a normal campaign but stop working in TQL 😅

PR #468, opened 2019-04-05
Adding support for X2AbilityToHitCalc_StatCheck

### LINKED ISSUE #467: give X2AbilityToHitCalc_StatCheck::GetHitChance the ability to apply modifiers
give mods the ability to adjust the outcome of X2AbilityToHitCalc_StatCheck::GetHitChance via modifiers

statusNone:
closes #467


Xymanek:
Just in case you missed it on discord: you pulled a lot of unrelated stuff from aim into stat check, it shouldn't be there

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityToHitCalc_StatCheck.uc
```diff
@@ -0,0 +1,401 @@
...
+	{
+		AddModifier(ShotModifier.Value, ShotModifier.Reason, m_ShotBreakdown, ShotModifier.ModType, bDebugLog);
+	}
+	// end issue #467
+
+	AddModifier(BaseValue, GetBaseString(), m_ShotBreakdown, eHit_Success, bDebugLog);
+	AddModifier(AttackVal, GetAttackString(), m_ShotBreakdown, eHit_Success, bDebugLog);
+	AddModifier(-DefendVal, GetDefendString(), m_ShotBreakdown, eHit_Success, bDebugLog);
+	// start issue #467: modified the method of getting a final result from vanilla
+	FinalizeHitChance(m_ShotBreakdown, bDebugLog);
```
MrNiceUK: That's overkill, all you really needed to do was
`m_ShotBreakdown.FinalHitChance = m_ShotBreakdown.ResultTable[eHit_Success];`
instead of the original `m_ShotBreakdown.FinalHitChance = TargetRoll;`
`FinalizeHitChance()` does do that, but it does other busy work which is only applicable if Crit and/or Graze are in play.

MrNiceUK: Hmmm, `AddModifier()` *already* does that to keep the `FinalHitChance` in sync with ResultTable[eHit_Success], so as long as the result table is only changed via `AddModifier()`, as seems to be the case here, then there is no need at all to adjust `FinalHitChance` manually?

Issue #476, opened 2019-04-11
Make AbilityRequiresTargetingActivation()/IsTargetingMethodActivated() use a configurable list


MrNiceUK:
(Controller Support)

UITacticalHUD_AbilityContainer.AbilityRequiresTargetingActivation does this:
```
if (ClassIsChildOf(AbilityState.GetMyTemplate().TargetingMethod, class'X2TargetingMethod_Grenade') ||
	ClassIsChildOf(AbilityState.GetMyTemplate().TargetingMethod, class'X2TargetingMethod_Cone') ||
	ClassIsChildOf(AbilityState.GetMyTemplate().TargetingMethod, class'X2TargetingMethod_Line') ||
	ClassIsChildOf(AbilityState.GetMyTemplate().TargetingMethod, class'X2TargetingMethod_BlasterLauncher'))
{
	return true;
}
```
`IsTargetingMethodActivated()` does similar. These functions indicates that after selecting the ability, you have to press A again to aim the ability, which allows the user to rotate the camera with the buttons that would normally cycle the ability selection.
However, this skips some TargetingMethods -- RocketLauncher, for example is a Grenade, but grappling and BlazingPinions are not covered by the list. Making this list configurable would allow a more consistent behaviour.

mburbea:
I think it makes sense to fold the useful missing ones Grappling, Blazing Pinions, and Pathtarget into the highlander.

PR #477, opened 2019-04-11
Issue #476 (making which Targeting Methods need activation for controllers configurable)

MrNiceUK:
Resolution for #476. Should we add obvious "missing" targeting methods, or leave that to individual users/other mods?

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
@@ -110,6 +110,10 @@ var config(Content) array<name> EyeMaterial;
 var config(Content) array<name> FlagMaterial;
 // End Issue #356
 
+// Start Issue #476
+var config array<name> RequiresTargetingActivation;
+// End Issue #476
+
 // Start Issue #123
 simulated static function RebuildPerkContentCache() {
 	local XComContentManager		Content;
@@ -351,3 +355,20 @@ static function XComLWTuple BuildDefaultTuple(XComGameState_Unit UnitState)
 	return Tup;
 }
 // End Issue #257
+
+// Start Issue #476
+static function bool TargetingClassRequiresActivation(class<X2TargetingMethod> TargetingClass)
+{
+	local int i;
+
+	for (i = 0; i < default.RequiresTargetingActivation.Length; i++)
+	{
+		if (TargetingClass.IsA(default.RequiresTargetingActivation[i]))
+		{
+			return true;
+		}
+	}
+
+	return false;
+}
+// End Issue #476
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
@@ -357,13 +357,13 @@ static function XComLWTuple BuildDefaultTuple(XComGameState_Unit UnitState)
 // End Issue #257
 
 // Start Issue #476
-static function bool TargetingClassRequiresActivation(class<X2TargetingMethod> TargetingClass)
+static function bool TargetingClassRequiresActivation(X2TargetingMethod TargetingMethod)
 {
 	local int i;
 
 	for (i = 0; i < default.RequiresTargetingActivation.Length; i++)
 	{
-		if (TargetingClass.IsA(default.RequiresTargetingActivation[i]))
+		if (TargetingMethod.IsA(default.RequiresTargetingActivation[i]))
 		{
 			return true;
 		}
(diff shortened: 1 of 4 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
```diff
@@ -351,3 +355,20 @@ static function XComLWTuple BuildDefaultTuple(XComGameState_Unit UnitState)
...
 // End Issue #257
+
+// Start Issue #476
+static function bool TargetingClassRequiresActivation(class<X2TargetingMethod> TargetingClass)
+{
+	local int i;
+
+	for (i = 0; i < default.RequiresTargetingActivation.Length; i++)
+	{
+		if (TargetingClass.IsA(default.RequiresTargetingActivation[i]))
```
mburbea: This comparison will always be false `TargetingClass` extends class. You need to use `ClassIsChildOf` and dynamically load the object. I haven't a clue about how to cache the loaded objects.

See my changes where I got this actually working.
https://github.com/mburbea/FixControllerGrappling/blob/master/FixControllerGrappling/Src/FixControllerGrappling/Classes/FCG_UITacticalHUD_AbilityContainer.uc#L11

robojumper: > I haven't a clue about how to cache the loaded objects.

You don't. Loading is cheap because the object already exists.

> This comparison will always be false TargetingClass extends class.

I don't believe this is true. Much of the base game code (and code I have written) makes use of `ClassIsChildOf`/`IsA` working like this. The only thing that doesn't work with subclasses is a class/name comparison à la `_.Class == class'_'` or `_.Class.Name == _`.

mburbea: @robojumper
I originally had my code like this, with `TargetingMethod.IsA(...)` and it wasn't working.

The callsite in `AbilityRequiresTargetingActivation` is passing a class instance which fails the `IsA` test, but succeeds using the `ClassIsChildOf` test.

robojumper: You will have to provide the code that shows `IsA` isn't working while ClassIsChildOf is. Common usage[1], documentation[2] and prior experience all indicate that it respects child classes.

[1] https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/ba117b928f4acae9bac89d90da5abd3fe3d88f57/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc#L629-L639

[2] https://api.unrealengine.com/udk/Three/MasteringUnrealScriptFunctions.html#IsA(Name%20_ClassName)

mburbea: This is not an issue of subtyping. I agree with you that `IsA` should work. However, the instance being passed is actually a `class<Class>` not a `class<X2TargetingMethod>`. I'm assuming unrealscript uses some form of type erasure here so it basically doesn't care that the target invocation is wrong. 
Since, I think the issue is that maybe I don't speak the lingo here's my translation to C# of the problem.
```csharp
Type type = targetingMethod.GetType();

var isA = type is X2TargetingMethodGrenade; // always false as type is an instance of Type.
var childClassOf = type.IsSubClassOf(typeof(X2TargetingMethodGrenade)));
```

I'm no expert, but I had the code like that in my initial commit to github, and it broke ActivatedTargeting for everything including grenades.
You can see my older revision here:
https://github.com/mburbea/FixControllerGrappling/commit/a5f94beac7047033396d9f9fb93015fd08c47acb.

Does this make sense?

robojumper: > However, the instance being passed is actually a class<Class> not a class<X2TargetingMethod>

Aaaaah, there was my mistake. You are correct, a `class<X2TargetingMethod>` is actually of type `Class`, which extends `Object`. If we had an instance of a targeting method, `IsA` would work.

@MrNiceUK Mind fixing this?

MrNiceUK: @mburbea thanks for that! I might fix the issue the other way around though. As you will have noticed, I wanted to unify the two checks done previously, one of which starts with the class, the other with an instance, so I settled for passing the class, since that's a property of the instance. But I'll flip it to always taking an instance, and pass the CDO in the case where you start with just the class

mburbea: Forgive my lack of knowledge but what's a `CDO`? 
I was hoping to cache the result of DynamicLoadObject, but I couldn't figure out how to do that on the static property. I could assign it as a property of the current abilityContainer, but I didn't know the re-usability of such a thing.

MrNiceUK: @mburbea "Class Default Object", it's the default instance of the class. You indirectly interact with it whenever you use `.static` or `.default`, but you can get a direct reference to it, but you shouldn't mess around with it too much! Just look at my most recent commit. If you look in `Engine` and `Xcomengine`, you can also get the CDO by name or string as opposed to a class reference.

Issue #485, opened 2019-04-24
allow additional ambush risk sources


abatewongc:
title. Currently, the UI does not support new ambush types:
```
function bool HasAmbushRisk()
{
    local name RiskName;
    local int idx;

    for (idx = 0; idx < Risks.Length; idx++)
    {
        RiskName = Risks[idx].RiskTemplateName;
        if (RiskName == 'CovertActionRisk_Ambush' && NegatedRisks.Find(RiskName) == INDEX_NONE)
        {
            return true;
        }
    }

    return false;
}```

PR #487, opened 2019-04-24
multiple ambush risk sources attempt 2

abatewongc:
#485


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
```diff
@@ -2183,4 +2190,11 @@ function int GetMinDaysToComplete()
...
 {
 	return `ScaleStrategyArrayInt(GetMyTemplate().MaxActionHours);
-}
\ No newline at end of file
+}
+
+// Start Issue #485
+defaultproperties
+{
+	AmbushMissionSource="MissionSource_ChosenAmbush"
```
robojumper: Is this backwards compatible? I would expect existing campaigns to now have an empty source. Can you change uses of `AmbushMissionSource` to instead check for an empty name?

Issue #493, opened 2019-04-28
REACTION_DASHING_FINALMOD doesn't work


MrNiceUK:
`X2AbilityToHitCalc_StandardAim::GetReactionAdjust()` is bugged and will never return `REACTION_DASHING_FINALMOD`.
```
OldTarget = XComGameState_Unit(History.GetGameStateForObjectID(Target.ObjectID, eReturnType_Reference, History.GetCurrentHistoryIndex() - 1));

//  Add penalty if the target was dashing. Look for the target changing position and spending more than 1 action point as a simple check.
if (OldTarget != None && OldTarget.TileLocation != Target.TileLocation)
{
	if (OldTarget.NumAllActionPoints() > 1 && Target.NumAllActionPoints() == 0)
	{		
		return default.REACTION_DASHING_FINALMOD;
	}
}
return default.REACTION_FINALMOD;
```
This code is flawed, because reaction fire, since it interrupts the ability it is reacting to, means that the ability is completed in more than one game state. Costs are only applied in the final stage, so at the point the reaction fire HitCalc is used, the cost hasn't been applied. Therefore to correctly determine if the unit being targeted is "dashing" or not, the `XComGameState_Ability` for what is being reacted to must be consulted. Also the GameState for this will not always be the most recent submitted gamestate, so it is necessary to iterate back to find it.

PR #494, opened 2019-04-28
#493 Dashing Reaction

MrNiceUK:
Fix for #493


MrNiceUK:
Force Push because I left in one `` `log()`` from testing...

Issue #503, opened 2019-05-03
Fix pathing for flying units


pledbrook:
Flying units that are pod leaders (solo drones in LW2 are a common example) often fail to patrol due to a problem in the logic that determines valid paths for units. There is a very simple fix that can be applied from the LW2 highlander.

PR #505, opened 2019-05-03
Fix pathing issue for flying units (#503)

pledbrook:
Pods that have a "flying" unit (like an Archon or drone) were not
patrolling properly because the logic for calculating the tile paths was
treating any non-flat tile as impassable.

This fix comes straight from the original LW2 and has been tested with
ADVENT and drone patrols.


pledbrook:
I would wait on merging this until it's had a fair bit of testing in LW2 [WOTC].

Issue #511, opened 2019-05-09
Create a DLCInfo loadorder system


Musashi1584:
By adding 
```
function array<string> GetRunBeforeDLCIdentifiers();
function array<string> GetRunAfterDLCIdentifiers();
```
we create directed acyclic graph (dag) of mod dependencies.

Implement in a mod it would look like

```
[ModPackage.X2DownloadableContentInfo_MyMod]

DLCIdentifier="MyMod"
+RunBefore="OtherMod"
+RunAfter="AnotherMod"
```
```
class X2DownloadableContentInfo_MyMod extends X2DownloadableContentInfo config(ModConfig);
var config array<string> RunBefore;
var config array<string> RunAfter;

function array<string> GetRunBeforeDLCIdentifiers()
{
	return default.RunBefore;
}

function array<string> GetRunAfterDLCIdentifiers()
{
	return default.RunAfter;
}

```
With subclassing `XComOnlineEventMgr` and intercepting `GetDLCInfos` it is possible to reorder the dag of DLCInfo classes using topological sorting.

This results effectively in a order control for dlcinfo hooks like `OnPostTemplatesCreated`.

Xymanek:
> With subclassing `XComOnlineEventMgr`

This is effectively an MCO, which CHL is supposed to avoid. Is this something we are willing to break in this case?

Musashi1584:
not really, its implemented by

XComEngine
```
defaultproperties
{
	OnlineEventMgrClassName = "XComGame.CHOnlineEventMgr"
}
```

Xymanek:
This is still a class-level replacement, effectively an MCO, regadless of how it's implemented. If there is another mod out there that MCOs `XComOnlineEventMgr`, it will break

robojumper:
> This results effectively in a order control for dlcinfo hooks like OnPostTemplatesCreated.

Native to native function calls cannot be intercepted by script. This would work for all non-native hooks, but we can change the order there without creating a new OnlineEventMgr already. Maybe investigate whether the OnlineEventManager maintains an array of X2DownloadableContentInfo objects you can sort, and whether we have a good place to do it (after bootstrapping the OnlineEventMgr, before creating templates)?

Musashi1584:
Lets take this to the discord guys.

Xymanek:
[...]

PR #512, opened 2019-05-09
Issue #511 DLCinfo RunOrder

Musashi1584:


MrNiceUK:
Might be a bit late to do anything about it now, but to my ears it should be "Run Order" not "Load Order", which would affect the method names and enums, to be more clear about what this PR actually does.

PR #515, opened 2019-05-12
Allow the git commit versioning to be used when the CHL is a submodule

Xymanek:
Also fixes the commit field which [was broken in 1.18rc](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/commit/b9bd25d416760648b572c3f6e095531ac0ff48b6#diff-fbe217507dcd06b27b5bf1f239e1aca9L10)


Xymanek:
I've also introduced caching to prevent using commit hash from triggering a full package recompile

PR #517, opened 2019-05-12
Fix missing return value causing missing default mission intros

### LINKED ISSUE #516: v1.18.0 skips Skyranger and other cutscenes
Hi all,

First of all thanks for making this mod.

Apparently, v1.18.0 of the Highlander skips the Skyranger drop-in cutscenes and also skips the story-related cutscenes (like in the Lost & Abandoned mission).

Tested with a clean a install running only the Highlander. Attached is my launch.log:
[log.txt](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/files/3170489/log.txt)

Downgraded to v1.17.0 and did not experience this behavior.

Going off a hunch I think it's related to #395, but I ain't no expert.

Hope you will resolve this soon!

robojumper:
Closes #516.


MrNiceUK:
Looks good to me!

PR #523, opened 2019-05-21
Add support for yellow alert reflex actions

pledbrook:
These are a bunch of changes from the old LW2 highlander that enable yellow alert behaviour. See individual commits for details of the changes.


X2WOTCCommunityHighlander/Config/XComGame.ini
```diff
@@ -33,3 +33,6 @@ bEnableVersionDisplay=true
 +RequiresTargetingActivation=X2TargetingMethod_BlasterLauncher
 +RequiresTargetingActivation=X2TargetingMethod_BlazingPinions
 +RequiresTargetingActivation=X2TargetingMethod_Grapple
+
+; Range (in meters) of the 'AlertedByYell' alert produced by fleeing civilians through the DoNoiseAlert action.
++NoiseAlertSoundRange=27
```
pledbrook: Perhaps this needs to be set to a very high number to mimic vanilla behaviour?

pledbrook:
@Musashi1584 I would remove this from 1.19.0. Needs more testing.

MrNiceUK:
Rather than relying on a "large" alert range in config to replicate original behaviour, maybe just have `GetUnitsInRangeOnTeam()` allow all units if the range is zero? Somewhat similar to other config variables where the unconfigured "0" value is treated as "always", like in the effect apply chance or the percent hitcalc.

PR #532, opened 2019-05-31
#531 Bending reed Fix

### LINKED ISSUE #531: Fix Assassin's Bending Reed mobility boost
Bending Reed gives the Assassin a 50% boost to mobility after a melee attack so that she can run quite a long way away after attacking a someone. But the implementation is currently bugged such that the mobility bonus is retained for the rest of the mission. And then every additional melee attack stacks up the mobility! LWOTC players have seen her with 60+ mobility.

The issue appears to be in `X2Ability_ChosenAssassin.CreateBendingReed()`. The persistent effect is created with 1 turn and not infinite duration. But Mr. Nice said that the watch rule also needed to be set. I don't really understand what effect these arguments to `BuildPersistentEffect()` have, so I'll leave it to others to confirm what the solution should be.

MrNiceUK:
#531 Not much to say...


Xymanek:
Slight nitpick: from `README`:

> We should be able to find every line of code related to an issue by searching for "Issue #928" (or whatever) in the codebase

you omitted the word `Issue`. Please include as it really helps when searching the code

Issue #539, opened 2019-06-02
Provide event hooks for end-of-month processing


pledbrook:
This would allow mods to control supply income and perform any other work that they want.

PR #542, opened 2019-06-03
End of month events (#539)

pledbrook:
See second commit for details of the events added.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersResistance.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersResistance.uc
@@ -321,6 +321,12 @@ function OnEndOfMonth(XComGameState NewGameState)
 	XComHQ = XComGameState_HeadquartersXCom(History.GetSingleGameStateObjectForClass(class'XComGameState_HeadquartersXCom'));
 	XComHQ = XComGameState_HeadquartersXCom(NewGameState.ModifyStateObject(class'XComGameState_HeadquartersXCom', XComHQ.ObjectID));
 
+	// Start Issue #539
+	//
+	// Notify listeners of the start of end-of-month processing.
+	`XEVENTMGR.TriggerEvent('PreEndOfMonth', self, self, NewGameState);
+	// End Issue #539
+
 	bEndOfMonthNotify = true;
 	GiveSuppliesReward(NewGameState);
 	SetProjectedMonthInterval(`STRATEGYRULES.GameTime);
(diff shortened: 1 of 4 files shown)


Xymanek:
What's the point of `Monthly Report Closed`? Why not just use a UISL?

pledbrook:
I don't know what the original reason was. I pretty much just ported as is in terms of the number of events. One thing this approach has in its favour is consistency in how and where the listeners are implemented. It makes it easy for mods to keep all their end-of-month processing together.

pledbrook:
And if the event were to stay, it should probably be renamed to `PostEndOfMonth` for consistency with the first one.

pledbrook:
I've justified to myself that the 'PostEndOfMonth' event is worth it because:

 * It balances nicely with 'PreEndOfMonth'
 * It's association with the monthly report screen is coincidental
 * It's better for mod authors if they have all the events they need provided in a consistent way - telling them to use a screen listener for this one aspect seems off
 * It allows mod authors to put all their end-of-month processing code in one place

pledbrook:
@Musashi1584 I don't know if this has the green light to go. @Xymanek Any more feedback?

Xymanek:
I kinda fail to understand what `NegativeMonthlyIncome` does. Perhaps a better name is in order?

pledbrook:
I'll take another look at NegativeMonthlyIncome. I think it's combining two jobs into one, which makes it a bit confusing.

pledbrook:
@Musashi1584 Did this get merged? If so, my mistake. Should have marked it as WIP.

pledbrook:
@Xymanek I've split the 'NegativeMonthlyIncome' event into two. Hopefully the purpose of each of them is much clearer now.

pledbrook:
I made a boo boo, which meant the 'OverrideSupplyDrop' event was never triggered. Merged the change into the previous commit and force pushed, since I don't want the other two commits squashed during the merge.

pledbrook:
@Xymanek Was the change sufficient?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersResistance.uc
```diff
@@ -397,32 +403,83 @@ function GiveSuppliesReward(XComGameState NewGameState)
...
+	Tuple.Data[0].i = SupplyValue;  // The amount of negative supplies
+
+	`XEVENTMGR.TriggerEvent('ProcessNegativeIncome', Tuple, self, NewGameState);
 }
+// End Issue #539
 
 //---------------------------------------------------------------------------------------
-function int GetSuppliesReward(optional bool bUseSavedPercentDecrease)
+// Issue #539: Added the optional NewGameState argument
+function int GetSuppliesReward(optional bool bUseSavedPercentDecrease, optional XComGameState NewGameState)
```
Xymanek: This breaks BC policy.

If any mod MCOs `XComGameState_HeadquartersResistance` and implements this function, this change will cause a hard crash

pledbrook: Good catch. I've restored backwards compatibility by introducing `GetSuppliesReward_CH()` as the new function that takes a game state argument.

Issue #245, opened 2018-07-17
Add a DLCInfo hook to XGWeapon:Init


Musashi1584:
This allows mods to change weapon actor properties like `CustomUnitPawnAnimsets` or `DefaultSocket` conditionally.

PR #545, opened 2019-06-04
Fix for Issue #245

Musashi1584:
Moves the dlc info calls from `XGWeapon.Init` to `XGWeapon.CreateEntity`.

Fixes an issue where weapons where in a unpatched state on tactical begin in Covert Infiltration missions that create cosmetic attachements for the misssion loading screen.

Some code paths involving creating cosmetic attachments apparently directly call `XGWeapon.CreateEntity` or `XGWeapon.CreateEntityFromWeaponState` bypassing `XGWeapon.Init` and the dlc info calls.


Xymanek:
Just to clarify: this issue triggers when `return`ing `true` from `LoadingScreenOverrideTransitionMap` (#388)

Issue #548, opened 2019-06-05
Make sure civilians run from XCOM on unconcealed missions


pledbrook:
If XCOM starts a mission unconcealed with hostile civilians on it, those civilians will just walk away from XCOM soldiers when they come near rather than run. Not only does that slow down the action, but it also seems really weird and breaks immersion hard.

pledbrook:
This was implemented in LWOTC itself.

Xymanek:
How did you implement it?

pledbrook:
A [template modification](https://github.com/long-war-2/lwotc/commit/ecef266f1f3ce16d0220e10f2812302765b4fea1) of the 'CivilianPanicked' template.

pledbrook:
I think the current implementation doesn't handle the approach of red alert enemy units though. Probably need to add that as a trigger.

PR #549, opened 2019-06-05
WIP Make civilians run from XCOM (#548)

pledbrook:
If XCOM doesn't start concealed, then hostile civilians won't run from the
soldiers. They'll walk instead. This change fires a 'SquadConcealmentBroken'
event at the start of unconcealed missions, which ensures that civilians
are panicked by XCOM's presence and hence run from them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -152,6 +152,7 @@ All notable changes to Vanilla 'War Of The Chosen' Behaviour will be documented 
 - Gremlins (and other Cosmetic Units) are now correctly tinted and patterned (#376)
 - Register tactical event listeners in TQL (#406)
 - Allow mods to decide which team(s) are granted an ability via X2SitRepEffect_GrantAbilities and better document that class (#445)
+- Make civilians run from XCOM on missions where XCOM doesn't start concealed (#548)
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


MrNiceUK:
Are there any other side effects of  firing the event?

pledbrook:
@Musashi1584 This should not be merged for 1.19.0.

rcd47:
I just built this change locally and verified that it would fix a different bug I've been investigating. SquadConcealmentBroken is also not triggered if the High Alert dark event is active. That seems to be a problem for supply extraction missions: ADVENT never starts marking crates. With this change in place, they start marking crates at the end of XCOM's first turn.

pledbrook:
@rcd47 That mission just uses "squad loses concealment" as a lazy way of "ADVENT/aliens have seen a member of your squad". There may be other instances where this is the case, although I don't know of any.

Yet it seems risky to trigger this event that may have many listeners just to fix the problem of a badly-coded mission. I'm honestly torn on this one. It feels like a 50-50, where some things will be fixed with the change while a similar number of things will break. In a situation like that, perhaps it is best to stick to the devil you know.

pledbrook:
I have fixed this with a template modification in LWOTC. Civilians now only go into panic when an unconcealed XCOM soldier approaches within 7 tiles.

Issue #551, opened 2019-06-06
Rescue rings disappear after a neutral unit swaps to team XCom and moves


pledbrook:
To see the problem, watch [this video](https://youtu.be/7P4nhyUEZwc?list=PLj_Goi54wf0dRs0dsBfzwzTyjKDMdyMIc&t=791) of Christopher Odd playing the Gather Survivors mission. Everything seems OK when he moves to rescue the resistance soldier, but when he moves that resistance soldier, see how the rings disappear from the other VIPs.

pledbrook:
Note that this is probably less of an issue for Gather Survivors than it is for the terror missions in LW2.

PR #552, opened 2019-06-06
Fix missions with rescue rings (#551)

pledbrook:
In Gather Survivors and several LW2 missions, XCOM has to rescue units and
take them to evac. To enable player control of the rescued unit, its team
is swapped to XCom. But XGUnit then removes the range indicators from all
members of that unit's squad, which is not XCOM at that stage but team
Neutral.

This fix ensures that the swap team action changes the unit's squad (and
the pawn's team) in addition to the existing XGUnit team change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -246,7 +246,8 @@ RunPriorityGroup=RUN_STANDARD
   a pod tries to patrol to any of them (#508)
 - Make disorient reapply to disoriented units so that things like flashbangs can
   still remove overwatch from disoriented units (#475)
-
+- Make sure that rescue rings do not disappear on other rescuable units after a
+  neutral unit swaps to team XCom (#551)
 
 ## Miscellaneous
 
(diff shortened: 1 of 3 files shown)


Xymanek:
What is the argument for/against using a config gate?

pledbrook:
Argument for: this is a low-level change that could have unintended side effects. The config variable ensures that the player can pull out of the change if it breaks something in their game. Also, tracktwo and I are a bit nervous because we thought that Firaxis would have done this already when fixing the rescue rings for Gather Survivors. So perhaps the changes introduce other issues that couldn't be resolved? We just don't know.

Argument against: it's a bug fix, plain and simple. If it works without breaking anything else, it absolutely should not be gated behind a config variable.

So the config var is simply a defensive measure and not about giving players the option to play with one behaviour vs another.

pledbrook:
I'm going to keep this as a draft PR until it's had some real field testing with LWOTC.

pledbrook:
I haven't noticed any issues with this and it has been tested extensively in LWOTC now. Any argument against merging it into beta?

Xymanek:
> we thought that Firaxis would have done this already when fixing the rescue rings for Gather Survivors

Did you test this in vanilla game?

pledbrook:
In terms of whether the rescue rings remained in place? Pretty sure I did when I first created the fix.

Xymanek:
Considering that 

* Rings in Gather Survivors are broken with vanilla XComGame
* If you save and reload right after, the pawn would be initialized with the values that this fix sets

I would argue for removing the config gate

pledbrook:
@Xymanek Removed the config var guard.

PR #556, opened 2019-06-09
Add hook to allow mods to override hit chance

pledbrook:
This change adds an 'OverrideFinalHitChance' event that allows mods to override the shot breakdown for abilities. This is necessary to implement the graze band for LW2, but could be used by others as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -142,6 +142,9 @@ RunPriorityGroup=RUN_STANDARD
 - Allow Mods/DLC to modify encounters after creation (#136)
 - Allow Mods/DLC to modify encounters generated as reinforcements (#278)
 - Allow Mods/DLC to alter mission data after SitRep creation (#157)
+- Add an array of `OverrideFinalHitChance` function delegates to `X2AbilityToHitCalc`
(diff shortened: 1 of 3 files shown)


pledbrook:
This shouldn't be merged until we can work out why the existence of this event trigger is causing an occasional, but reproducible hang/freeze in the game. Could be something else in LWOTC, but don't know yet.

pledbrook:
I'll be replacing this implementation, probably with one that uses a delegate or array of delegates in place of the event hook. Firing an event in the `FinalizeHitChance()` function seems very occasionally to cause a hang, even if there are no listeners for the corresponding event. It seems to happen exclusively when a unit is undertaking a melee attack, such as with Rend or LW2's Fleche.

pledbrook:
New implementation ready for review.

pledbrook:
Is the new implementation OK with folks? It seems to work reliably now. May just need testing in beta.

MrNiceUK:
Ok, but late now, and somewhat just an aesthetic thing, but would have preferred the array of delegates to be called `OverrideFinalizeHitChanceFns`, since `FinalizeHitChance()` is the name of the method whose behaviour being overridden...

PR #558, opened 2019-06-09
Fix carrying out soldier multiple times (#557)

### LINKED ISSUE #557: Allow soldiers to be carried out multiple times in a campaign
See [this LWOTC issue](https://github.com/long-war-2/lwotc/issues/212) for more details, but in essence, the base game prevents the player from carrying out a soldier more than once *during a whole campaign*. So if they're carried out once, and then collapse bleeding out in a subsequent mission, you have to leave them there.

I'd class this as a bug fix.

pledbrook:
The problem was that the game was setting the `bBodyRecovered` flag to true and then never clearing it again. So you could carry a soldier out once, but never again on any future missions.

This fix simply clears the flag at the beginning of tactical play. Fixed #557.


Xymanek:
Once this is merged, I think we should leave a comment on the [fix mod](https://steamcommunity.com/sharedfiles/filedetails/?id=1378606631) that it's no longer required

Issue #562, opened 2019-06-13
Allow mods to influence XP gain


pledbrook:
I propose to allow mods three hooks:

 1. An event that allows mods to cap kill XP, i.e. preventing a soldier from gaining more kill XP after a certain limit
 2. An event that allows mods to override the amount of XP a soldier has with respect to ranking up (so they could discount Wet Work or add their own XP tracking)
 3. An event that allows mods to hook into the post-mission XP distribution to perform their own stuff

In fact, there is no post-mission XP distribution in the base game unless the full XP system is enabled (which it's not by default). So the last event simply gives mods an opportunity to _add_ behaviour rather than override it.

Issue #221, opened 2018-04-18
Let enemy units track kills they make


RealityMachina:
Basically just an enhancement I want so the MOCX posters that track the kills an operative has made work properly.

robojumper:
I think we merged something that fixed this.

pledbrook:
Would this issue be adversely affected if I put the kill count increment behind a "CanAwardKillXp" event? The idea would be to allow mods to prevent the accumulation of kill XP based on any critiera they choose. I would like it to implement a cap on the total kill XP that can be gained in a mission.

The code would look something like:
```
if (TriggerCanAwardKillXp(Killer))
{
	Killer.KillCount += GetMyTemplate().KillContribution;
}
```

pledbrook:
@RealityMachina Do you know whether the above change I mentioned breaks the behaviour you want or not?

RealityMachina:
It's good, I only need KilledUnits itself, not the KillCount variable used for levelling up.

PR #563, opened 2019-06-13
Override xp distribution

pledbrook:
Implementation of #562 that's based on the LW2 highlander changes. The new events provide a way to influence the XP gain using the existing base game mechanic.

Note that one change may conflict with #221, because it's not clear whether the intention of that issue was to track kills or kill XP.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -6620,12 +6620,16 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
 	local XComGameState_Destructible DestructibleKiller;
 	local X2Effect EffectCause;
 	local XComGameState_Effect EffectState;
+	// Variables for issue #562
+	local float KillXp, BonusKillXp, KillAssistXp;
+	local int WetWorkXp;
+	// End Issue #562
 
 	local StateObjectReference objRef;
-	local X2CharacterTemplate myTemplate;
+	// local X2CharacterTemplate myTemplate;   // Unneeded with fix for issue #562
 
 	objRef = GetReference();
-	myTemplate = GetMyTemplate();
+	// myTemplate = GetMyTemplate();   // Unneeded with fix for issue #562
 
 
 	LogMsg = class'XLocalizedData'.default.UnitDiedLogMsg;
@@ -6708,7 +6712,6 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
 	{
 		Killer = XComGameState_Unit(NewGameState.ModifyStateObject(Killer.Class, Killer.ObjectID));
 		Killer.KilledUnits.AddItem(GetReference());
-		Killer.KillCount += GetMyTemplate().KillContribution; // Allows specific units to contribute different amounts to the kill total
 	}
 	//end issue #221
 	if( GetTeam() == eTeam_Alien || GetTeam() == eTeam_TheLost || GetTeam() == eTeam_One || GetTeam() == eTeam_Two) //issue #188 - let eTeam_One and eTeam_Two units count as enemies when they die
@@ -6720,20 +6723,8 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
 				/* issue #221 - move this section out of this check so all units track kill counts
 				Killer = XComGameState_Unit(NewGameState.ModifyStateObject(Killer.Class, Killer.ObjectID));
 				Killer.KilledUnits.AddItem(GetReference());
-				Killer.KillCount += GetMyTemplate().KillContribution; // Allows specific units to contribute different amounts to the kill total
 				*/ 
 				//end issue #221
-				// If the Wet Work GTS bonus is active, increment the Wet Work kill counter
-				XComHQ = XComGameState_HeadquartersXCom(History.GetSingleGameStateObjectForClass(class'XComGameState_HeadquartersXCom', true));
-				if(XComHQ != none)
-				{
-					if(XComHQ.SoldierUnlockTemplates.Find('WetWorkUnlock') != INDEX_NONE)
-					{
-						Killer.WetWorkKills++;
-					}
-					
-					Killer.BonusKills += (XComHQ.BonusKillXP);
-				}
 
 				if (Killer.bIsShaken)
 				{
@@ -6741,6 +6732,32 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
 				}
 
 				CheckForFlankingEnemyKill(NewGameState, Killer);
+
+				// If the Wet Work GTS bonus is active, increment the Wet Work kill counter
+				XComHQ = XComGameState_HeadquartersXCom(History.GetSingleGameStateObjectForClass(class'XComGameState_HeadquartersXCom', true));
+				if(XComHQ != none)
+				{
+					if(XComHQ.SoldierUnlockTemplates.Find('WetWorkUnlock') != INDEX_NONE)
+					{
+						WetWorkXp = 1;  // Issue #562: store in a variable so it can be overridden
+					}
+
+					BonusKillXp = XComHQ.BonusKillXP;  // Issue #562: store in a variable so it can be overridden
+				}
+
+				// Start Issue #562
+				/// HL-Docs: ref:OverrideKillXp
+				// Get the default values for kill XP and kill assist XP and then allow mods to
+				// override these values before applying them to the unit.
+				KillXp = GetMyTemplate().KillContribution; // Allows specific units to contribute different amounts to the kill total
+				KillAssistXp = KillXp;
+
+				TriggerOverrideKillXp(Killer, KillXp, BonusKillXp, KillAssistXp, WetWorkXp, NewGameState);
+
+				Killer.KillCount += KillXp;
+				Killer.BonusKills += BonusKillXp;
+				Killer.WetWorkKills += WetWorkXp;
+				// End Issue #562
 
 				//  Check for and trigger event to display rank up message if applicable
 				if (Killer.IsSoldier() && Killer.CanRankUpSoldier() && !class'X2TacticalGameRulesetDataStructures'.static.TacticalOnlyGameMode())
@@ -6760,7 +6777,7 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
 					{
 						KillAssistant = XComGameState_Unit(NewGameState.ModifyStateObject(Iter.Class, Iter.ObjectID));
 						KillAssistant.KillAssists.AddItem(objRef);
-						KillAssistant.KillAssistsCount += myTemplate.KillContribution;
+						KillAssistant.KillAssistsCount += KillAssistXp;  // Issue #562: use mod-provided values if overridden
 
 						//  jbouscher: current desire is to only display the rank up message based on a full kill, commenting this out for now.
 						//  Check for and trigger event to display rank up message if applicable
@@ -6773,7 +6790,7 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
 						//		EventManager.TriggerEvent('RankUpMessage', KillAssistant, KillAssistant, NewGameState);
 						//		KillAssistant.SetUnitFloatValue('RankUpMessage', 1, eCleanup_BeginTactical);
 						//	}
-						//}		
+						//}
 					}
 				}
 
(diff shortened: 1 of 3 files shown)


pledbrook:
@Xymanek Are you using this for Covert Infiltration? Does it have everything you need? Is it good to go?

Xymanek:
Ok, there is a small problem for me. For my needs, I need to override the `GetMyTemplate().KillContribution` value. In theory I could block existing code via `CanAwardKillXp` and do it manually, but then I would need to reimplement this whole block:

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/61f9887a1c32efda0cfbcfae59a1da8239596027/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc#L6349-L6411

Which I don't want to do for obvious reasons. Therefore, I would like 2 following changes:

* A new event `OverrideUnitKillContribution` (with the `NewGameState`, I'll need it)
* Pass `NewGameState` with `CanAwardKillXp`, so that it mirrors the above, and maybe someone will need it

pledbrook:
In what way do you need to override `KillContribution`? I mean, it's a template value, so you could do it in `OnPostTemplatesCreated`. Assuming that you want to "override" it dynamically, why not simply use the `XpKillShot` event to modify `KillCount` (if that's what you want to do) or use it to track extra XP via a custom unit value?

I just want to avoid making this more of a tangled web of XP hooks than it already is in order to satisfy the specific requirements of all mods that want to modify the XP system. Happy to consider changes to the existing hooks to make them as or more generally applicable, while making them easier to use.

Xymanek:
> In what way do you need to override `KillContribution`?

Scale by some multiplier, which is determined dynamically

> why not simply use the `XpKillShot` event to modify `KillCount`

Good idea, but unfortunately it won't work as the event is after the promotion popup logic, so it can cause scenarios where the popup triggered but the unit didn't actually level up. Additionally, it would require reversing the kill assist logic, which would be a mess to do

pledbrook:
So, would you want the kill contribution to apply to the kill assists or not? And whichever way you go on that, you may change your mind later, making the suggested hook ineffective.

However, you make a good point about the promotion popup, so it seems to me that we should add a new event that mimics `'XpKillShot'` but goes immediately before the promotion popup check. LWOTC would benefit from that too.

Xymanek:
> So, would you want the kill contribution to apply to the kill assists or not? 

It already does (`KillAssistant.KillAssistsCount += myTemplate.KillContribution;`), I just want to be able to scale it.

In any case, I just realized that I will need to rethink my approach due to `Killer.WetWorkKills++` and `Killer.BonusKills += XComHQ.BonusKillXP`.

But yes, an event like `XpKillShotCalculated` or `XpKillShotPrePromotionCheck` would be useful in any case

pledbrook:
I'd like to reorder the XP allocation steps from:

1. Allocate XP to killer
2. Shaken and bond stuff
3. Show promotion notification
4. Distribute assist XP
5. Fire `XpKillShot` event

to:

1. Allocated CP to killer
2. Shaken and bond stuff
3. Distribute assist XP
4. Fire `OnUnitDiedXp` event
5. Show promotion notification
6. Fire `XpKillShot` event

Thoughts?

Xymanek:
That would be problematic for any mods that rely on `RankUpMessage` event being triggered before kill assist ditribution

Xymanek:
Just found this amazing piece of code (see last line)

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/f3e185762061ab1c399360478d17bb450adcb8af/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StatusEffects.uc#L1995-L2018

Xymanek:
A small issue I found with current implementation of this PR: if `CanAwardKillXp` results in `false`, the following code is also blocked from running:

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/61f9887a1c32efda0cfbcfae59a1da8239596027/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc#L6368-L6373

Which to me seems like it should run even if "kill XP throttling" or other similar mechanics are active

MrNiceUK:
That's a remnant of the old pre-WotC shaken mechanic. Recap if you don't remember the details: There was a chance that a unit which was gravely wounded in a mission would become "shaken", which set their will to zero. To recover, had to participate in a mission and a) Not get Injured, and b) Get a kill. That both recovered their will, *and* gave a will bonus. A unit could only become shaken this way *once*, so sadly couldn't be abused to get super human will :P .

pledbrook:
It's not just the old Shaken behaviour, but the new cohesion mechanic in `CheckForFlankingEnemyKill()`.

Is StasisLance the Skullmine ability? If so, that probably means LWOTC awards double XP points for Skullmine kills 😆

MrNiceUK:
Right, didn't notice that method call, just the conditional `UnitsKilledWhileShaken` increment...
While yes, the skull jack is referred to as the stasis lance in code in several places, LWotC is safe, `StasisLanceEffectAdded` isn't used (`CreateStasisLanceEffect()` which uses that delegate is never called, and the localization strings it passes to `SetDisplayInfo()` aren't in `XComGame.INT`)

Musashi1584:
Is this good to go? @pledbrook @Xymanek

Xymanek:
This isn't ready, concerns in https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/563#issuecomment-546947814 and https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/563#issuecomment-547849403 are not addressed

pledbrook:
I have moved the code that handles the legacy Will restoration and the code that handles squad mate cohesion to before the check for whether XP can be earned for the kill. I have also added a 'XpKillShotPrePopup' event that's triggered before the promotion popup can appear.

Let me know if there are any other changes requested. This should cover the two outstanding ones.

Xymanek:
The problem with the new implementation for CI is that the kill assists are added after the popup. Which means that I have to subscribe to 2 events (`XpKillShotPrePopup` for killer and `XpKillShot` for assists) and do the exactly same thing... Ugly.

The easiest solution would be to move kill assists award before the `XpKillShotPrePopup`, but BC is a thing. Which leaves 3 options:

1) Do same thing in 2 events
2) Do everything in `XpKillShotPrePopup`, preemptively removing assists and hope that nothing breaks
3) Make a new event that overrides the value of kill contribution

Additionally, there is the matter of bonus kills which are currently handled same way (unapply, scale, apply) and I'm honestly not sure how to best organize those...

CI code for reference: https://github.com/WOTCStrategyOverhaul/CovertInfiltration/blob/95e1c181b4a1bf216f90e3ac5955ca4f11681d60/CovertInfiltration/Src/CovertInfiltration/Classes/X2EventListener_Infiltration.uc#L1109-L1140 Note that `XpMult = class'X2Helper_Infiltration'.static.GetKillContributionMultiplerForKill` (slightly above the linked block) does relatively heavy math and I would prefer to not run it multiple times.

pledbrook:
Can we not add a config option (disabled by default) to reorder what happens? I just don't see a way to sort out the mess otherwise without hacks on the listener side. Yes, there may be a mod that won't work with CI/LWOTC, but I think that's a price worth paying.

Xymanek:
Yeah, that's probably the best solution. It's a very minor thing anyway, the incompatibility chance is pretty low

pledbrook:
Another approach for CI might be to modify `Can

pledbrook:
Another option for CI might be to rename 'CanAwardKillXp' and use it to apply your own custom XP gain, since it looks like you roll back most of what vanilla does anyway.

Xymanek:
I think I wanted to do that in the beginning but then decided against it in order to preserve compat with CHL changes to this logic... But yeah, kinda a mess now.

Another issue with that approach is that `CanAwardKillXp` is triggered twice per kill

P.S. We should probably move this discussion to discord, this thread is getting too long

Xymanek:
The docs build is failing, see https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/runs/435642281

pledbrook:
I think I have a reasonable solution to this. Stay tuned!

Xymanek:
The suspense!

pledbrook:
@Xymanek I think this is a simpler implementation and should do exactly what you need.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -6615,42 +6604,73 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
 
 				CheckForFlankingEnemyKill(NewGameState, Killer);
 
-				//  Check for and trigger event to display rank up message if applicable
-				if (Killer.IsSoldier() && Killer.CanRankUpSoldier() && !class'X2TacticalGameRulesetDataStructures'.static.TacticalOnlyGameMode())
+				// Issue #562 - Allow mods to cap kill XP
+				/// HL-Docs: ref:CanAwardKillXp
+				if (TriggerCanAwardKillXp(Killer))
```
pledbrook, line 6609: In fact, we could probably do away with this, since the `OverrideKillXp` event could simply set all +XP to 0 if the kill count is above whatever cap is implemented.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -6493,12 +6493,14 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
 	local XComGameState_Destructible DestructibleKiller;
 	local X2Effect EffectCause;
 	local XComGameState_Effect EffectState;
+	// Variables for issue #562
+	local float KillXp, BonusKillXp, KillAssistXp;
+	local int WetWorkXp;
+	// End Issue #562
 
 	local StateObjectReference objRef;
-	local X2CharacterTemplate myTemplate;
```
Xymanek, line 6498: Could you leave this commented out instead of completely removed?

pledbrook:
I asked RealityMachina about `KillCount` and it's unnecessary for #221. That issue is about tracking killed units rather than XP and moving the `KillCount` changes into the XP handling simplifies things if I remember correctly.

Xymanek:
Sure, it might not be required for that particular use case. But you can't know that it's the only case that uses that functionality. It's already in CHL - hence covered by the BC policy. 

While I recognize that working around this detail will make this PR much messier, imagine what would happen if you were to break some mod with this and then need to fix it after this PR landed in a stable release...

Xymanek (approved):
Ok, I changed my mind. The reason is the following comment:

> the kill XP is awarded even if `CanEarnXp()` returns `false`

https://github.com/long-war-2/X2WOTCCommunityHighlander/pull/4#discussion_r472136139

I can accept the original API as too obscure and badly designed.

pledbrook:
I've commented out (rather than removed) the `myTemplate` variable and removed the `CanAwardKillXp` event because it's no longer needed.

pledbrook:
@Xymanek The `OverrideKillXp` event now passes the `NewGameState`, bringing it into line with the other events in that bit of code (and because I need it :smile:). Shouldn't affect unless you also want to use the `NewGameState` in your listener(s).

pledbrook:
OMG, this PR is so broken. Please bear with me. Some really stupid mistakes that mean it doesn't work.

pledbrook:
Sorry for the noise. This is now working. The tuple's data array was being set to the wrong size before and on top of that, I wasn't actually copying the tuple data into the trigger function's `out` parameters 🤦

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -6602,47 +6605,61 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
...
+				// End Issue #562
 
 				//  Check for and trigger event to display rank up message if applicable
 				if (Killer.IsSoldier() && Killer.CanRankUpSoldier() && !class'X2TacticalGameRulesetDataStructures'.static.TacticalOnlyGameMode())
 				{
 					Killer.GetUnitValue('RankUpMessage', RankUpValue);
 					if (RankUpValue.fValue == 0)
 					{
-						EventManager.TriggerEvent('RankUpMessage', Killer, Killer, NewGameState);
+						`XEVENTMGR.TriggerEvent('RankUpMessage', Killer, Killer, NewGameState);
```
robojumper, line 6650: What's up with this local -> macro change?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -6602,47 +6605,61 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
...
 						Killer.SetUnitFloatValue('RankUpMessage', 1, eCleanup_BeginTactical);
 					}
 				}
 
 				//  All team mates that are alive and able to earn XP will be credited with a kill assist (regardless of their actions)
-				foreach History.IterateByClassType(class'XComGameState_Unit', Iter)
+				foreach `XCOMHISTORY.IterateByClassType(class'XComGameState_Unit', Iter)
```
robojumper, line 6656: ... and this?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
pledbrook: It's an old PR. I probably thought the macro was the canonical way to do it at the time.

pledbrook:
Just an FYI, this is waiting on a review by Covert Infiltration team to assess whether it's suitable for their needs.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -6788,6 +6805,62 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
 
 }
 
+// Start Issue #562
+//
+/// HL-Docs: feature:OverrideKillXp; issue:562; tags:tactical
+/// Allows listeners to override the XP values granted by a kill, including
+/// the normal kill contribution, any bonus XP from resistance orders and the
+/// like, the XP granted for assists, and any Wet Work bonus. Mods can even
+/// set these values to zero to prevent XP gain.
```
Xymanek, line 6814: I think it's important to mention here (or even adjust the event name) that this event is triggered at most once per kill, since in both our cases it's used to implement some form of kill-xp-limiting

Xymanek (commented):
BC wise - excluding (/taking in account) the previous discussion (and its outcome) I see no issues in the current implementation (or they are really obscure).

Feature wise - this is indeed an improvement in terms of API design compared to the previous approach I had to use.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -6788,6 +6805,62 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
...
+///       inout float KillXp,
+///       inout float BonusKillXp,
+///       inout float KillAssistXp,
+///       inout int WetWorkXp,
+///       in XCGS_Unit Killer
+///     ]
+/// }
+/// EventSource: KilledUnit (XCGS_Unit)
+/// NewGameState: yes
+/// ```
```
Xymanek, line 6829: Please update to the new event doc syntax

pledbrook:
@robojumper @Xymanek I believe I've addressed all your latests comments. Let me know if I missed anything.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -6918,6 +6935,61 @@ protected function OnUnitDied(XComGameState NewGameState, Object CauseOfDeath, c
...
+/// This event is called no more than once for each kill.
+///
+/// ```event
+/// EventID: OverrideKillXp,
+/// EventData: [
+///       inout float KillXp,
+///       inout float BonusKillXp,
+///       inout float KillAssistXp,
+///       inout int WetWorkXp,
+///       in class[class<XComGameState_Unit>] Killer ],
```
Xymanek, line 6955:
```suggestion
///       in XComGameState_Unit Killer 
/// ],
```

No need for the `class[class<` syntax here

Xymanek (changes requested):
Rest LGTM

Xymanek (approved):

Xymanek:
Merging as I assume that no one else is interested/planning to review this

Issue #290, opened 2018-10-07
Change UIScreenStack to always consider subclasses


robojumper:
`UIScreenStack` has a number of functions, most notably `GetScreen`, that deal with classes but do not account for subclasses. 
```unrealscript
    // Returns the first instance of a Screen of the target class type.
    simulated function UIScreen GetScreen( class<UIScreen> ScreenClass )
    {
	    local int Index;
    	for( Index = 0; Index < Screens.Length;  ++Index)
	    {
		    if( ScreenClass ==  Screens[Index].Class )
			    return Screens[Index];
    	}
	    return none; 
    }
```
This causes a number of bugs and general weirdness in a lot of base game code and makes interoperability with mods even worse.

Even though changing it to account for subclasses is changing the behavior of the function and *may* introduce some bugs, we would still remove a huge number of bugs. Additionally, I have never seen code that relies on the current behavior of `GetScreen` (non-constructed examples welcome!)

I would suggest adding an optional boolean parameter (`true` by default) that controls whether subclasses are considered. One can still avoid a dependency on the Highlander by doing something like `GetScreen(class'UIArmory_Promotion').class == class'UIArmory_Promotion'` if subclasses are not desired.

pledbrook:
Isn't the normal procedure not the change the signature of functions in vanilla classes? Is `UIScreenStack` a special case?

robojumper:
@pledbrook Optional parameters can be added without problems if we don't expect the function to be overridden in mod subclasses.

pledbrook:
So should that be the approach I take in the pull request, i.e. of just adding an option argument to the existing methods? I'm guessing that no one would or should subclass `UIScreenStack`, but I don't know that for sure.

PR #568, opened 2019-06-17
Make `UIScreenStack` functions respect subclasses (#290)

pledbrook:
I have modified `GetScreen()` and `IsCurrentScreen()` to return `true` by default if screen types are subclasses of the given screen class. Although a breaking change, this fixes a lot of code that mistakenly
assumes that certain screen classes won't ever have subclasses.

As a fallback option, mods can use `GetScreen_CH()` and `IsCurrentScreen_CH()` to get the old behaviour.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
@@ -614,17 +614,42 @@ simulated function PopIncludingClass( class<UIScreen> ClassToRemove, optional bo
 //----------------------------------------------------------------------------
 //----------------------------------------------------------------------------
 
-// Returns the first instance of a Screen of the target class type.
+// Start Issue #290
+//
+// Lots of code in the base game and mods seems to use `GetScreen()` when
+// it should be using `GetFirstInstanceOf()`. This change means that
+// `GetScreen()` will also return any screen that's a subclass of the
+// specified one.
+//
+// If any code wants to ignore subclasses, then it can use the new
+// `GetScreen_CH()` method instead.
+
+// Returns the first instance of a Screen of the target class type (or a subclass).
 simulated function UIScreen GetScreen( class<UIScreen> ScreenClass )
+{
+	return GetScreen_CH(ScreenClass, true);
+}
+
+simulated function UIScreen GetScreen_CH( class<UIScreen> ScreenClass, bool IncludeSubTypes )
 {
 	local int Index;
 	for( Index = 0; Index < Screens.Length;  ++Index)
 	{
-		if( ScreenClass ==  Screens[Index].Class )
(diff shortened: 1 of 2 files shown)


pledbrook:
I'm wondering whether to gate this behind a config var so that players do have a fallback option if things break. But then mods might have to take into account having the var flag set and not set.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -614,17 +614,39 @@ simulated function PopIncludingClass( class<UIScreen> ClassToRemove, optional bo
...
 	{
-		if( ScreenClass ==  Screens[Index].Class )
+		if( IncludeSubTypes && ClassIsChildOf(Screens[Index].Class, ScreenClass) )
+		{
+			return Screens[Index];
+		}
+		else if( ScreenClass ==  Screens[Index].Class )
+		{
 			return Screens[Index];
+		}
```
Xymanek: If `IncludeSubTypes` is `true` but `ClassIsChildOf(Screens[Index].Class, ScreenClass)` check fails, the current logic will still attempt to check `if( ScreenClass ==  Screens[Index].Class )` (which will obviously fail)

pledbrook: Worth changing or not?

Xymanek: I think yes 😄 at least for a minor pref boost

pledbrook: Would you mind taking a look at the latest version? I think it's correct.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -614,17 +614,46 @@ simulated function PopIncludingClass( class<UIScreen> ClassToRemove, optional bo
...
+		if( IncludeSubTypes )
+		{
+			if (ClassIsChildOf(Screens[Index].Class, ScreenClass))
+			{
+				return Screens[Index];
+			}
+			else
+			{
+				return none;
+			}
```
MrNiceUK: This effectively means that only `Index = 0` gets checked surely, since either code path leads to a return if `IncludeSubTypes == true` regardless?

pledbrook: Oops. I'll come back to this when my brain is at least partially functioning.

pledbrook: Thanks for the catch. Updated.

Issue #571, opened 2019-06-21
Allow mods to override whether soldiers and loot are recovered at the end of a mission


pledbrook:
Vanilla XCOM 2 will only retrieve soldiers (bleeding out, unconscious, dead) and loot if a mission has a tactical objective. However, a tactical objective also implies corpse retrieval. This makes it impossible to design sweep missions that don't also provide corpses.

The specific use case for LW2 is the Network Tower mission, which completes when all the enemies have been killed. Since you've killed all the enemies, you should be able to retrieve incapacitated and dead members of your team, but there is no scope to send in a retrieval team to get all the corpses.

pledbrook:
Implemented by #572

PR #572, opened 2019-06-21
Allow mods to override body/loot recovery (#571)

pledbrook:
There is a new event, OverrideBodyAndLootRecovery, that allows mods to override the default behaviour of whether incapacitated/dead soldiers and loot are automatically recovered at the end of a mission.

The default behaviour is for this to happen only when a mission has tactical objectives and those objectives have been completed.

The event takes the form:
```
 {
   ID: OverrideBodyAndLootRecovery,
   Data: [inout bool DoBodyAndLootRecovery],
   Source: XComGameState_BattleData
 }
```

Implements #571.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -187,6 +187,12 @@ RunPriorityGroup=RUN_STANDARD
 - `OverrideAbilityIconColor` provides a tuple with the same ID as the
   event and data of the form `[bool IsObjective, string Color]` that allows
   mods to override the color of soldier abilities in the tactical HUD (#400)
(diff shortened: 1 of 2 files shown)


Xymanek:
As I said on discord: would be really nice to make it 2 separate booleans.

My main concern is that in case anyone would need such functionality in future, implementing it on top of this event will be ugly (ie. see `CanAdd[...]` mess)

pledbrook:
@Xymanek I have split it into separate body recovery and loot recovery events.

pledbrook:
@Xymanek Are you OK with the change to two events?

pledbrook:
@Xymanek Feedback addressed, assuming I understood it correctly.

PR #573, opened 2019-06-24
#561 SitReps additional checks event

### LINKED ISSUE #561: Add event in X2SitRepTemplate::MeetsRequirements
This will allow mods to run additional checks

Xymanek:
Closes #561 

Note that this PR includes commit 8b9e4d8 from #567 - this is the best I could do to prevent them from conflicting as they both augment `X2SitRepTemplate` which is currently not included in CHL's master

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2SitRepTemplate.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2SitRepTemplate.uc
@@ -107,8 +107,34 @@ function bool MeetsRequirements(XComGameState_MissionSite MissionState)
 		return false;
 	}
 
+	// Start issue #561
+	if (!TriggerAdditionalChecks(MissionState))
+	{
+		return false;
+	}
+	// End issue #561
+
 	return true;
 }
+
+// Start issue #561
+function bool TriggerAdditionalChecks(XComGameState_MissionSite MissionState)
+{
+	local XComLWTuple Tuple;
+
+	Tuple = new class'XComLWTuple';
+	Tuple.Id = 'SitRepCheckAdditionalRequirements';
+	Tuple.Data.Add(2);
+	Tuple.Data[0].kind = XComLWTVBool;
+	Tuple.Data[0].b = true;
+	Tuple.Data[1].kind = XComLWTVObject;
+	Tuple.Data[1].o = MissionState;
+
+	`XEVENTMGR.TriggerEvent('SitRepCheckAdditionalRequirements', Tuple, self);
+
+	return Tuple.Data[0].b;
+}
+// End issue #561
 
 static function bool CanLaunchMission(XComGameState_MissionSite Mission, out string FailReason)
 {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,6 +128,7 @@ RunPriorityGroup=RUN_STANDARD
 - Triggers the event `CovertAction_AllowResActivityRecord` to allow mods to enable/disable "covert action completed"
   record for the monthly resistance report (#696)
 - Triggers the event `OverrideDarkEventCount` to allow mods to change the number of dark events in the monthly report (#711)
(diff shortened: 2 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2SitRepTemplate.uc
```diff
@@ -100,9 +100,35 @@ function bool MeetsRequirements(XComGameState_MissionSite MissionState)
...
+	{
+		return false;
+	}
+	// End issue #561
+
 	return true;
 }
 
+// Start issue #561
+function bool TriggerAdditionalChecks(XComGameState_MissionSite MissionState)
```
pledbrook: Could you please add to the comment at least the form of the event so mod authors don't need to parse the code in order to understand how to use it? I've been using the format:

```
{
  ID: SitRepCheckAdditionalRequirements,
  Data: [inout bool SitRepAllowed, in XCGS_MissionSite Mission],
  Source: self (X2SitRepTemplate)
}
```

Xymanek: The problem I have with such approach is that I don't think this is the correct place for explanations. If I need to drill down to where the event is triggered, I'll see and understand what's passed there anyway in 95% of cases.

I was considered other documentation methods and the GitHub wiki is probably the best place for it

pledbrook: Then we have a fundamental disagreement :smile: I think code is a perfect place for explanations because that's where coders spend most of their time, and it's right by the place that benefits from the explanation.

That's not to say it should be the _only_ place for explanation. A wiki would be nice. But that means the wiki needs updating every time the code is updated and such situations typically result in the wiki getting out of sync with the code. I still think a wiki would be a nice _additional_ source of information.

Other considerations:

* Documentation here helps reviewers (such as myself) understand what's going on
* It's often a lot of cognitive effort to understand code, and the _why_ is often impossible to understand without accompanying comments
* Some folks will be digging through the code to see how things work (I do it all the time). They may then see an event that seems to do something they want, so why not make it easier for them by explaining the event a little right where they are rather than forcing them to a wiki?
* Updating the wiki page (if it happens) will be easier for events created in the past if they're documented in words, not just code.

Xymanek: > I think code is a perfect place for explanations because that's where coders spend most of their time

Yes, but I don't want to be digging into CHL/XComGame internals to use the event. If I do so, then in 95% of cases the extra explanations will not benefit **me**

> It's often a lot of cognitive effort to understand code, and the why is often impossible to understand without accompanying comments

Aside from few big reworks (eg. unit status) I disagree. CHL events are generally very simple and exist with purpose of overriding one variable only

> why not make it easier for them by explaining the event a little right where they are 

I strongly believe in self-documenting code (clear naming of functions and variables, types, etc) and I generally dislike reading comments that simply re-state what I see already from glancing at code. That said, there are obviously places which require more thought than just glancing at code, there I do leave comments (but this case isn't one of them IMO)

> Documentation here helps reviewers understand what's going on

It's a valid point to keep comments in the git diff. Furthermore, in perfect world, we would have some sort of docs autogen tool which would parse the source. Unfortunately that's not going to be the case here

An alternative method I can think of, is a file like `CHANGELOG.md` (so something like `HOOKS_DOCS.md`) which would contain explanations for all the hooks, so that I can ctrl+f for "sitreps" and instantly see how I would do something related to them (without the need to search the entire codebase)

pledbrook:
Would you like me to merge this as is, or do you want to add API.md to it?

Musashi1584:
Please update status @pledbrook @Xymanek

Xymanek:
Documentation is now handled by #700 instead of individual PRs, please merge without it

Issue #533, opened 2019-05-31
Allow mods to control which stats appear in UISoldierHeader


pledbrook:
`UISoldierHeader` has too little to space to display all unit stats even in the base game (Where Art Thou, Defence?) Throw in Psi Offense and any others that might be added/supported by mods, it makes sense to allow mods to control which stats are displayed in the armoury header.

I suggest we create an event that sends an array of stat types (from enum `ECharStatType`), which can be reordered or have elements replaced by mods.

Another option would be to send an array of stat type & value pairs, so that mods can override the displayed value as well.

pledbrook:
I wonder if this is something that @robojumper's Squad Select would also want to use with Show Me The Stats.

Xymanek:
Instead of allowing mods to change the value, perhaps pass a bool that controls whether to show the max or not?

pledbrook:
Isn't that what happens now? It displays max +/- modifier? Or is that just a LW thing?

Xymanek:
The will is `current`/`max`

pledbrook:
I don't have a strong opinion either way. Happy to add a bool. Although it would help to know how best to pass arrays of pairs via the tuple. Any suggestions?

Xymanek:
> Although it would help to know how best to pass arrays of pairs via the tuple

You would need to pass tuple as `Tuple.Data[0].o`, `Tuple.Data[1].o`, etc. I would argue for a specific event data class, instead of a tuple

jmartinez989:
ok so what I am thinking is to make an object that has 3 different items in it, one of which is the label for the stat and the other the value of the stat and the last is a bool which would determine whether or not we should show that stat. Then an array of those can be passed to SetSoldierStats (obviously the parameter list of the function would have to change).

jmartinez989:
So I was writing some code to implement this functionality where I was going to fire an event in `UISoldierHeader.SetSoldierStats()` but then I found that another class called `UIInventory_Implants` has a function called `SelectedItemChanged()` which calls `SetSoldierStats()` as well which could cause a problem to how this would work.

this might not actually be too much of a problem though because we can always get a referrence to the unit that we will be getting stats in the `SetSoldierStats()` function and pass it to the event to order the stats. @Xymanek any input on this? Would you be ok with just getting the unit referrence in the `SetSoldierStats()` function of `UISoldierHeader`?

Also would we want an option to color the stat labels and values like how will and psi are?

[...]

PR #575, opened 2019-07-04
Address Issue #533

jmartinez989:
This addresses the issue of allowing mods to override how the UISoldierHeader displays its stats. It uses 2 new classes that assist with this. The first is XComSolidierHeaderElement which is an object that holds information about a particular stat to display in the header like should the stat's max value be shown and should the value be colored. The other obect is XComSoldierHeader which is just a class that holds an array of XComSoldierHeaderElement objects and also a few referrences to other objects such as the Unit the stats are being displayed for and other items used to assist with stat calculation.

The XComSoldierHeader object will be passed to the 'OverrideSoldierHeader' event where the event listener can rearrange stats in any way and also add new stats if so desired. This also takes all of the stat calculation found in the PopulateData() function in UISoldierHeader and moves to a function called CalcAndDisplaySoldierStats() calculates stats and removes the call to SetSoldierStats(). SetSoldierStats has been left in the UISoldierHeader class in case other mods still make a call to it.


jmartinez989:
Closing this. Going to redo the pull request with the branch based off of beta.

PR #589, opened 2019-07-13
Commander actions

### LINKED ISSUE #499: "Commander actions" API for mods
The tactical HUD has a bar top-right that provides icons for non-soldier-specific actions. Unfortunately, the current implementation relies heavily on hardcoded flash stuff but I think it can be overcome.

**The API** would be an event with an array of actions, each of which would contain:

* Icon
* Text
* Callback on usage
* Maybe more things, depends

**For controller**, if

* 1 action - show as "evac" text is shown in vanilla
* 2 or more - show "commander actions" and on RS press, bring up a pause-like menu, but filled with the options (including the default evac option)

**For kb+m**, the icons would be positioned starting after the "end turn" one and before the chosen icon. Note that this will require some manual positioning recalculations but I think that it's quite possible to do with a few unsolicited flash calls. For buttons themsleves, we can take inspiration from unused `UITacticalHUD_CommanderHUD`

Using this, we could also fix the lack of ability to open sitrep and chosen information mid-battle with controllers

Xymanek:
Closes #499

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_AbilityContainer.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_AbilityContainer.uc
@@ -1053,11 +1053,13 @@ simulated function UpdateAbilitiesArray()
 	
 	if (`ISCONTROLLERACTIVE)
 	{
+		// TODO
 		UITacticalHUD(screen).UpdateSkyrangerButton();
 	}
 	else
 	{
-		UITacticalHUD(screen).m_kMouseControls.SetCommandAbilities(arrCommandAbilities);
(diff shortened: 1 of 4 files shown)


Xymanek:
Notes to self:

* `XComTacticalInput::ActiveUnit_Moving::Stick_R3` and `XComTacticalInput::UsingTargetingMethod::Stick_R3`
* `UITacticalHUD_AbilityContainer::DirectConfirmAbility` - `m_iMouseTargetedAbilityIndex` (current)
* Usages of `m_kMouseControls.CommandAbilities`

pledbrook:
Should this be in the 1.20.0 milestone?

Xymanek:
I don't think so. I have other priorities right now, so I don't think I'll finish it in time by 1.20

PR #591, opened 2019-07-15
Add OnSortFinished to UIPersonnel

VampireMonkey:
Changed to use Event instead of delegate


Xymanek:
Could you make the event name more specific? They are global and `OnSortFinished` is very generic

VampireMonkey:
There you go

PR #597, opened 2019-08-03
Allow mods to temporary prevent DEs from completing

### LINKED ISSUE #596: Allow mods to temporary prevent DEs from completing

Xymanek:
Closes #596 

~~[Better compare](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/compare/1eca2b885e808dcae6a5e8445b6a3112680cda7a...460080496db565157cac47bbf291ebdd041b8a94)~~ seems no longer necessary after the rebase

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_DarkEvent.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_DarkEvent.uc
@@ -27,7 +27,7 @@ var StrategyCost					   RevealCost;
 /// HL-Docs: feature:TemporarilyBlockDarkEventActivation; issue:597; tags:strategy,compat
 /// While this flag is turned on, the dark event cannot activate/complete (transition from preparing to active)
 /// and will simply "wait" until this flag is turned off. Another possible way of looking at it 
-/// is that the DE is postponed indefinetly while this flag is active but only if the EndDateTime was reached.
+/// is that the DE is postponed indefinitely while this flag is active but only if the EndDateTime was reached.
 /// HL-Include:
 var bool							   bTemporarilyBlockActivation;
 ///
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_DarkEvent.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_DarkEvent.uc
@@ -24,7 +24,7 @@ var float							   TimeRemaining;
 var StrategyCost					   RevealCost;
 
 // Issue #596
-/// HL-Docs: feature:TemporarilyBlockDarkEventActivation; issue:597; tags:strategy,compat
+/// HL-Docs: feature:TemporarilyBlockDarkEventActivation; issue:597; tags:strategy
 /// While this flag is turned on, the dark event cannot activate/complete (transition from preparing to active)
 /// and will simply "wait" until this flag is turned off. Another possible way of looking at it 
 /// is that the DE is postponed indefinitely while this flag is active but only if the EndDateTime was reached.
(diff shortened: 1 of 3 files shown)


Xymanek:
> This seems to be an unusual approach for the highlander, which normally uses events to allow mods to influence vanilla behaviour

This is generally done out of necessity, because the behaviour is in either

* Native gamestate classes, where we can't change the data layout
* Non-gamestate classes, where an event/dlc info hook is the only way to make things non-static (non-config)

> why not use an event 

Because in this case we aren't bound by the 2 limitations above

> check whether any mods want to keep the dark event active?

That's not what this PR does though. It controls whether the DE can transition from preparing to active (NOT from active to expired)

MrNiceUK:
Agree with calling it "Activation" not "Completion". On the Hungarian notation point, for whateverreason it's far from uncommon in the XCom 2 UC source to use the `b` for bool flags even when Hungarian isn't used for other properties/variables. Speculation: I think it's because you can't past bools as out parameters in UC, so in some areas flags are stored as 0/1 ints, which often also have the `i` prefix, where it has to pass through out parameters (see that a lot in the various step/peek out related methods for example). So may have just become normalized by Firaxis devs to note if a flag is a bool or an int...
(aside, I can't seem to find what to click on to actually see the detail of  your requested change?)

Xymanek:
I agree that the naming is confusing, but I went with FXS's terms that you can see elsewhere in code. For example, 2 lines later: `if(DarkEventState.CanComplete())`

On the point of `Temporary`: I specifically included it to make it obvious that the flag isn't meant for permanent changes (like forever disabling the DE)

pledbrook:
Then it should be `bTemporarilyPreventCompletion`, but I still don't think that really helps with clarity. To be honest, I think it would be clearer to have "BlockCompletion()" and "UnblockCompletion()" functions instead.

Xymanek:
Do you mean making the flag protected/private and have only the functions update it?

pledbrook:
Yes. It makes it clearer that it's something one activates and then deactivates. It's not clear how to use the variable otherwise.

Xymanek:
I'm not certain that adding mutator funcs for a boolean member is a good solution.... Would extensive commentary around the variable be a good alternative?

pledbrook:
> I'm not certain that adding mutator funcs for a boolean member is a good solution

Why not? It will also allow the opportunity to accept a `NewGameState` parameter, which if not provided, will result in the mutator function setting up a new game state change container for the modification.

Musashi1584:
Open change requests still valid? @Xymanek

Xymanek:
Docs fail sees unrelated

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_DarkEvent.uc
```diff
@@ -23,6 +23,15 @@ var TDateTime						   EndDateTime;
 var float							   TimeRemaining;
 var StrategyCost					   RevealCost;
 
+// Issue #596
+/// HL-Docs: feature:TemporarilyBlockDarkEventActivation; issue:597; tags:strategy,compat
```
robojumper, line 27:
```suggestion
/// HL-Docs: feature:TemporarilyBlockDarkEventActivation; issue:597; tags:strategy,compatibility
```

If this has the compatibility tag, a section describing the potential compatibility issue and workarounds/fixes is required, see https://x2communitycore.github.io/X2WOTCCommunityHighlander/compatibility/

Alternatively, remove the tag

Xymanek: Oof, I didn't even notice that was there. Removed - it shouldn't be

robojumper:
> 
> 
> Docs fail sees unrelated

pushed 83463250227150a0ad6369b5fd5627c94095a9b5 to #899 to catch this sort of issue

Issue #590, opened 2019-07-14
Add more data types to `XComLWTuple`


pledbrook:
Add the following data types to `XComLWTuple`:

* `Vector`
* `Rotator`
* `TTile`
* Array of `Object`
* Array of `int`
* Array of `float`
* Array of `string`

Musashi1584:
Might also be useful:

Arrays of rotator, vector, ttile, name

pledbrook:
Is it sensible to simply dump a load of variables into `XComLWTValue`? I assume these are all taking space when not being used. If folks think it's fine, happy to add as many as makes sense.

Musashi1584:
Dont think that it would have any significant memory or perfomance impact to have a few variable declared that dont get used.

Iridar:
@robojumper are we sure this issue needs docs? It seems XComLWTuple is already well documented.

PR #598, opened 2019-08-08
Add more data types to XComLWTuple (#590)

pledbrook:
Implements #590 by adding `Vector`, `Rotator`, `TTile` and array version of those + existing basic types to `XComLWTuple`.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComLWTuple.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComLWTuple.uc
@@ -3,12 +3,18 @@
 //  AUTHOR:  tracktwo / Pavonis Interactive
 //  PURPOSE: A general-purpose data structure for sharing information between mods.
 //
+// == IMPORTANT ==
+//
+// XComLWTuple is based on the original LWTuple that many other mods package themselves,
+// but it cannot be used interchangeably with that class. That means you *have* to use
+// XComLWTuple for community highlander hooks that rely on it.
+//
 // == DESCRIPTION ==
 //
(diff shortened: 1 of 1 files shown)


Xymanek:
I would also clarify in the comment how this one differs from non-CHL one and why we can't add stuff there

pledbrook:
Which comment? The one in the source file or the commit message?

Xymanek:
I meant the one in the file

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComLWTuple.uc
```diff
@@ -3,6 +3,15 @@
...
+// == IMPORTANT NOTE ==
+//
+// This class has now changed, making it incompatible with mods that contain their own
+// copy of the old LWTuple class. This is because mods that require the tuple can simply
+// depend on the community highlander. As a result, we can make changes to this class
+// that make it easier to mod XCOM 2.
+//
+// The description below comes from the original LWTuple class created by tracktwo.
+//
 // == DESCRIPTION ==
```
pledbrook: @Xymanek Was this the kind of thing you were thinking of?

Xymanek: > making it incompatible with mods that contain their own copy of the old LWTuple class

That's actually not true. `XComLWTuple` is different class to `LWTuple` and both can be used simultaneously.

The issue that I wanted to point out in the comment is that the 2 classes were identical... until this point. So the comment should explain why one was changed and the other wasn't

Issue #600, opened 2019-08-10
Allow mods to override the UI classes used for promotions


pledbrook:
`XComHQPresentationLayer.ShowPromotionUI()` is currently hard coded to use specific UI classes for each of the promotion screens: `UIArmory_Promotion`, `UIArmory_PromotionHero` and `UIArmory_PromotionPsiOp`. This results in New Promotion Screen By Default (and any other mod that wants to override the promotion screens) requiring hacks to do their thing.

I suggest we add an event that allows mods to specify alternative UI classes for these three screens.

PR #601, opened 2019-08-10
Add event hook to override promotion UI screens (#600)

pledbrook:
I have added an 'OverridePromotionUIClasses' event that's fired from `XComHQPresentationLayer.ShowPromotionUI()`. This event allows mods to specify different UI classes to use for the three different promotion screens: standard, hero, and psi operative.

The event takes the form:
```
 {
    ID: OverridePromotionUIClasses,
    Data: [inout string StandardPromotionUIClassName,
           inout string HeroPromotionUIClassName,
           inout string PsiOpPromotionUIClassName],
    Source: XComHQPresentationLayer
 }
```

There is also another new event, 'OverridePromotionBlueprintTagPrefix', that allows mods to override the blueprint tag used within the after action cinematic (`UIAfterAction`). This means that can position the camera correctly for the promotion screen they're displaying.

Implements #600.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,8 @@ RunPriorityGroup=RUN_STANDARD
   other ways than just the tooltip and image (#537)
 - Triggers the event `MissionIconSetScanSite` to allow mods to customize a scan site's icon in other
   ways than just the tooltip and image (#537)
+- Triggers the event `OverridePromotionUIClasses` to allow mods to override the UI classes used for the
+  three different promotion screens (#600)
 
 
 ### Modding Exposures
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAfterAction.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAfterAction.uc
@@ -1,0 +1,1457 @@
+//---------------------------------------------------------------------------------------
+//  *********   FIRAXIS SOURCE CODE   ******************
+//  FILE:    UIAfterActionReport
+//  AUTHOR:  Sam Batista -- 5/20/14
+//  PURPOSE: This file controls the post-mission squad view.
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//--------------------------------------------------------------------------------------- 
+
+class UIAfterAction extends UIScreen;
+
+struct TUIAfterActionFlyoverInfo
+{
+	var string ColorStr; 
+	var string Label;
+	var string Icon; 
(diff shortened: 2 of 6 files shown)


Xymanek:
This seems a very weird way to do it. Why not just have

* the vanilla logic picks a class
* feed that class into an event 
* listeners modify the class
* `ShowPromotionUI` uses the final class to spawn the screen

pledbrook:
Weird in what way? Because it passes three UI classes in? Or because it converts classes to strings and back again? Both?

The code basically does what you suggest, except it allows mods to change all three classes without worrying about which one the game wants to display. We could pass a single class, but then I think we should also pass an enum or similar to identify which class is requested. This protects against changes to the class/class name by other listeners or by the highlander in the future.

Xymanek:
> Weird in what way? Because it passes three UI classes in? 

Yes. More specifically, because if I want to change the promo UI for a specific unit, I need to know which of 3 is going to be used

> Or because it converts classes to strings and back again?

I didn't notice it at first, but yeah, this is weird too. You can just pass the class inside the object field of the tuple

pledbrook:
> Yes. More specifically, because if I want to change the promo UI for a specific unit, I need to know which of 3 is going to be used

You don't need to know with the current implementation. You change the class for each promotion screen you want to override.

I'm not fussed either way, so if you would prefer the event to request a specific UI screen class, then I'll change it.

Xymanek:
> so if you would prefer the event to request a specific UI screen class, then I'll change it.

Yeah, I think that would be much less confusing for the event consumers

pledbrook:
Updated with the fix for handling of "NOT_" prefixes on ability prerequisites.

pledbrook:
@robojumper @Xymanek Do the changes address your concerns?

Xymanek:
I'm a bit confused by the enum... Why not just make `TriggerOverridePromotionUIClass` accept `class<UIArmory_Promotion>` as argument?

pledbrook:
I have a sense of deja vu. Perhaps this was raised in Discord. It was basically because I wasn't sure whether classes were objects that could be passed around. I think.

pledbrook:
Actually, I think I remember what the enum is for: it's for situations where you have multiple mods that are overriding the screen class, so you need at least one argument that doesn't change and identifies the screen. That could still be the screen class, but I considered an enum simpler for the purpose of identity (which screen are you requesting?).

pledbrook:
@Xymanek @MrNiceUK Is the concern specifically the enum as an argument to that function or also the enum passed in the event tuple? Either way, I prefer the enum as a way to identify which screen is required without the confusion of whether the screen class is the original one requested or an overridden one. I think the enum is necessary/important to the event data, but I also think it simplifies comprehension in having it as the argument to the `TriggerOverridePromotionUIClass()` function too.

Issue #602, opened 2019-08-15
Allow PCS granting PsiOffense to be equiped by other classes than PsiOperative


boundir:
`CanEquipImplant()` specifies if units is allowed to equip PCS. So far it forbid units to equip PCS granting PsiOffense on any other class than PsiOperative.
I would like to change this to allow units with PsiOffense to be able to equip it.

Class: UIInventory_Implants

Xymanek:
Please note that despite the rather extensive mentions spam above this comment, this issue is **not** implemented - there were several candidate PRs, some were even merged, but then they were reverted in #721 due to outstanding issues with the implementation

PR #604, opened 2019-08-16
Psi op change

pledbrook:
Changes the requirements for Psi PCSes so that they just require a soldier with a Psi Offence > 0 rather than specifically being a Psi Operative.

Implements #602.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIInventory_Implants.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIInventory_Implants.uc
@@ -1,0 +1,437 @@
+
+class UIInventory_Implants extends UIInventory;
+
+var localized string m_strNoImplants;
+
+var localized string m_strInstallImplantTitle;
+var localized string m_strInstallImplantText;
+var localized string m_strReplaceImplantTitle;
+var localized string m_strReplaceImplantText;
+
+var array<XComGameState_Item> Implants;
+
+simulated function InitScreen(XComPlayerController InitController, UIMovie InitMovie, optional name InitName)
+{
+	super.InitScreen(InitController, InitMovie, InitName);
+
+	List.OnItemClicked = OnItemSelected;
+
+	if(`ISCONTROLLERACTIVE)
+		List.OnItemDoubleClicked = OnItemSelected;
+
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UIInventory_Implants.uc
```diff
@@ -131,7 +131,7 @@ simulated function bool CanEquipImplant(StateObjectReference ImplantRef)
 			return false;
 	}
 
-	return class'UIUtilities_Strategy'.static.GetStatBoost(Implant).StatType != eStat_PsiOffense || Unit.IsPsiOperative();
+	return class'UIUtilities_Strategy'.static.GetStatBoost(Implant).StatType != eStat_PsiOffense || Unit.GetCurrentStat(eStat_PsiOffense) > 0; // # Issue 602
```
pledbrook: @boundir Actually, this breaks with LWOTC because *all* soldiers have a PsiOffense score, but aren't necessarily Psi Operatives.

I think it would be better to add a config array containing valid psi soldier classes that mods could add to via INI. Another thing to consider is whether those soldier classes want to be more generally considered as PsiOp classes, in which cases an event that allows mods to override `IsPsiOperative()` might be useful.

boundir: Is it bad if LW Rangers _can_ equip Psi PCS?
Will have to check if we can modify `IsPsiOperative()` without undesired aftereffects.

pledbrook: When I get round to this, I'll add a config array of soldier classes. That will not change the behaviour of the base game, while still allowing mods to enable Psi PCSes for their own soldier classes.

Musashi1584:
I added a changelog entry in the merge commit

Xymanek:
Why was this merged? There is a pending report of it breaking things: https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/604#discussion_r339296674

Musashi1584:
Sorry missed that

robojumper:
This change was **reverted** on master.

PR #610, opened 2019-08-23
Fix avenger defense abilities

pledbrook:
Modifies `SeqAct_SpawnUnitFromAvenger` to add the unit's items to the new game state that used to initialise the unit's abilities. This ensures that `MergeAmmosAsNeeded()` has access to them for the merge.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqAct_SpawnUnitFromAvenger.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqAct_SpawnUnitFromAvenger.uc
@@ -1,0 +1,292 @@
+///---------------------------------------------------------------------------------------
+//  FILE:    SeqAct_SpawnUnitFromAvenger.uc
+//  AUTHOR:  David Burchanowski  --  11/05/2014
+//  PURPOSE: Spawns a unit from the avenger reserves
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+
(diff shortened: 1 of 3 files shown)


Xymanek:
I don't think that this is the proper solution. The proper solution would be to convert `MergeAmmoAsNeeded` to use history, as history already returns states from the pending gamestate

pledbrook:
@robojumper As you suggested the solution in this pull request, at least as an option, what are your thoughts on this?

robojumper:
@Xymanek's suggestion is probably a bit more robust, though I'm not 100% sure about the "as history already returns states from the pending gamestate" thing -- at the end of the day, we want write access to the state object and need to ensure the history frame has it.

pledbrook:
So any idea what the code needs to be? I assume it needs to take into account the `StartState` argument passed in to `MergeAmmoAsNeeded()`.

pledbrook:
Is the argument even a start state? It looks to me like it can be any sort of `NewGameState`.

pledbrook:
I think this is now implemented in the way @Xymanek suggested, but would appreciate some eyes on it. Does fix the original problem, which is a good sign.

MrNiceUK:
Looks good to me, that I can tell what the change is trying to achieve even without the comments is a sign of a good implementation! Just one comment, probably an artefact of an earlier implementation, but there no longer seems to be a reason to add `SeqAct_SpawnUnitFromAvenger.uc` to the CHL? The only change appears to be a blank line in the  middle (unless there are changes in the commit which adds it to the repo, which there shouldn't be!)

pledbrook:
The first commit adds _SeqAct\_SpawnUnitFromAvenger.uc_ to the highlander. I'll be removing that commit and the second one if the new implementation is accepted.

Musashi1584:
I added a changelog entry in the merge commit

PR #614, opened 2019-09-05
UIX2PanelHeader SetText automatically realize

### LINKED ISSUE #613: UIX2PanelHeader's SetText does nothing
Because there is an internal flash function, `realize` that handles updating the display and it's only called automatically as part of `mc.BeginFunctionOp("setWidth");`

Xymanek:
Closes #613


Xymanek:
The problem this is solving is that calling `SetText()` will have 0 visual effect. However, calling `SetHeaderWidth()` will update the visuals, possibly with new text.
The reason is that `setWidth` in flash calls `realize`, which is the function responsible for updating the visuals

pledbrook:
Quick question: why the extra property? Why not always invoke `realize` if the text changes? Are there circumstances in which the UI shouldn't be updated when the text is changed?

Xymanek:
Because some code might be relying on the vanilla behaviour? I feel like it's easy to break stuff by making this default, so I made it opt-in

PR #618, opened 2019-09-06
Fix rocket targeting

pledbrook:
`X2TargetingMethod_RocketLauncher` now checks the firing line between the step-out (peek) tile and the target, rather than between the unit tile and the peek tile. The latter meant that no matter what tile you targeted, it would be an unobstructed path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -246,6 +246,8 @@ RunPriorityGroup=RUN_STANDARD
   a pod tries to patrol to any of them (#508)
 - Make disorient reapply to disoriented units so that things like flashbangs can
   still remove overwatch from disoriented units (#475)
+- Fix rocket targeting so that it isn't always unobstructed when the shooter has
+  a valid step-out tile (#617)
 
 
 ## Miscellaneous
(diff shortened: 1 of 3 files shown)


MrNiceUK:
Although, as covered in more detail on the issue/discord, the requirement of LOS to get stepout does somewhat mitigate the fallout from the bug.

pledbrook:
So are folks OK with this change? @MrNiceUK?

MrNiceUK:
Forgot about this PR. Don't have a problem with it as a pure bugfix, just be prepared for the complaints about the CHL breaking/nerfing rocketlaunchers when it hits the workshop 😋

ansabalirai:
I would like to pre-emptively complain about the rocket launcher aim being definitely worse off than before, at least compared to vanilla 😆

pledbrook:
This is how egregious the cheating is:

image

Save file to reproduce (needs Skirmish+ mod):

[save_BUG Rocket Targeting.zip](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/files/3793427/save_BUG.Rocket.Targeting.zip)

I vote we just fix it rather than guarding it behind a config var.

PR #621, opened 2019-09-13
Better custom keybinds

### LINKED ISSUE #501: Improve custom keybinds usability
#198 introduced a hook for custom keybinds, but using it relies on an UISL switching the input function on/off in all 4 events. This is a lot of boilerplate, especially when the input function is not required for every instance of the screen (afaik lwotc project had something like that).

I propose adding `SCREENSTACK.RegisterOnInputForScreen(Screen, Delegate)`. The delegate would be evaluated before the screen itself is given an option to handle the input, right after

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/b9bd25d416760648b572c3f6e095531ac0ff48b6/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc#L86-L95

When the screen is removed, all associated callbacks would be removed

Xymanek:
Closes #501 

See https://github.com/WOTCStrategyOverhaul/CovertInfiltration/commit/573392c00d33e6608167eee9cfb0fc8dcde42e06 for sample usage

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
@@ -9,6 +9,14 @@
 
 class UIScreenStack extends Object;
 
+// Start issue #501
+struct InputDelegateForScreen
+{
+	var UIScreen Screen;
+	var delegate<CHOnInputDelegateImproved> Callback;
+};
+// End issue #501
+
 var bool IsInputBlocked; // Block UI system from handling input.
 var XComPresentationLayerBase Pres;
 
@@ -20,8 +28,10 @@ var bool bPauseMenuInput; //Named to the pause menu as this is not for general u
 var array<UIScreen> Screens;
 var array<UIScreen> ScreensHiddenForCinematic;
 var array< delegate<CHOnInputDelegate> > OnInputSubscribers;		// issue #198
+var protected array<InputDelegateForScreen> OnInputForScreenSubscribers; // Issue #501
 
 delegate bool CHOnInputDelegate(int iInput, int ActionMask);		// issue #198
+delegate bool CHOnInputDelegateImproved(UIScreen Screen, int iInput, int ActionMask); // Issue #501
 
 //----------------------------------------------------------------------------
 //----------------------------------------------------------------------------
@@ -96,6 +106,13 @@ simulated function bool OnInput( int iInput,  optional int ActionMask = class'UI
 		// If this Screen is not yet initialized, or is marked for removal stop the input chain right here - sbatista
 		if( !Screen.bIsInited )
 			return true;
+
+		// Start issue #501
+		if (ModOnInputForScreen(Screen, iInput, ActionMask))
+		{
+			return true;
+		}
+		// End issue #501
 
 		// Stop if a screen has handled the input or consumes it regardless.
 		if( (Screen.EvaluatesInput() && Screen.OnUnrealCommand( iInput, ActionMask )) || Screen.ConsumesInput() )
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -755,6 +773,77 @@ simulated function bool ModOnInput(int iInput, int ActionMask)
...
+function UnsubscribeFromOnInputForScreen (UIScreen Screen, delegate<CHOnInputDelegateImproved> Callback)
+{
+	local InputDelegateForScreen CallbackScreenPair;
+	local int i;
+
+	foreach OnInputForScreenSubscribers(CallbackScreenPair, i)
+	{
+		if (CallbackScreenPair.Screen == Screen && CallbackScreenPair.Callback == Callback)
+		{
+			OnInputForScreenSubscribers.Remove(i, 1);
```
pledbrook: Is it safe to remove an item from the collection while iterating over said collection?

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -755,6 +773,77 @@ simulated function bool ModOnInput(int iInput, int ActionMask)
...
+			return; // Since duplicates aren't allowed, we are done
+		}
+	}
+}
+
+function RemoveOnInputSubscribersForScreen (UIScreen Screen)
+{
+	local int i;
+
+	for (i = 0; i < OnInputForScreenSubscribers.Length; i++)
```
pledbrook: If you're removing items, isn't it better/safer to iterate from the end of the array to the beginning? You then wouldn't have that odd `i--` in the code.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -755,6 +773,77 @@ simulated function bool ModOnInput(int iInput, int ActionMask)
...
+	{
+		if (OnInputForScreenSubscribers[i].Screen == Screen)
+		{
+			OnInputForScreenSubscribers.Remove(i, 1);
+			i--;
+		}
+	}
+}
+
+simulated function bool ModOnInputForScreen (UIScreen Screen, int iInput, int ActionMask)
```
pledbrook: Is this part of the public API?

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
Xymanek: Yes in this case, because the iterator doesn't get to progress to next item if this code is hit

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
Xymanek: I personally find iterating from end more odd than this approach (and this is used actually in many places by FXS code).

But aside from personal preferences, both approaches are perfectly valid

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
Xymanek: Good point. In fact, `RemoveOnInputSubscribersForScreen` also shouldn't be public

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
MrNiceUK: Well, you also seen iterating from the end down, because it's editing that same array, also in Firaxis code. Anyway, I'm with pledbrook that iterating top down is less "odd" than modifying the iterating variable (which always look dodgy to me!).

Xymanek:
I still would like to make `ModOnInputForScreen` non-public before the beta hits stable as it's not supposed to be public API

Issue #633, opened 2019-09-23
Allow mods to control which research techs can be inspired


pledbrook:
The highlander should fire an event that allows mods to determine whether a tech can be researched or not. This provides them an important lever for balancing the strategy game. It's particularly important for repeatable research, which the inspiration mechanic doesn't seem to handle very well (the techs remain inspired even after the first inspired research of them is complete).

PR #634, opened 2019-09-23
Allow mods to control which techs can be inspired (#633)

pledbrook:
This adds a 'CanTechBeInspired' event fired from `XCGS_HeadquartersXCom.CheckForInspiredTechs()` that allows mods to block techs from being inspired. Mods simply need to implement a
listener that returns `false` in the tuple's bool field for those techs that it doesn't want to be inspired.

The event takes the form:

```
  {
    ID: CanTechBeInspired,
    Data: [in XCGS_Tech Tech, inout bool CanTechBeInspired],
    Source: XCGS_HeadquartersXCom,
 }
```

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersXCom.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersXCom.uc
@@ -6882,7 +6882,7 @@ function CheckForInspiredTechs(XComGameState NewGameState)
 				// Start Issue #633
 				//
 				// Check with mods to determine whether this tech can be inspired or not.
-				if (!TriggerCanTechBeInspired(AvailableTechState))
+				if (!TriggerCanTechBeInspired(AvailableTechState, NewGameState))
 				{
 					continue;
 				}
@@ -6921,25 +6921,23 @@ function CheckForInspiredTechs(XComGameState NewGameState)
 //
 //   {
 //      ID: CanTechBeInspired,
-//      Data: [in XCGS_Tech Tech, inout bool CanBeInspired],
-//      Source: self (XCGS_HeadquartersXCom)
+//      Data: [inout bool CanBeInspired],
+//      Source: TechState (XCGS_Tech)
 //   }
 //
-function bool TriggerCanTechBeInspired(XComGameState_Tech TechState)
+private function bool TriggerCanTechBeInspired(XComGameState_Tech TechState, XComGameState NewGameState)
 {
 	local XComLWTuple Tuple;
 
 	Tuple = new class'XComLWTuple';
 	Tuple.Id = 'CanTechBeInspired';
-	Tuple.Data.Add(2);
-	Tuple.Data[0].Kind = XComLWTVObject;
-	Tuple.Data[0].o = TechState;
-	Tuple.Data[1].Kind = XComLWTVBool;
-	Tuple.Data[1].b = true;
+	Tuple.Data.Add(1);
+	Tuple.Data[0].Kind = XComLWTVBool;
+	Tuple.Data[0].b = true;
 
-	`XEVENTMGR.TriggerEvent('CanTechBeInspired', Tuple, self);
+	`XEVENTMGR.TriggerEvent('CanTechBeInspired', Tuple, TechState, NewGameState);
 
-	return Tuple.Data[1].b;
+	return Tuple.Data[0].b;
 }
 // End Issue #633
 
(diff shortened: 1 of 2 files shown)


Xymanek:
I dislike passing `XCGS_HeadquartersXCom` as source - it's a singleton and we win nothing by passing it there. On the other hand, if we made the `XCGS_Tech Tech` the source then listeners can use native filtering (and the HQ can be easily fetched anyway)

pledbrook:
What's native filtering? I'm also wary of doing things like this for convenience. The HQ is technically the source of the event, so it's a bit confusing for something else to be passed as the source instead.

Xymanek:
> What's native filtering

6th argument to `EventManger.RegisterForEvent`.

> The HQ is technically the source of the event, so it's a bit confusing for something else to be passed as the source instead

MrNiceUK:
I agree with @Xymanek, and "technically the source" has no firm meaning.

pledbrook:
Thanks for the filtering explanation. So it does it compare the argument using value-based semantics or reference semantics? If value-based, how does Unrealscript know that one object is the same as another value-wise?

pledbrook:
My preferred option would be to move the "inspired" check into a `CanBeInspired()` method on `XCGS_Tech`. Not only would any code that has a Tech object be able to check for whether it can be inspired, but the "source" of the event would match the object that's producing the event (the `XCGS_Tech` instance in this case).

On the other hand, I expect some resistance to that based on it being an unnecessary refactoring. If that's the case, then I can hold my nose and make the `XCGS_Tech` the event source. But in that case, why make it part of the event data as well.

Also, "event source" in an OO system is conventionally the object that produces the event (if there is such an object - it gets a bit weird with non-object contexts), which applies to singletons as well as any others. Now, you could argue that Firaxis misnamed the argument and something like "EventCause" would be better. I could accept that.

MrNiceUK:
Yes, it would be redundant to put the XCGS_Tech in the event data, so just don't put it there. I doubt it would be too hard to find events where the source isn't `self` to show Firaxis didn't have rigid OOP conventions in mind for Events...

Xymanek:
> I doubt it would be too hard to find events where the source isn't `self` to show Firaxis didn't have rigid OOP conventions in mind for Events...

Some cases I've found:
```uc
class XComGameStateContext_RevealAI extends XComGameStateContext
	native(Core);

local XComGameState_AIGroup GroupState;

EventManager.TriggerEvent('ScamperBegin', GroupState, GroupState, NewGameState);
```
```uc
class XComGameState_AIGroup extends XComGameState_BaseObject
	dependson(XComAISpawnManager, XComGameState_AIUnitData)
	native(AI)
	config(AI);

function MarkGroupSighted(XComGameState_Unit SightingUnit, XComGameState NewGameState)
{
	`XEVENTMGR.TriggerEvent('EnemyGroupSighted', SelfObj, SightingUnit, NewGameState);
}

**Comment by pledbrook** [comment]
As I said, I'd accept that argument. I'm more interested in thoughts on the refactoring. Don't want to do it if folks are definitely against the idea.

**Comment by pledbrook** [comment]
The existing code isn't very amenable to the refactoring. Not surprising I guess. So just held my nose and made the tech the event "source".

File: X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersXCom.uc [hunk 1]
```diff
@@ -6931,15 +6931,13 @@ function bool TriggerCanTechBeInspired(XComGameState_Tech TechState)
...
-	Tuple.Data[0].Kind = XComLWTVObject;
-	Tuple.Data[0].o = TechState;
-	Tuple.Data[1].Kind = XComLWTVBool;
-	Tuple.Data[1].b = true;
+	Tuple.Data.Add(1);
+	Tuple.Data[0].Kind = XComLWTVBool;
+	Tuple.Data[0].b = true;
 
-	`XEVENTMGR.TriggerEvent('CanTechBeInspired', Tuple, self);
+	`XEVENTMGR.TriggerEvent('CanTechBeInspired', Tuple, TechState);
```
Xymanek: Pass the `NewGameState` to the event, it may be important for some checks. Also make the `TriggerCanTechBeInspired` private

pledbrook:
Updated (again)

pledbrook:
Oops, bug found. If you don't have any techs that can be inspired (and at least one that can't be inspired due to the result of the hook), this code enters an infinite loop. Yay!

pledbrook:
Bug fixed. Would like another pair of eyes to give it a once over in case I've made another boo boo.

PR #642, opened 2019-09-30
Allow cosmetic body parts to contain more than just skeletal meshes (#641)

### LINKED ISSUE #641: Allow cosmetic body parts to contain more than just skeletal meshes
This will allow things like particle effects to be attached to body parts

Xymanek:
Closes #641

Currently the implementation has a small problem - the effects aren't being removed after existing from cinematic. This can be encountered in 2 places;

* Armory (eg. class intro movie)
* Mission intro

image

Will provide video if needed

cc @robojumper 

P.S. Here is the use of this PR: https://github.com/WOTCStrategyOverhaul/CovertInfiltration/pull/259

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComBodyPartContentAdvanced.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComBodyPartContentAdvanced.uc
@@ -1,0 +1,10 @@
+//-----------------------------------------------------------
+//	Class:	XComBodyPartContentAdvanced
+//	Author: xyman
+//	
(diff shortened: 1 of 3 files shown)


Xymanek:
Turns out the issue was iterator invalidation in `DestroyAttachedActors`. However, while debugging it was found that other logic attaches actors to `XComUnitPawn`, so blanket destroying all of them may not be the best idea. I'll make an array dedicated to tracking actors spawned by this PR

pledbrook:
This is way out of my area of experience, so don't feel at all qualified to review it. Suggest getting someone who knows more to review.

PR #646, opened 2019-10-05
Ensure effects tick correctly after mind control (#643)

### LINKED ISSUE #643: Stun "effect" continues after mind control ends
Here's the scenario: an enemy unit mind controls one of your squad. You then stun that squad member so that they can't do any harm during XCOM's turn. If the mind control effect then wears off before the stun effect, the stunned soldier remains in a "stunned" state (both visually and ability-wise), but can move. And the stunned state is never removed.

To reproduce, load up the following save (requires LW2 Secondaries and Skirmish+ mods), end the turn and see what happens to the mind controlled Grenadier:

[save_BUG Stun and mind control.zip](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/files/3685321/save_BUG.Stun.and.mind.control.zip)

The problem is that `XCGS_Effect.OnGroupTurnTicked` is tied to the player who was in control of the unit at the point that the stun was applied. When the mind control wears off, the player changes back to XCOM and the listener stops firing. On top of that, the listener uses a filter of the `XCGS_AIGroup` which also changes. So there are two reasons it doesn't fire.

pledbrook:
Mind control is a nightmare. The issue here is that many effects use `XCGS_Effect.OnPlayerTurnTicked()` and `OnGroupTurnTicked()` to manage effect behaviour. However, those listeners are typically registered for specific players or unit groups. Mind control changes those, so the registered listeners stop working.

This change swaps those registered listeners for ones that use the correct player/unit group filter when the mind control is lost. In addition, it ensures the effect's `PlayerStateObjectRef` is updated to
the new player, because otherwise the listeners will skip the tick behaviour.

As this is a bit of a brute force technique, it only applies to effects that are listed in `CHHelpers.TickOnMindControlRemovalEffects`.

Fixes #643.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_MindControl.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_MindControl.uc
@@ -126,6 +126,10 @@ simulated function OnEffectRemoved(const out EffectAppliedData ApplyEffectParame
 		}
 	}
 	UpdateAIData(NewGameState, UnitState, false);
+
+	// Start Issue #643
+	`XEVENTMGR.TriggerEvent('MindControlLost', UnitState, UnitState, NewGameState);
+	// End Issue #643
 }
 
 simulated function AddX2ActionsForVisualization(XComGameState VisualizeGameState, out VisualizationActionMetadata ActionMetadata, name EffectApplyResult)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -187,6 +187,8 @@ RunPriorityGroup=RUN_STANDARD
 - `OverrideAbilityIconColor` provides a tuple with the same ID as the
   event and data of the form `[bool IsObjective, string Color]` that allows
   mods to override the color of soldier abilities in the tactical HUD (#400)
+- `MindControlLost` fires whenever a unit stops being mind controlled or hacked. The event
+  passes the affected unit state as both event data and event source (#643)
 
 ### Configuration
 - Added ability to modify default spawn size (#18)
@@ -246,7 +248,6 @@ RunPriorityGroup=RUN_STANDARD
   a pod tries to patrol to any of them (#508)
 - Make disorient reapply to disoriented units so that things like flashbangs can
   still remove overwatch from disoriented units (#475)
-
 
 ## Miscellaneous
 


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_MindControl.uc
```diff
@@ -126,7 +134,173 @@ simulated function OnEffectRemoved(const out EffectAppliedData ApplyEffectParame
...
+		EventMgr.RegisterForEvent(EffectObj, 'PlayerTurnEnded', EffectState.OnPlayerTurnTicked, ELD_OnStateSubmitted,, EndPlayer );
+	}
+	else if( EffectTemplate.WatchRule == eGameRule_UnitGroupTurnBegin )
+	{
+		EventMgr.UnRegisterFromEvent(EffectObj, 'UnitGroupTurnBegun');
+		EventMgr.RegisterForEvent(EffectObj, 'UnitGroupTurnBegun', EffectState.OnGroupTurnTicked, ELD_OnStateSubmitted,, UnitState.GetGroupMembership() );
+	}
+	else if( EffectTemplate.WatchRule == eGameRule_UnitGroupTurnEnd )
+	{
+		EventMgr.UnRegisterFromEvent(EffectObj, 'UnitGroupTurnEnded');
```
pledbrook: I'm wondering whether these listeners need to be removed. They typically won't trigger after the team swap anyway. I guess there's the potential issue of Mind Control -> Loss of MC -> MC again -> Loss of MC.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_MindControl.uc
```diff
@@ -126,7 +134,173 @@ simulated function OnEffectRemoved(const out EffectAppliedData ApplyEffectParame
...
+// stunned state.
+//
+// The following code transfers effect tick handling on mind control loss to the
+// appropriate event with the appropriate player/unit group filter. It's a bit of a
+// sledgehammer and can cause issues if it removes event listeners it shouldn't, which
+// is why it will only apply to configured effects.
+//
+// Original Pavonis comment follows.
+//
+//PI : On team change we need to perform updates for effects on or sourced by this unit that need to tick. These
```
pledbrook: Don't need the PI here as Pavonis is mentioned on previous line. Also consider using "vanilla highlander" instead of Pavonis.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_MindControl.uc
```diff
@@ -126,7 +134,173 @@ simulated function OnEffectRemoved(const out EffectAppliedData ApplyEffectParame
...
+		if (class'CHHelpers'.default.TickOnMindControlRemovalEffects.Find(EffectState.GetX2Effect().EffectName) == INDEX_NONE)
+		{
+			// Skip this effect
+			continue;
+		}
+		UpdateOneTickingEffect(UnitState, EffectState, StartPlayer, EndPlayer, NewGameState, true);
+	}
+}
+
+// PI Added: Update a single effect either on this unit or sourced by this unit.
```
pledbrook: Use "Original vanilla highlander comment" or similar. "PI" may be too obscure for folks.

pledbrook:
Mind control + stun behaviour is so flaky, I decided to go a different route that only requires a 'MindControlLost' event at the end of `X2Effect_MindControl.OnEffectRemoved()`. Should be much easier to approve this change 😆

PR #655, opened 2019-10-17
modbuddy final_release bugfix

MrNiceUK:
There's a bug in the `.x2proj` file, which puts the `final_release` tag inside the gamepath quotes! Should be a Hot Patch!
#656


Musashi1584:
Commit got lost by rebase or force push. Fixed in https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/commit/43c4582cc7787575de8ce66fd1211b62071aa390

Issue #659, opened 2019-10-19
Dual arms unconditionally hide arm deco ("shoulders")


robojumper:
This is something that already existed on the UI level, bit a change was made on the pawn level in https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/351 and landed in v1.19.0. It seems like this caused breakage for users of Unrestricted Customization, which bypasses UI level checks, see <https://www.reddit.com/r/xcom2mods/comments/dk2c45/please_help_character_pool_broken_my_soldiers/>.

In vanilla, this change would be entirely correct. The following picture shows dual arms on the left, and arm deco ("shoulders") on the right. All single arms basically only customize the forearm (not shown):

![grafik](https://user-images.githubusercontent.com/14299449/67147066-970e7e80-f291-11e9-96fa-53d38a1336ac.png)

This is a bad situation because the point of UC is creating setups we cannot algorithmically verify to not conflict, but UC only disables UI level checks, not pawn level checks. Tagging as HL bug because we want to investigate what to do here.

MrNiceUK:
Ok, well simplest would be to put *back* the vanilla code which blanked out the shoulder slots when dual arms are selected in the UI, and *remove* the Pawn code which hides shoulders when dual arms are in use (ah, for a sane world where the arm archetype had a flag for hiding shoulders or not, as it does for forearms!)

Issue #350, opened 2018-11-21
Parts Validation on Torso Change


MrNiceUK:
As well as updating based on the specific body part selection changed, `XComCharacterCustomization::OnCategoryValueChange()` validates related body parts, and switches them to valid parts. However, the validation on Torso changes is stuck in an Anarchy's Children release world, without the proliferation of other deco slots considered, most of which use FilterByTorsoAndArmorMatch so should be checked for validity on torso changes. Even for arm selection, it makes assumptions about validity which may be true with just the base game + AC, but may not hold for customized setups and mods. Specifically, it implicitly assumes that all  separate left/right arm part are valid for all Torsos, and also that if the current joint arms part is invalid, then all other joint arms parts must also be invalid, so immediately swaps to using separate left/right arms, which compounds the previous assumption that since left/right arms are always valid, there will be valid left/right arms to switch to.
In the worst case, where a new torso only has one set of valid arms which are not currently selected, changing to this torso won't automatically switch to them, because the above assumptions don't hold, yet you can't switch to them manually either, since slot selection is disabled for slots which only have one valid option (which the game implicitly assumes are already selected).
Also, there is inconsistent behaviour with forearm parts in general. Joint arms and left/right arms use the same Archetype, and both have a flag to hide forearm parts. In line with this, changing the currently selected forearm decos is also disabled if this flag is set (but the current setting is remembered for when switch to arms without this flag set). However, changing the selection of this slot is also disabled if join arms parts are selected at all. Any currently selected forearm decos are still shown though, if the flag is not set. In the base game, the Soldier character type has no forearm decos available, so this inconsistency doesn't show up without mods.

Iridar:
@MrNiceUK is this issue still open only because it needs docs? If so, could you add some, please?

PR #661, opened 2019-10-19
#659 shoulder hide fix

MrNiceUK:
See #659


MrNiceUK:
That second commit is probably another reason #350 did the way it did, you have to catch when it goes back to dual arms because of torso validation, not just when the user directly selects dual arms!

PR #670, opened 2019-10-26
#461 #321 Minimum Damage through Armor Fix

MrNiceUK:
Fixes the issues identified with PR #461 which implemented Issue #321. See #461 comments for background details.


pledbrook:
Shouldn't this have an entry in CHANGELOG.md? I notice that the original pull request didn't either (tut tut).

Xymanek:
1) Why was this merged when there were requested changes?
2) Still needs a changelog entry

Musashi1584:
I thought the fix is more important than the changelog. IIRC i pinged @MrNiceUK to add a changelog entry.

PR #673, opened 2019-10-27
Make IsPsiOperative() function accept a list of defined Soldier classes

boundir:
This will let defined classes use Psi PCS #602

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
@@ -438,19 +438,3 @@ static function array<name> GetAmbushRiskTemplateNames()
 	return TemplateNames;
 }
 // End Issue #485
-
-// Start issue #602
-simulated static function bool IsAssimilatedPsiClass(Name SoldierClass)
-{
-	local Name ClassName;
-
-	foreach default.CLASS_AS_PSI_CLASS(ClassName)
-	{
-		if(SoldierClass == ClassName)
-		{
-			return true;
-		}
-	}
-	return false;
-}
-// End issue #602
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
@@ -145,6 +145,10 @@ var config bool UseNewPersonnelStatusBehavior;
 var config bool bSkipCampaignIntroMovies;
 // End Issue #543
 
+// Start issue #602
+var config array<name> CLASS_AS_PSI_CLASS;
+// End issue #602
+
 // Start Issue #123
 simulated static function RebuildPerkContentCache() {
 	local XComContentManager		Content;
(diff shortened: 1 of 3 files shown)


robojumper:
Are there any places in the base game where we want to replace `UnitState.GetSoldierClassTemplateName() == 'PsiOperative'` with `UnitState.IsPsiOperative()` now that they aren't equivalent?

X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
```diff
@@ -434,3 +438,19 @@ static function array<name> GetAmbushRiskTemplateNames()
...
+	local Name ClassName;
+
+	foreach default.CLASS_AS_PSI_CLASS(ClassName)
+	{
+		if(SoldierClass == ClassName)
+		{
+			return true;
+		}
+	}
+	return false;
```
MrNiceUK: There's a much shorter way to check an array for a particular value:
```
return default.CLASS_AS_PSI_CLASS.Find(ClassName) != INDEX_NONE;
```

MrNiceUK:
Quick check for `'PsiOperative'` in the code and where it's used to check the soldier class, some MP/Skirmish checks which probably don't matter. In the single player campaign, seem all related to the psi-training mechanic and related issues (for completeness, also the photobooth has a select/case block which goes through the base game classes too).

pledbrook:
There are quite a few significant locations that check for the soldier class directly:

 * `UIFacility_PsiLabSlot.OnPersonnelSelected()`
* `X2DynamicNarrative_DefaultDynamicNarratives.CreateTemplates()` (perhaps not important)
* `X2StrategyElement_DefaultStaffSlots.ShouldDisplayPsiChamberSoldierToDoWarning()`
* `X2StrategyElement_DefaultStaffSlots.IsUnitValidForPsiChamberSoldierSlot()` (but can be overridden by event)
* `X2StrategyElement_DefaultStaffSlots.GetPsiChamberSoldierBonusDisplayString()`
* `XComGameState_HeadquartersProjectPsiTraining.SetProjectFocus()`
* `XComGameState_HeadquartersXCom.MeetsSoldierRankGates()`
* `XComGameState_HeadquartersXCom.GetPsiTrainingEvents()`
* `XComGameState_MissionSite.ResumePsiOperativeTraining()`
* `XComGameState_Unit.HasAvailablePerksToAssign()`
* `XComGameState_Unit.RankUpSoldier()`
* `XComGameStateContext_HeadquartersOrder.CompleteUnitHealing()`
* `XComGameStateContext_HeadquartersOrder.CompletePsiTraining()`
* `XComGameStateContext_StrategyGameRule.PostMissionUpdatePsiOperativeTraining()`
* `XComHQPresentationLayer.ShowPromotionUI()`
* `XComUnitPawnNativeBase.PassesPlayLimitsTestUnit()`

Got to love the consistency.

pledbrook:
So really, this is a question of what the actual intent of the change is. If it's simply to allow soldier classes to use the Psi PCS, then I would go with what I suggested in #604: add a configurable array of soldier classes that can explicitly use that PCS. The proposed change in this pull request is much more fundamental and requires a review of usages of `IsPsiOperative()` and the check for `'PsiOperative'` soldier class.

boundir:
I don't know why we wouldn't want the change to affect `UISoldierHeader.uc` where it seems to be used to display stats.
I made this change because it didn't seem to affect facility training or functions specifically related to 'PsiOperative' class. 

Maybe what we want then is to replace calls to `IsPsiOperative()` where we want to allow access to other classes. That way we could safely replace `UnitState.GetSoldierClassTemplateName() == 'PsiOperative'` with `UnitState.IsPsiOperative()` as robo suggested.

Xymanek:
I think this PR has too many unresolved comments, I do not consider it ready to be part of the stable CHL yet.

Suggestion: revert

robojumper:
This change was **reverted** on master.

Issue #676, opened 2019-11-13
Add an event to XCGS_Unit::ApplyBestGearLoadout


Iridar:
Add an **'OnBestGearLoadoutApplied'** event to XCGS_Unit::ApplyBestGearLoadout so that mods can change Unit's loadout after it has been applied. Would be used by RPGO's Random Classes Second Wave Option. 
Can be potentially used by any mod that want to add arbitrary equipment restrictions via CanAddItemToInventory hook.

Iridar:
PR merged, closing issue.

PR #680, opened 2019-11-14
Pull Request for issue #676

Iridar:
Added 'OnBestGearLoadoutApplied' event to `XCGS_Unit::ApplyBestGearLoadout()`.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -10746,7 +10746,7 @@ function ApplyBestGearLoadout(XComGameState NewGameState)
 	ValidateLoadout(NewGameState);
 	
 	// Issue #676 Start
-	`XEVENTMGR.TriggerEvent('OnBestGearLoadoutApplied', self, self, none);
+	`XEVENTMGR.TriggerEvent('OnBestGearLoadoutApplied', self, self, NewGameState);
 	// Issue #676 End
 }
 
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -10744,6 +10744,10 @@ function ApplyBestGearLoadout(XComGameState NewGameState)
 
 	// Always validate the loadout after upgrading everything
 	ValidateLoadout(NewGameState);
+	
+	// Issue #676 Start
+	`XEVENTMGR.TriggerEvent('OnBestGearLoadoutApplied', self, self, none);
```
Xymanek: Why not pass the `NewGameState`?

Iridar: Wasn't aware I need to do it manually. It wasn't necessary for my use case, but I'll add it just in case.

Xymanek: It would allow listeners to subscribe to this event with `ELD_OnStateSubmitted` deferral. May be useful for someone

Iridar: Uhhhhhhhh
That's the deferral I've been using. Works A-OK as far as I can tell.

Xymanek: If the third argument to `TriggerEvent` is `none`, then only `ELD_Immediate` will ever work

Iridar: I'm pretty sure this goes against my experience, but maybe I've been using `ELD_Immediate ` after all, I'll get back to you on that.

Iridar: Yep, I've been using `ELD_OnStateSubmitted` with no Game State being passed when triggering the Event, and it was working absolutely fine from what I saw ingame and in logs. Though I was looking at only my log lines, so if there were any warnings from the game, I might have missed them.

But the "If the third argument to TriggerEvent is none, then only ELD_Immediate will ever work" statement clearly doesn't apply to my use case.

Issue #694, opened 2019-12-19
Add Event OnItemEquipped & OnItemUnEquipped


VampireMonkey:
The event would return the Item Instance, however i'm not sure where the best place would be to put it.

in UIArmory_Loadout.EquipItem or CHItemSlot.AddItemToSlot

likewise OnItemUnEquipped could be in CHItemSlot.RemoveItemFromSlot

I think CHItemSlot would be best

VampireMonkey:
Added Event ItemAddedToSlot & ItemRemovedFromSlot for this Issue in Pullrequest #703

PR #703, opened 2019-12-23
Added Event ItemAddedToSlot & ItemRemovedFromSlot for Issue #694

VampireMonkey:
Added Event ItemAddedToSlot & ItemRemovedFromSlot for Issue #694

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHItemSlot.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHItemSlot.uc
@@ -123,6 +123,10 @@ function AddItemToSlot(XComGameState_Unit Unit, XComGameState_Item Item, XComGam
 	if (AddItemToSlotFn != none)
 	{
 		AddItemToSlotFn(self, Unit, Item, NewGameState);
+		
+		// Start Issue #694
+		`XEVENTMGR.TriggerEvent('ItemAddedToSlot', Item, self, NewGameState);
+		// End Issue #694
 	}
 }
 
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/CHItemSlot.uc
```diff
@@ -123,6 +123,10 @@ function AddItemToSlot(XComGameState_Unit Unit, XComGameState_Item Item, XComGam
 	if (AddItemToSlotFn != none)
 	{
 		AddItemToSlotFn(self, Unit, Item, NewGameState);
+		
+		// Start Issue #694
+		`XEVENTMGR.TriggerEvent('ItemAddedToSlot', Item, self, NewGameState);
```
Xymanek: What happens if `AddItemToSlotFn == none`? No event?

VampireMonkey: CanAddItemToSlot :
`Called from XComGameState_Unit::AddItemToInventory if the slot is templated and the Game doesn't know what to do`

SlotIsTemplated : 
`Returns true if and only if this slot is within the Templated Slot range and has an existing template`

Hmm, might not be the right place to put it then but there doesn't seem to be a better place.

Why aren't the vanilla slots using the highlander template system?

robojumper:
Hotl3looded pointed out on Discord that the changelog entry is there, but the events are missing (73644569b93e2bb2b1e2bf1dbec305db33c4e745 with merge b3dffc07d72ce5adc3c5c74c92717a4bfc0a4588). It seems like this PR didn't actually land.

PR #705, opened 2019-12-23
Add More Generic Macros

VampireMonkey:
Like `XCOMHQ, but also `RESISTANCEHQ, `ALIENHQ and so on

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -367,3 +367,6 @@ RunPriorityGroup=RUN_STANDARD
   UIScreen (#341)
 - Appearances now update correctly when a part change differs only by material override (#354)
 - All relevant body parts are now correctly validated when the torso is changed. (#350)
+
+### Macros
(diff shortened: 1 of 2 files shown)


Xymanek:
Ummmm... the file is `Src/Core/Globals.uci` (not `Src/Globals.uci`)?

VampireMonkey:
Can you move it? I can't make a new folder

pledbrook:
Are there any issues with adding more entries to *Globals.uci*? Wondering if just moving the file will unblock this PR.

VampireMonkey:
> Are there any issues with adding more entries to _Globals.uci_? Wondering if just moving the file will unblock this PR.

I can't see any problems with adding more entries

Iridar:
Closing this PR, because it:
1. Wasn't fixed in nearly a year.
2. Doesn't have a tracking Issue. 
That said, I do support the idea of adding more macros, so I'm gonna open an Issue about that.

PR #725, opened 2020-01-17
Pull request for Issue #724.

### LINKED ISSUE #724: Pistol Overwatch is not displayed under Unit Flag
Looks like this function in `UIUnitFlag` is responsible for displaying the Overwatch "Eye" icon under the unit flag, and it makes an assumption that the soldier will be using only 'Overwatch' reserve action points for Overwatch, which is not the case for at least Pistol Overwatch, and potentially for any other custom Overwatch Action Point used by a weapon equipped on the soldier.

```
//This function will be spammed, so please only send changes to flash.
simulated function RealizeOverwatch(optional XComGameState_Unit NewUnitState = none)
{
    if( NewUnitState == none )
    {
        NewUnitState = XComGameState_Unit(History.GetGameStateForObjectID(StoredObjectID));
    }

    if( NewUnitState != none )
    {

        AS_SetOverwatchIcon(NewUnitState.ReserveActionPoints.Find('Overwatch') > -1);
    }
}
```
The ideal solution would be to scan the soldier's inventory for Weapon Templates and check if the unit has any of their ` Template.OverwatchActionPoint` present on the soldier, but since the dev comment indicates that "this function will be spammed a lot", it was decided to keep things humble, and expose the list of Reserve AP we're gonna check for to a config array.

None of the weapon mods I have use any other kind of custom overwatch point, so the array will include only `Overwatch` and `PistolOverwatch`  by default. I assume there is no negative repercussion to performance as opposed to just checking specifically for these two AP in the hardcoded way.

Iridar:
The list of Reserve Action Points that make the soldier display the Overwatch "Eye" icon under the Unit Flag is now configurable in XComUI.ini. Now includes Pistol Overwatch by default.

Closes #724

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
@@ -9,7 +9,7 @@
 //--------------------------------------------------------------------------------------- 
 
 class UIUnitFlag extends UIPanel
-	dependson(XComGameState_Unit);
+	dependson(XComGameState_Unit) config(UI);
 
 enum EUnitFlagTargetingState
 {
@@ -88,6 +88,9 @@ var localized string m_strReinforcementsTitle;
 var localized string m_strReinforcementsBody;
 
 var int VisualizedHistoryIndex;
+
+//	Variable for Issue #724
+var config(UI) array<name> ValidReserveAPForUnitFlag;
 
 // kUnit, the unit this flag is associated with.
 simulated function InitFlag(StateObjectReference ObjectRef)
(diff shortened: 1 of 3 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
```diff
@@ -89,6 +89,9 @@ var localized string m_strReinforcementsBody;
 
 var int VisualizedHistoryIndex;
 
+//	Variable for Issue #724
+var config(UI) array<name> ValidReserveAPForUnitFlag;
```
Xymanek: I think it's better if the variable went into `CHHelpers`. Reason: config and inheritance is finicky and we don't want to screw up if someone extends this class and adds their own config to the child class

Iridar:
Closing PR because I need to nuke my Fork again.

PR #595, opened 2019-08-02
Allow mods to prevent certain CAs from being randomly spawned

Xymanek:
Closes #594

PR #738, opened 2020-01-21
Small improvements to #621 and #595 for 1.20.0 beta

Xymanek:
Private-d non-public-API functions from #621 
Fixed copy paste comment screw up in #595


pledbrook:
Isn't this a breaking change now, at least the switch to `private`?

Xymanek:
Well, this is awkward

pledbrook:
I've set this to Waiting on Author since it seems like it's a breaking change, as I said. If it's not, happy to change it back. But it certainly shouldn't be Ready for Merge with this question mark hanging over it.

robojumper:
Let's just go with this -- we can turn them `private` later if wanted.

PR #740, opened 2020-01-21
Issue #735 Extend HasGrenadePocket() and HasAmmoPocket()

### LINKED ISSUE #735: Extend HasGrenadePocket() and HasAmmoPocket()
On popular request i propose to extend the ability check in `HasGrenadePocket()` and `HasAmmoPocket()`.

Currently HasSoldierAbility(CheckAbility) is used which just checks for earned soldier abilities (afaik).

This check should be expanded to include item granted abilities and character template abilities.

Additionally an event should be triggered to allow modders to further customize conditions.

Related: https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/issues/217

### LINKED ISSUE #753: Fire an event when a dark event expires (is deactivated)
`XCGS_DarkEvent` has an `OnDeactivated()` function that's called when a dark event expires or is otherwise deactivated, but there is no easy way for mods to hook into this. The only solution I can think of is to wrap every Dark Event template's `OnDeactivatedFn` with a wrapper that does whatever the mod wants. This is a pretty obtuse solution to a problem that could be easily solved with a simple event fired at the end of `OnDeactivated()`.

One idea is to have a UI notification pop up when a dark event expires, but I'm sure there could be other features that would benefit from the hook.

Musashi1584:
Closes #735

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -3848,7 +3848,7 @@ function bool HasGrenadePocket()
 /// Finally the event OverrideHasAmmoPocket is triggered that allows mods to override the final result
 ///
 /// ```unrealscript
-/// EventID: OverrideHasGrenadePocket
+/// EventID: OverrideHasAmmoPocket
 /// EventData: XComLWTuple {
 /// 	Data: [
 /// 	  inout bool bHasAmmoPocket
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -3783,6 +3783,7 @@ function bool HasGrenadePocket()
 	local XComLWTuple Tuple;
 	local name Ability;
 	local bool bHasGrenadePocket;
+	// End Variables for Issue #735
 
 	// Start Issue #735 (1/2)
 	foreach class'X2AbilityTemplateManager'.default.AbilityUnlocksGrenadePocket(CheckAbility)
@@ -3867,6 +3868,7 @@ function bool HasAmmoPocket()
 	local XComLWTuple Tuple;
 	local name Ability;
 	local bool bHasAmmoPocket;
+	// End Variables for Issue #735 (2/2)
 
 	// Start Issue #735 (2/2)
 	foreach class'X2AbilityTemplateManager'.default.AbilityUnlocksAmmoPocket(CheckAbility)
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -3842,7 +3842,7 @@ function bool HasGrenadePocket()
 	// End Issue #735 (1/2)
 }
 
-/// HL-Docs: feature:OverrideHasAmmorPocket; issue:735; tags:itemslots,strategy
+/// HL-Docs: feature:OverrideHasAmmoPocket; issue:735; tags:itemslots,strategy
 /// Extends the ability check in `HasAmmoPocket()` for the config array `AbilityUnlocksAmmoPocket` (`XComGameData.ini`) to item granted abilities
 /// and abilities granted by the character template.
 /// Finally the event AbilityUnlocksAmmoPocket is triggered that allows mods to override the final result
@@ -3920,7 +3920,7 @@ function bool HasAmmoPocket()
 	Tuple.Data[0].kind = XComLWTVBool;
 	Tuple.Data[0].b = bHasAmmoPocket;
 
-	`XEVENTMGR.TriggerEvent('OverrideHasGrenadePocket', Tuple, self, none);
+	`XEVENTMGR.TriggerEvent('OverrideHasAmmoPocket', Tuple, self, none);
 	
 	return Tuple.Data[0].b;
 	// End Issue #735 (2/2)
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -3799,7 +3799,7 @@ function bool HasAnyOfTheAbilitiesFromCharacterTemplate(array<name> AbilitiesToC
 }
 // End helper for Issue #735
 
-/// HL-Docs: feature:OverrideHasGrenadePocket; issue:735; tags:itemslots,strategy
+/// HL-Docs: feature:OverrideHasGrenadePocket; issue:735; tags:loadoutslots,strategy
 /// Extends the ability check in `HasGrenadePocket()` for the config array `AbilityUnlocksGrenadePocket` (`XComGameData.ini`) to item granted abilities
 /// and abilities granted by the character template.
 /// Finally the event OverrideHasGrenadePocket is triggered that allows mods to override the final result
@@ -3854,7 +3854,7 @@ function bool HasGrenadePocket()
 	// End Issue #735 (1/3)
 }
 
-/// HL-Docs: feature:OverrideHasAmmoPocket; issue:735; tags:itemslots,strategy
+/// HL-Docs: feature:OverrideHasAmmoPocket; issue:735; tags:loadoutslots,strategy
 /// Extends the ability check in `HasAmmoPocket()` for the config array `AbilityUnlocksAmmoPocket` (`XComGameData.ini`) to item granted abilities
 /// and abilities granted by the character template.
 /// Finally the event OverrideHasAmmoPocket is triggered that allows mods to override the final result
@@ -3910,7 +3910,7 @@ function bool HasAmmoPocket()
 }
 
 // Check is for squad select UI
-/// HL-Docs: feature:OverrideHasExtraUtilitySlot; issue:735; tags:itemslots,strategy
+/// HL-Docs: feature:OverrideHasExtraUtilitySlot; issue:735; tags:loadoutslots,strategy
 /// Extends the ability check in `HasExtraUtilitySlot()` for the config array `AbilityUnlocksExtraUtilitySlot` (`XComGameData.ini`) to item granted abilities
 /// and abilities granted by the character template.
 /// Finally the event OverrideHasExtraUtilitySlot is triggered that allows mods to override the final result
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -3758,7 +3758,7 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
 	return true;
 }
 
-// Start helper for Issue #735
+// Start helper methods for Issue #735
 function bool HasAnyOfTheAbilitiesFromInventory(array<name> AbilitiesToCheck)
 {
 	local array<XComGameState_Item> CurrentInventory;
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3758,28 +3758,172 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
...
+	Tuple.Data[0].kind = XComLWTVBool;
+	Tuple.Data[0].b = bHasGrenadePocket;
+
+	`XEVENTMGR.TriggerEvent('OverrideHasGrenadePocket', Tuple, self, none);
+	
+	return Tuple.Data[0].b;
+	// End Issue #735 (1/2)
+}
+
+/// HL-Docs: feature:OverrideHasAmmorPocket; issue:735; tags:itemslots,strategy
```
Xymanek: OverrideHasAmmo**r**Pocket

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3758,28 +3758,172 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
...
+/// Extends the ability check in `HasAmmoPocket()` for the config array `AbilityUnlocksAmmoPocket` (`XComGameData.ini`) to item granted abilities
+/// and abilities granted by the character template.
+/// Finally the event AbilityUnlocksAmmoPocket is triggered that allows mods to override the final result
+///
+/// ```unrealscript
+/// EventID: OverrideHasGrenadePocket
```
Xymanek: Copy paste slip up

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3758,28 +3758,172 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
...
+		}
+	}
+
+	Tuple = new class'XComLWTuple';
+	Tuple.Id = 'OverrideHasAmmoPocket';
+	Tuple.Data.Add(1);
+	Tuple.Data[0].kind = XComLWTVBool;
+	Tuple.Data[0].b = bHasAmmoPocket;
+
+	`XEVENTMGR.TriggerEvent('OverrideHasGrenadePocket', Tuple, self, none);
```
Xymanek: Copy paste slip up

Musashi1584: fixed

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
Musashi1584: fixed

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
Musashi1584: fixed

Musashi1584:
> Are we sure that making it by default pull from item abilities array is backwards compatible?

I see no scenarion why it shouldnt be. If someone manually added an ability from an item to the config array the only difference is that it works now as he expected it.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3758,28 +3758,172 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
...
+/// 	  inout bool bHasGrenadePocket
+///     ]
+/// }
+/// EventSource: XComGameState_Unit
+/// ```
 function bool HasGrenadePocket()
 {
 	local name CheckAbility;
 
+	// Variables for Issue #735 (1/2)
```
pledbrook: Would you mind adding an `// End` line or other boundary marker so we know how many of the following variables are associated with issue #735? Don't want to have to look at the associated commit just to find out that information.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3758,28 +3758,172 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
...
+		{
+			bHasGrenadePocket = true;
+			break;
+		}
+	}
+	
+	if (!bHasGrenadePocket)
+	{
+		CurrentInventory = GetAllInventoryItems();
+		foreach CurrentInventory(InventoryItem)
```
pledbrook: Might a
```
for (idx = 0; i < CurrentInventory.Length && !bHasGrenadePocket)
```
be slightly more comprehensible than using `break`s? Not a deal breaker, it's just the following code is not super easy to grok.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3758,28 +3758,172 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
...
+/// 	  inout bool bHasAmmoPocket
+///     ]
+/// }
+/// EventSource: XComGameState_Unit
+/// ```
 function bool HasAmmoPocket()
 {
 	local name CheckAbility;
 
+	// Variables for Issue #735 (2/2)
```
pledbrook: Same suggestion for an `// End Issue` line or similar.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
pledbrook: Same comment applies to `HasAmmoPocket()`. Perhaps there isn't a nice way to do this without factoring out some additional functions. Go with whatever you prefer.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
Musashi1584: Below should only come variables of the issue. If another issue adds variables they should start with a new comment

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
MrNiceUK: Depends what you're used to, keeping the for condition simple (or entirely implicit with foreach) then breaking when you find an object of interest is a pattern I've seen often, including in Firaxis X2 code, for it to grok fine

MrNiceUK: Depends what you're used to, keeping the for condition simple (or entirely implicit with foreach) then breaking when you find an object of interest is a pattern I've seen often, including in Firaxis X2 code, for it to grok fine

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
pledbrook: Let me put it another way: if I come across this code 3 months from now, could I be confident at first glance that those variables are all linked to that one issue? No, because I don't know who wrote the code or what conventions they were using at the time. Anyway, if you'd prefer to leave it as is, no worries.

Musashi1584:
@pledbrook 
I refactored the code a bit and added the expanded checks and the event `HasExtraUtilitySlot` as well. Can you do another review when you have some time to spare?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3758,50 +3758,221 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
...
+/// Finally the event OverrideHasGrenadePocket is triggered that allows mods to override the final result
+///
+/// ```unrealscript
+/// EventID: OverrideHasGrenadePocket
+/// EventData: XComLWTuple {
+/// 	Data: [
+/// 	  inout bool bHasGrenadePocket
+///     ]
+/// }
+/// EventSource: XComGameState_Unit
```
pledbrook: I would add a doc line for `NewGameState`, i.e. `NewGameState: yes`, or `NewGameState: no`. This is just the let mod authors know whether the event is passed with a `NewGameState` instance or not (if they use `ELD_Immediate`) that they can use to make more game state changes.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3758,50 +3758,221 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
...
+	
+	if (!bHasGrenadePocket)
+	{
+		bHasGrenadePocket = HasAnyOfTheAbilitiesFromInventory(class'X2AbilityTemplateManager'.default.AbilityUnlocksGrenadePocket);
+	}
+
+	if (!bHasGrenadePocket)
+	{
+		bHasGrenadePocket = HasAnyOfTheAbilitiesFromCharacterTemplate(class'X2AbilityTemplateManager'.default.AbilityUnlocksGrenadePocket);
+	}
```
pledbrook: Would it make sense to have an additional public function that will check whether a unit has any (and maybe one for all) abilities from any source? I was thinking that something like Quick End Turn would benefit from a function that could be used to determine whether a unit has a Steady Weapon ability, regardless of where it came from.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3758,50 +3758,221 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
...
+
+	Tuple = new class'XComLWTuple';
+	Tuple.Id = 'OverrideHasGrenadePocket';
+	Tuple.Data.Add(1);
+	Tuple.Data[0].kind = XComLWTVBool;
+	Tuple.Data[0].b = bHasGrenadePocket;
 
+	`XEVENTMGR.TriggerEvent('OverrideHasGrenadePocket', Tuple, self, none);
+	
+	return Tuple.Data[0].b;
```
pledbrook: I would prefer that this block of code were put into a separate function to isolate the tuple boilerplate from the real meat of the function. The tuple stuff is just a distraction and doesn't help readers understand what the `HasGrenadePocket()` is doing.

With this change and the one suggested above, the function would look like:
```
function bool HasGrenadePocket()
{
    local bool bHasGrenadePocket;

    bHasGrenadePocket = HasAnyOfTheAbilitiesFromAnySource(class'X2AbilityTemplateManager'.default.AbilityUnlocksGrenadePocket);
    return TriggerOverrideHasGrenadePocket(bHasGrenadePocket);
}
```
In fact, it could be made a one liner, but that may be a bit too over the top.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3758,50 +3758,221 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
...
+/// Finally the event OverrideHasAmmoPocket is triggered that allows mods to override the final result
+///
+/// ```unrealscript
+/// EventID: OverrideHasAmmoPocket
+/// EventData: XComLWTuple {
+/// 	Data: [
+/// 	  inout bool bHasAmmoPocket
+///     ]
+/// }
+/// EventSource: XComGameState_Unit
```
pledbrook: Same comments for HasGrenadePocket apply to HasAmmoPocket of course.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3758,50 +3758,221 @@ function bool MeetsAbilityPrerequisites(name AbilityName)
...
+/// Finally the event OverrideHasExtraUtilitySlot is triggered that allows mods to override the final result
+///
+/// ```unrealscript
+/// EventID: OverrideHasExtraUtilitySlot
+/// EventData: XComLWTuple {
+/// 	Data: [
+/// 	  inout bool bHasExtraUtilitySlot
+///     ]
+/// }
+/// Even
```
pledbrook: Typo or copy/paste error? Assume this should be 3 backticks to close off the code block.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
pledbrook: In fact, the tuple boilerplate is the same for all three cases. The only difference is the event name. So you really only need one `TriggerHasPocketOfTypeEvent(name EventName)` function that can be called from all three places.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
MrNiceUK: Just to note, while in Tactical `HasSoldierAbility()` already does that (if the second parameter is true, but that's the default value for it anyway). As the comment on it says, that uses `FindAbility()` behind the scenes, which itself uses the `Abilities` array, which is only populated during tactical. Quick reminder, the whole issue here is that the `Has...Pocket()` methods have to work from loadout/armory too.

pledbrook: OK, good to know, thanks. I still think `HasAny/AllTheAbilitiesFromAnySource()` functions would be useful, if only to be reused from the three places in this pull request.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
Musashi1584: I think my latest changes address all change requests. Please check and resolve if everything looks good.

Iridar:
Not sure if it's of much of a concern, but `HasAnyOfTheAbilitiesFromInventory` does not check abilities attached to weapon upgrades.

PR #744, opened 2020-01-23
Pull Request for Issue #724

### LINKED ISSUE #724: Pistol Overwatch is not displayed under Unit Flag

Iridar:
~~Fix for issue #724 is implemented and documented, but don't merge the PR just yet, as I'm still adding docs for my previous issues.~~

Fixes #724

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
@@ -1712,7 +1712,7 @@ simulated function RealizeOverwatch(optional XComGameState_Unit NewUnitState = n
 		/// Compatibility: If you override `UIUnitFlag::RealizeOverwatch`, your code may undo this change.
 		foreach class'CHHelpers'.default.ValidReserveAPForUnitFlag(OverwatchActionPointName)
 		{	
-			if (NewUnitState.ReserveActionPoints.Find(OverwatchActionPointName) > -1)
+			if (NewUnitState.ReserveActionPoints.Find(OverwatchActionPointName) > INDEX_NONE)
 			{
 				AS_SetOverwatchIcon(true);
 				return;
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
@@ -1703,7 +1703,7 @@ simulated function RealizeOverwatch(optional XComGameState_Unit NewUnitState = n
 		/// and also allows mods that add weapons with custom Overwatch Action Point to specify that 
 		/// action point name in XComGame.ini:
 		///
-		/// ```unrealscript
+		/// ```ini
 		/// [XComGame.CHHelpers]
 		/// +ValidReserveAPForUnitFlag = "overwatch"
 		/// +ValidReserveAPForUnitFlag = "pistoloverwatch"
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
@@ -1684,7 +1684,10 @@ simulated function Remove()
 
 //This function will be spammed, so please only send changes to flash.
 simulated function RealizeOverwatch(optional XComGameState_Unit NewUnitState = none)
-{
+{	
+	//	Variable for Issue #724
+	local name OverwatchActionPointName;
+
 	if( NewUnitState == none )
 	{
 		NewUnitState = XComGameState_Unit(History.GetGameStateForObjectID(StoredObjectID));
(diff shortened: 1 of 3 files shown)


Xymanek:
Please do not introduce new features and docs for old features in same PR.
Please make a separate PR for docs for old issues

Iridar:
As you command. Then this PR can be merged at y'all's leisure :)

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
```diff
@@ -1684,16 +1684,42 @@ simulated function Remove()
...
-		AS_SetOverwatchIcon(NewUnitState.ReserveActionPoints.Find('Overwatch') > -1);
+		//	Start Issue #724
+		/// HL-Docs: feature:DisplayCustomOverwatchActionPointOnUnitFlag; issue:724; tags:tactical,compat
+		/// The base XCOM 2 behavior is to show an Overwatch "eye" icon under the unit flag
+		/// only if the soldier has at least one "overwatch" Reserve Action Point. 
+		/// The icon is not displayed for Pistol Overwatch. This change addresses that issue,
+		/// and also allows mods that add weapons with custom Overwatch Action Point to specify that 
+		/// action point name in XComGame.ini:
+		///
+		/// ```unrealscript
```
Xymanek:
```suggestion
		/// ```ini
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
```diff
@@ -1684,16 +1684,42 @@ simulated function Remove()
...
+		/// [XComGame.CHHelpers]
+		/// +ValidReserveAPForUnitFlag = "overwatch"
+		/// +ValidReserveAPForUnitFlag = "pistoloverwatch"
+		/// ```
+		///
+		/// Compatibility: If you override `UIUnitFlag::RealizeOverwatch`, your code may undo this change.
+		foreach class'CHHelpers'.default.ValidReserveAPForUnitFlag(OverwatchActionPointName)
+		{	
+			if (NewUnitState.ReserveActionPoints.Find(OverwatchActionPointName) > -1)
```
Xymanek:
```suggestion
			if (NewUnitState.ReserveActionPoints.Find(OverwatchActionPointName) > INDEX_NONE)
```

Iridar: That's what the original code used, I assumed it was out of a performance concern and left it as is, but if you like.

Xymanek: From `Object.uc`:
```
const INDEX_NONE = -1;
```

Both "versions" will result in same thing in the compiled code, but for reading it's better to use the "named version"

pledbrook:
Is this approved then?

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUnitFlag.uc
```diff
@@ -1684,16 +1684,42 @@ simulated function Remove()
...
 	{
 		NewUnitState = XComGameState_Unit(History.GetGameStateForObjectID(StoredObjectID));
 	}
 
 	if( NewUnitState != none )
 	{
-
-		AS_SetOverwatchIcon(NewUnitState.ReserveActionPoints.Find('Overwatch') > -1);
+		//	Start Issue #724
+		/// HL-Docs: feature:DisplayCustomOverwatchActionPointOnUnitFlag; issue:724; tags:tactical,compat
```
pledbrook: We will hopefully have a standard approach to documenting config options at some point, but that shouldn't hold up this PR.

pledbrook: Oh, but that "compat" should definitely be a "compatibility" now.

Iridar: Done.

Iridar:
Approved by @Musashi1584 and @pledbrook , so I assume it's ready for merge.

PR #747, opened 2020-01-24
Add config option to take damage on armor before shields (#743)

### LINKED ISSUE #743: Add config option to have damage hit armor before ablative
Similar to #321 except I'm going to do a decent job with it this time
I'll also add documentation for this and #321 because for some reason that issue is still open

cannonfodder1:
Closes #743 
Tested with guns, grenades, and melee. Tested with shredding and without shredding. Tested with psi and bluescreens. Tested on ADVENT and XCOM. Everything works fine, I think it's good for review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -251,6 +251,8 @@ RunPriorityGroup=RUN_STANDARD
   for will loss purposes (#44)
 - Added option to mitigate all weapon damage using armor instead of always taking at
   least 1 damage (#321)
+- Added option to reduce incoming damage using the target's armor before the
+  target's shields/ablative HP are damaged (#743)
 - Able to mark custom targeting methods as `RequiresTargetingActivation` for controller input (#476)
 
 ### Modding Exposures
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -251,8 +251,6 @@ RunPriorityGroup=RUN_STANDARD
   for will loss purposes (#44)
 - Added option to mitigate all weapon damage using armor instead of always taking at
   least 1 damage (#321)
-- Added option to reduce incoming damage using the target's armor before the
-  target's shields/ablative HP are damaged (#743)
 - Able to mark custom targeting methods as `RequiresTargetingActivation` for controller input (#476)
 
 ### Modding Exposures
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -251,6 +251,8 @@ RunPriorityGroup=RUN_STANDARD
   for will loss purposes (#44)
 - Added option to mitigate all weapon damage using armor instead of always taking at
   least 1 damage (#321)
+- Added option to reduce incoming damage using the target's armor before the
+  target's shields/ablative HP are damaged (#743)
```
Xymanek: CHANGELOG.md is depreciated, don't add entries to it anymore

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -6065,37 +6065,97 @@ event TakeDamage( XComGameState NewGameState, const int DamageAmount, const int
...
+	/// ```
+	if ((ShieldHP > 0) && !bIgnoreShields) // If there is a shield
+	{
+		if (class'X2Effect_ApplyWeaponDamage'.default.ARMOR_BEFORE_SHIELD) // Armor should take damage before shield, then spill to HP
+		{
+			`LOG("Beginning Armor-Shield-Health Processing!");
+			`LOG("Incoming Damage: " $ (DamageAmount + MitigationAmount));
+			`LOG("Armor Mitigation: " $ MitigationAmount);
+			`LOG("Armor Shredded: " $ ShredAmount);
+			`LOG("Leaking Damage: " $ DamageAmount);
```
Xymanek: I'm not a huge fan of this `ScriptLog` spam, it will spew for anyone not running a final release CHL. Please use `COMBATLOG` like the rest of that function

Iridar:
> Tested with shredding and without shredding. 

What does this mean? Does the armor get Shredded even if you still have Shield HP remaining?

cannonfodder1:
> Does the armor get Shredded even if you still have Shield HP remaining?

Yup. If armor becomes the outer layer of defense and ablative becomes the inner layer, then it only makes sense to allow shredding to take effect before the ablative layer is broken.

Iridar:
> > Does the armor get Shredded even if you still have Shield HP remaining?
> 
> Yup. If armor becomes the outer layer of defense and ablative becomes the inner layer, then it only makes sense to allow shredding to take effect before the ablative layer is broken.

Agreed. I have an ability that behaves in a similar way, guess I'll have to add a workaround for your config option when it goes live.

Issue #746, opened 2020-01-23
Make "Legacy Operations" respect Encounter.ini changes


Deadput:
This would be so the preplaced enemy encounters like say the first pod of Faceless on Blast from the Past could have their places swapped out for say a new custom unit or something like that. 

So as an exaggerated example. 

-ConfigurableEncounters=(EncounterID="GasStationDiner_GasPumps_Easy", \\
	MaxSpawnCount=2, \\
	ForceSpawnTemplateNames[0]="Faceless", \\
	ForceSpawnTemplateNames[1]="Faceless", \\
	bGroupDoesNotAwardLoot=true)

to

+ConfigurableEncounters=(EncounterID="GasStationDiner_GasPumps_Easy", \\
	MaxSpawnCount=2, \\
	ForceSpawnTemplateNames[0]="SuperFacelessDude", \\
	ForceSpawnTemplateNames[1]="NotSuperFacelessBoy", \\
	bGroupDoesNotAwardLoot=true)

MrNiceUK:
I have a few ideas in my head for implementing this...

MrNiceUK:
Finally noticed that `PopulateFromNarrativeConfig()` is never called in the normal process, so some changes to `NarrativeLadderConfig` config entries are not picked up, since they aren't directly used but rather copied to `XCGS_LadderData` properties. Noticeably this affects the upgrade choices available at the end of each mission.

PR #748, opened 2020-01-26
Issue #746 Implementation (XComEncounter.ini changes taking effect in Legacy Ops)

MrNiceUK:
Implements #746. Changes to the encounter groups are now accommodated, with some limitations. It only works with `ForceSpawnTemplateNames`, if `EncounterLeaderSpawnList` or `EncounterFollowerSpawnList` are not blank, then no changes are made for that encounter at all. It changes the Encounter to match the `ForceSpawnTemplateNames` list, up to the specified `MaxSpawnCount` limit. The new list can be shorter *or* longer than for the original encounter.
Also brought enhancements for #307. The #308 PR for that stopped short of allowing changes to the squad size. That is now possible, along with allowing the CharacterTemplate to be different, so switching between normal soldiers and faction soldiers or SPARKs is now supported, for example. Or switching in out Central or Shen, who have custom character templates.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_LadderProgress.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_LadderProgress.uc
@@ -339,6 +339,9 @@ static function ProceedToNextRung( )
 			NextMissionLadder = XComGameState_LadderProgress( History.GetSingleGameStateObjectForClass( class'XComGameState_LadderProgress' ) );
 			NextMissionLadder.CumulativeScore = LadderData.CumulativeScore;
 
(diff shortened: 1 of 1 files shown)


MrNiceUK:
New style inline documentation to come, obviously.

MrNiceUK:
Some minor signs that this PR *may* be messing up which unit is the "Pod Leader" in some cases, but more investigation needed to be sure on that.

Iridar:
@MrNiceUK this PR has been soaking almost a year, would you consider picking this up? Would be a shame for your work to go to waste, though I wouldn't blame anyone for not having interest in Legacy Ops.

Iridar:
Archiving this due to lack of progress over three years.

Issue #749, opened 2020-01-30
Change #400 #401 to Trigger on All Abilities


VampireMonkey:
Will allow mods to change the color of any Ability Icon for any arbitrary reason.

A pull request will come soon

VampireMonkey:
Added pull request #750

PR #750, opened 2020-01-30
Changed how #400 worked to trigger an event for all abilities

VampireMonkey:
Pull request for Issue #749


Xymanek:
@pledbrook You had concerns about increasing the scope of the existing event on discord (especially since the event is now part of the stable CHL), could you please express them here?

MrNiceUK:
I'll only comment that the *original* PR should have been crafted more wildly in the first place, not tailored so close to LWotC's requirements. Actually you as good as said that yourself back in April...
https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/401#issuecomment-486096471

MrNiceUK:
> listeners should be called before or after the default colours are picked. The latter would allow for mods to perform colour substitutions. Not sure whether that's a good enough reason for the added (slight) complexity.
I would say do the latter. Not sure where the (slight) complexity comes in? Code wise would seem slightly more compex for the event trigger to have to check *if* any listeners set a color, then decide if the normal color picking process has to be run, rather than just trigger after and just use whatever color is in the tuple afterwards, without caring if any listeners even existed or not?

VampireMonkey:
I'd argue against adding another event trigger and that having a bool to check for override might be better for performance than comparing strings, as this event will be triggered for each icon in the ability hud, and it will be triggered everytime the hud is refreshed, which happens on ex. soldier rotations.

However since we're changing colors only, we might consider not having anything in UITacticalHUD_Ability and instead use UITacticalHUD_AbilityContainer.InitAbilityContainer
https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/09e98ed14b4a057bf522ba21ddaa9c122a301cef/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_AbilityContainer.uc#L86-L102
This _might_ only run once and could be a better overall solution to all problems
Further testing and research would have to be done to be sure.

But the event in question could just return the whole UITacticalHUD_AbilityContainer object allowing any direct modifications of its children.

MrNiceUK:
But I *think* the Ability Container is just created once per load/tactical start, so colouring couldn't be dynamic (charges left, just for an example)?

MrNiceUK:
One thing I might agree with PledBrook on is how possible is it to get new functionality out of the existing event *without* breaking the BC contract?

VampireMonkey:
https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/09e98ed14b4a057bf522ba21ddaa9c122a301cef/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_AbilityContainer.uc#L1289-L1314
You could possibly do it after this loop, but if that would force flash to update or not, i don't know.
I'm afraid that you could end up with having the color update and then having the override update afterwards on screen, resulting in flickering/flashing.

If we're considering only having 1 event trigger that is.

MrNiceUK:
> having a bool to check for override might be better for performance than comparing strings
Why would you compare strings? Essentially, get rid of all the cases in the switch block individually calling `EnableMouseAutomaticColor()`, and have them set the foreground/background color variables instead (including anything done by the original event in the relevant case). After the switch block, trigger the new event, passing those colors in the tuple, then call `EnableMouseAutomaticColor()` using whatever colours are now in the tuple. No need to check if anything has changed or not, and the ony extra performance cost is storing the two colours and the event trigger itself, which is fairly cheap if there are nolisteners, and if there *are* listeners, the calling the delegate is the performance hit, which you incur whether there is one trigger or two.

VampireMonkey:
The string comparison is how the event worked before, i tried not to be too invasive with the changes but if we're going all out then by all means.
You'd have to set DefaultForegroundColor & DefaultBGColor in UIIcon and then EnableMouseAutomaticColor() when you're done to insure flash will update them.
I still think raising the event inside UITacticalHUD_AbilityContainer.PopulateFlash() might be the better overall solutions and would only incur a single event

MrNiceUK:
The issue of if you can squeeze the functionality within the existing event (required if you want to only incur a single event) without breaking BC is a separate issue really about where/when to fire the event. But as said, firing an event is cheap, it's the *listeners* which cost. And there would be the same amount of those with either approach.

VampireMonkey:
All should be good now

X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Ability.uc
```diff
@@ -179,36 +182,27 @@ simulated function UpdateData(int NewIndex, const out AvailableAction AvailableA
...
-// The event itself takes the form:
-//
-//   {
-//      ID: OverrideAbilityIconColor,
-//      Data: [in bool IsObjective, out string IconColor],
-//      Source: Ability
-//   }
-//
-static function TriggerOverrideAbilityIconColor(XComGameState_Ability Ability, bool IsObjective, out string IconColor)
+// Start Issue #400, #749
```
pledbrook: This needs highlander docs, as described [here](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/README.md#documenting-your-contributions).

Example doc [here](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc#L323). You should include the event source alongside event ID and event data in the docs. Just ask in #highlander if you have any questions.

pledbrook:
Oh, also please fix the conflict. I think CHANGELOG.md may have undergone a bunch of changes. Not sure that you need to document changes in it any more.

VampireMonkey:
> Oh, also please fix the conflict. I think CHANGELOG.md may have undergone a bunch of changes. Not sure that you need to document changes in it any more.
What conflict?

X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Ability.uc
```diff
@@ -182,7 +182,21 @@ simulated function UpdateData(int NewIndex, const out AvailableAction AvailableA
 	RefreshShine();
 }
 
-// Start Issue #400, #749
+/// HL-Docs: feature:OverrideAbilityIconColor; issue:400,749; tags:tactical,compatibility
+/// This event triggers each time an ability icon is updated in the UITacticalHUD, allowing mods to override the colors of each icon
+///	from a tuple that also contains all the date about the icon and its related ability.
```
pledbrook: Typo on "date"

X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Ability.uc
```diff
@@ -182,7 +182,21 @@ simulated function UpdateData(int NewIndex, const out AvailableAction AvailableA
...
+///
+/// ```unrealscript
+/// EventID: OverrideAbilityIconColor
+/// EventData: XComLWTuple {
+///     Data: [
+///       inout bool IsObjective,
```
pledbrook: This is technically an `in` parameter. The mod can't override that aspect.

pledbrook:
There was a merge conflict, but you fixed it with the merge commit.

I would really like to retain the old event and give the new one a new name. You can use `TriggerOverrideAbilityIconColor()` for the old and new event because the tuple is backwards compatible. So you just need to add the event name as a function argument of `TriggerOverrideAbilityIconColor()`. And then call `TriggerOverrideAbilityIconColor('OverrideAbilityIconColor', ...)` from the `if` and `else if` branches. Not sure what to call the new event. Maybe just 'OverrideAbilityFgBgIconColor' or 'OverrideAbilityIconColorNew'.

VampireMonkey:
> There was a merge conflict, but you fixed it with the merge commit.
> 
> I would really like to retain the old event and give the new one a new name. You can use `TriggerOverrideAbilityIconColor()` for the old and new event because the tuple is backwards compatible. So you just need to add the event name as a function argument of `TriggerOverrideAbilityIconColor()`. And then call `TriggerOverrideAbilityIconColor('OverrideAbilityIconColor', ...)` from the `if` and `else if` branches. Not sure what to call the new event. Maybe just 'OverrideAbilityFgBgIconColor' or 'OverrideAbilityIconColorNew'.
Seems overkill as they don't conflict with each other and shouldn't break back-compatibility, IconColor was afterall just the background color that is still in the same position in the tuple.

pledbrook:
> shouldn't break back-compatibility

It does break backwards compatibility simply by triggering the listener for more abilities than it did before. I'll try to create a change suggestion.

CHANGELOG.md
```diff
@@ -230,9 +230,7 @@ RunPriorityGroup=RUN_STANDARD
   want to handle it themselves (#507)
 - 'DrawDebugLabels' allows mods to draw their own debug information on the canvas used by
   `XComTacticalController.DrawDebugLabels()` (#490)
-- `OverrideAbilityIconColor` provides a tuple with the same ID as the
-  event and data of the form `[bool IsObjective, string Color]` that allows
-  mods to override the color of soldier abilities in the tactical HUD (#400)
+- `OverrideAbilityIconColor` Provides a tuple that allows mods to override the color of soldier abilities in the tactical HUD (#400, #749)
```
Musashi1584:
```suggestion
```
Please remove deprecated CHANGELOG.md documentation

X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Ability.uc
```diff
@@ -128,44 +129,53 @@ simulated function UpdateData(int NewIndex, const out AvailableAction AvailableA
...
+
+		if (IsObjectiveAbility)
 		{
-			IconColor = IsObjectiveAbility ? class'UIUtilities_Colors'.const.OBJECTIVEICON_HTML_COLOR : AbilityTemplate.AbilityIconColor;
-			TriggerOverrideAbilityIconColor(AbilityState, IsObjectiveAbility, IconColor);
-			Icon.EnableMouseAutomaticColor(IconColor, class'UIUtilities_Colors'.const.BLACK_HTML_COLOR);
+			BackgroundColor = class'UIUtilities_Colors'.const.OBJECTIVEICON_HTML_COLOR;
+
+			// DEPRECATED - Use the new OverrideBothAbilityIconColors event instead
+			TriggerOverrideAbilityIconColor(AbilityState, 'OverrideAbilityIconColor', IsObjectiveAbility, BackgroundColor, ForegroundColor);
```
Musashi1584: ForegroundColor is not initialised here

X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Ability.uc
```diff
@@ -128,44 +129,53 @@ simulated function UpdateData(int NewIndex, const out AvailableAction AvailableA
...
+		}
+		else if (AbilityTemplate.AbilityIconColor != "")
+		{
+			BackgroundColor = AbilityTemplate.AbilityIconColor;
+
+			// DEPRECATED - Use the new OverrideBothAbilityIconColors event instead
+			TriggerOverrideAbilityIconColor(AbilityState, 'OverrideAbilityIconColor', IsObjectiveAbility, BackgroundColor, ForegroundColor);
```
Musashi1584: ForegroundColor is not initialised here

X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Ability.uc
```diff
@@ -128,44 +129,53 @@ simulated function UpdateData(int NewIndex, const out AvailableAction AvailableA
...
 		}
 		else
 		{
-			switch(AbilityTemplate.AbilitySourceName)
+			switch (AbilityTemplate.AbilitySourceName)
```
Musashi1584: Dont change formatting cause it obfusates the real changes and makes diffs hard to read in the future.
```suggestion
			switch(AbilityTemplate.AbilitySourceName)
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Ability.uc
```diff
@@ -128,44 +129,53 @@ simulated function UpdateData(int NewIndex, const out AvailableAction AvailableA
...
+					BackgroundColor = class'UIUtilities_Colors'.const.BAD_HTML_COLOR;
+					break;
+				case 'eAbilitySource_Psionic':
+					BackgroundColor = class'UIUtilities_Colors'.const.PSIONIC_HTML_COLOR;
+					break;
+				case 'eAbilitySource_Commander': 
+					BackgroundColor = class'UIUtilities_Colors'.const.GOOD_HTML_COLOR;
+					break;		
+				case 'eAbilitySource_Item':
+				case 'eAbilitySource_Standard':
```
Musashi1584: Dont change formatting cause it obfusates the real changes and makes diffs hard to read in the future.
```suggestion
			case 'eAbilitySource_Perk':
 				BackgroundColor = class'UIUtilities_Colors'.const.PERK_HTML_COLOR;
 				break;

 			case 'eAbilitySource_Debuff':
 				BackgroundColor = class'UIUtilities_Colors'.const.BAD_HTML_COLOR;
 				break;

 			case 'eAbilitySource_Psionic':
 				BackgroundColor = class'UIUtilities_Colors'.const.PSIONIC_HTML_COLOR;
 				break;

 			case 'eAbilitySource_Commander': 
 				BackgroundColor = class'UIUtilities_Colors'.const.GOOD_HTML_COLOR;
 				break;

 			case 'eAbilitySource_Item':
 			case 'eAbilitySource_Standard':
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Ability.uc
```diff
@@ -179,36 +189,61 @@ simulated function UpdateData(int NewIndex, const out AvailableAction AvailableA
...
+/// template. This event allows mods to override the background color of each such icon.
+///
+/// ```unrealscript
+/// EventID: OverrideAbilityIconColor
+/// EventData: XComLWTuple {
+///     Data: [
+///       in bool IsObjective,
+///       inout string BackgroundColor
+///     ]
+/// }
```
Musashi1584:
```suggestion
/// GameState: no
/// }
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Ability.uc
pledbrook: It actually doesn't matter in this case because the listeners will ignore that value anyway. The old event only passed the background colour.

pledbrook: It probably makes sense to move the `ForegroundColor` initialisation to before this `if` condition.

pledbrook:
Is there anything that I can do to help move this forward?

Iridar:
This is needlessly complicated, I'm gonna drop this entirely and re-implement cleanly, with two separate event triggers.

PR #755, opened 2020-02-04
Prevent psi zombies from dropping loot (#682)

### LINKED ISSUE #682: Prevent Psi Zombies from dropping loot
If a unit is killed by an explosion (which stops loot from being dropped) and is then raised as a zombie, killing that zombie drops the original loot.

Issue reported in [LWOTC](https://github.com/long-war-2/lwotc/issues/180).

pledbrook:
If loot is supposed to be destroyed when a unit is killed, for example by explosives in LW2, and those units are raised as psi zombies, those zombies will drop the loot when *they* die.

This change clears the pending loot from units when they die, ensuring that they can't drop the loot if they're raised from the dead.

Fixes #682.


Iridar:
Will test in-game.

pledbrook:
@Iridar Have you tested it yet?

Iridar:
> @Iridar Have you tested it yet? Note that it has been active in LWOTC for quite a while and is basically a direct port of the LW2 change. That's not to say it doesn't need testing (particularly in vanilla), but more to point out that there has been some.

That's good to hear. Sorry, I kinda forgot to test this.

Iridar:
I have reproduced the bug without this fix, and confirmed this bug is fixed with this PR. I'm not particularly versed in loot code, so I can't vouch for the logic, but from what I understand, we just destroy the loot in all the same circumstances we display the "LOOT DESTROYED" message, which seems like it should be done anyway. So this PR has my blessing.

Iridar (approved):

PR #756, opened 2020-02-05
 Fire event on expiry of a dark event (#753)

### LINKED ISSUE #753: Fire an event when a dark event expires (is deactivated)

pledbrook:
The `XCGS_DarkEvent.OnDeactivate()` function now fires a 'DarkEventDeactivated' event before deactivating the DE.

The event takes the form:
```
  ID: DarkEventDeactivated
  Data: self (XCGS_DarkEvent)
  Source: self (XCGS_DarkEvent)
  NewGameState: yes
```

Resolves #753.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_DarkEvent.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_DarkEvent.uc
@@ -1,0 +1,477 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    XComGameState_DarkEvent.uc
+//  AUTHOR:  Mark Nauta
+//         
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+class XComGameState_DarkEvent extends XComGameState_BaseObject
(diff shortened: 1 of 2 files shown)


Xymanek:
I think I've said this before, I dislike events starting with `On[...]`/just stating the fact (this case is the latter) as it's unclear whether the vanilla behaviour has or has not happened yet, and this may be important for some consumers. A much better naming would be `Pre[...]`/`Post[...]`

pledbrook:
Whether the vanilla behaviour has happened or not depends on what type of listener you set up, as described in the documentation for the event. So calling it `PreDarkEventDeactivated` would be inaccurate if you set up an `ELD_OnStateSubmitted` listener.

Xymanek:
For `ELD_OnStateSubmitted` yes, but it does make a difference for `ELD_Immediate`

pledbrook:
Event renamed.

PR #742, opened 2020-01-22
Move index etc. out of py script, Add tags pages

robojumper:
standard page

tags page

Closes #754.

PR #760, opened 2020-02-07
Trigger docs build (but not deploy) on PR pushes as well

Xymanek:
**This is a concept**. @robojumper if this looks ok, I will proceed with testing this

Splits the `docs` workflow into 2 jobs - `build` and `deploy` (with latter requiring former). This way, we can run the `build` on PRs as well, so that we can check that the newly introduced code doesn't break the docs script. Deployment is still done only for `master` pushes.

A bonus point of this change is that you can download the built docs and open them locally (if you don't want to/can't build them yourself). Example screenshot from GitHub's docs:

image

These are stored for 90 days and then deleted

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,12 +1,9 @@
 name: docs
 
-on:
-  push:	
-    branches:	
-      - master
+on: [push, pull_request]
 
 jobs:
-  build-and-deploy:
+  build:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/setup-python@v1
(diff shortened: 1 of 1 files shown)


Xymanek:
Currently something's wrong with "is master push" check, doesn't seem to pass - https://github.com/Xymanek/X2WOTCCommunityHighlander/actions/runs/35858812

Xymanek:
Fixed, see https://github.com/Xymanek/X2WOTCCommunityHighlander/actions/runs/35865595 (obv there is no token)

Xymanek:
@robojumper could we merge this, so that it takes effect on upcoming PRs (especially #742)?

PR #762, opened 2020-02-07
Implemented scripted versions for ability target styles

Xymanek:
Based on discord suggestion by @robojumper, only affects abilities that use a scripted style(s)

Demo: https://github.com/Xymanek/X2WOTCCommunityHighlander/commit/36c5b96a626e6474e45dbac59aba100e53eaa570

```
[0107.18] CreateAbilityStateFromTemplate: HoloTargeting state created default
[0107.18] CreateAbilityStateFromTemplate: ChainShot state created default
[0107.18] CreateAbilityStateFromTemplate: HailOfBullets state created custom
[0107.18] CreateAbilityStateFromTemplate: BulletShred state created default
```

```
[0115.41] Log: X2TacticalGameRuleset_0 TurnPhase_UnitActions: Check for Available Actions
[0115.41] GatherAbilityTargets: Entered script version for HailOfBullets
[0115.41] X2AbilityTarget_Scripted: GetPrimaryTargetOptions
[0115.41] X2AbilityTarget_Scripted: CheckFilteredPrimaryTargets
[0115.41] ScriptLog: XComTacticalInput.uc, Transition to/from State None detected. BeginState ActiveUnit_Moving from None
```

cc @Iridar (I did this to prove you wrong :trollface: ), @pledbrook (I know LW2 needed this)

PR #764, opened 2020-02-07
Add `Object:IsClassNative` to check whether a class is native

### LINKED ISSUE #767: `Object:IsClassNative`
To check whether a given class is native.

robojumper:
In which we work our way into the innards of the UnrealScript virtual machine. Closes #767.

Original description below, see comments for current design: 

<details><summary>Click to expand</summary>
<p>

```unrealscript
// Must be called on a `Class` instance, i.e. if you have an X2AbilityTargetStyle and
// want to know whether it's native, check `TargetStyle.class.IsNative()`
final function bool IsNative()
{
	// This is the flag the UC VM uses to mark a class as native
	return (self.ObjectFlags.B & (1 << 26)) != 0;
}
```

You may call `<SomeObject>.class.IsNative()` to determine whether the class is native, or, if you are working with a `class<_>` already, skip the `.class` part.

</p>
</details>

Because I am a skeptical person and you should be too, here is a debug output that makes this implementation appear correct (interestingly enough, `GetClassDefaultObjects` yields objects for native classes first):

[https://pastebin.com/acG9Jc0a](https://pastebin.com/acG9Jc0a)

This may come in useful for #762.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_Components.uc
+++ b/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_Components.uc
@@ -3,11 +3,9 @@ class X2WOTCCH_Components extends Object config(CHLComponents);
 enum CHLComponentStatus
 {
 	// OK
-	eCHLCS_NotExpectedNotFound,
 	eCHLCS_OK,
 	// Warnings
 	eCHLCS_VersionMismatch,
-	eCHLCS_ExpectedNotFound,
 	// Errors
 	eCHLCS_RequiredNotFound,
 };
@@ -38,8 +36,6 @@ struct CHLComponentVersion
 
 var localized string VersionFormat;
 var localized string RequiredNotFound;
-var localized string ExpectedNotFound;
-var localized string NotExpectedNotFound;
 
 var localized string VersionMismatches;
 var localized string WarningsLabel;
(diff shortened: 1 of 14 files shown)


Xymanek:
Why not check `if (class(self) != none)` (or vice versa) and adjust the check automatically, so that you could call the method directly, without caring about "type of instance" that you are dealing with?

robojumper:
I'm conflicted. On one hand, it's more "magic" and can clash with the mental model where objects and their classes are entirely separate things (see https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/477#discussion_r275116456, where I got confused about this) and introduces an inconsistency with other functions in `Object` (see discussion I just linked).

On the other hand, the only kinds of `Object` that can have this flag set (and for which the function makes sense) are classes, but I can't add it to `Class.uc` because that doesn't exist.

Perhaps a
```unrealscript
final static function bool ClassIsNative(Class klass)
{
	// This is the flag the UC VM uses to mark a class as native
	return (klass.ObjectFlags.B & (1 << 26)) != 0;
}
```
would be better?

Xymanek:
Yeah, `ClassIsNative` resolves the confusion 👍

MrNiceUK:
Assume that's "as well as", so you call IsNative() or IsClassNative() depending what you're dealing with?

robojumper:
No, a static function `ClassIsNative`.

MrNiceUK:
Hmm, but if ask you *have* is a class reference, that means you have to instance it just to check? (or fetch the CDO)

Xymanek:
What about #765?

robojumper:
Fixed. I hope.

X2WOTCCommunityHighlander/Src/XComGame/Classes/CHXComGameVersion.uc
```diff
@@ -17,34 +16,10 @@ static function array<X2DataTemplate> CreateTemplates()
...
-	{
-		class'X2TacticalGameRuleset'.static.ReleaseScriptLog("X2WOTCCommunityHighlander: Creating Engine version template...");
-		`CREATE_X2TEMPLATE(class'CHXComGameVersionTemplate', XComGameVersion, 'CHEngineVersion');
-		XComGameVersion.MajorVersion = class'Engine.CHEngineVersion'.default.MajorVersion;
-		XComGameVersion.MinorVersion = class'Engine.CHEngineVersion'.default.MinorVersion;
-		XComGameVersion.PatchVersion = class'Engine.CHEngineVersion'.default.PatchVersion;
-		XComGameVersion.Commit = class'Engine.CHEngineVersion'.default.Commit;
-		class'X2TacticalGameRuleset'.static.ReleaseScriptLog("X2WOTCCommunityHighlander: Created Engine version template with version" @ XComGameVersion.MajorVersion $ "." $ XComGameVersion.MinorVersion $ "." $ XComGameVersion.PatchVersion);
-		Templates.AddItem(XComGameVersion);
-	}
```
Xymanek: Isn't this backwards incompatible? Since someone might be checking the engine version already

X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_Components.uc
```diff
@@ -3,11 +3,9 @@ class X2WOTCCH_Components extends Object config(CHLComponents);
...
 {
 	// OK
-	eCHLCS_NotExpectedNotFound,
 	eCHLCS_OK,
 	// Warnings
 	eCHLCS_VersionMismatch,
-	eCHLCS_ExpectedNotFound,
 	// Errors
 	eCHLCS_RequiredNotFound,
 };
```
Xymanek: Shouldn't we also remove the `NotExpectedNotFound` and `ExpectedNotFound` below?

Xymanek:
Also, the matter of name collision still remains. I think naming the function `CH_IsClassNative` would be much safer option

-----------

Also 2: Shouldn't `/// HL-Docs: ref:ComponentVersions` be added to the DLC2 version?

X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_Components.uc
```diff
@@ -49,86 +47,27 @@ delegate bool VersionInfoFetcher (out CHLComponentVersion VersionInfo);
...
-	Manager = class'X2StrategyElementTemplateManager'.static.GetStrategyElementTemplateManager();
 	DLCNames = class'Helpers'.static.GetInstalledDLCNames();
 
-	SelfVersion = Manager.FindStrategyElementTemplate('CHWOTCVersion');
-	Comps.AddItem(BuildComponent(SelfVersion, none, "X2WOTCCommunityHighlander", true, true));
-	Comps.AddItem(BuildComponent(Manager.FindStrategyElementTemplate('CHCoreVersion'), SelfVersion, "Core", true, true));
-	Comps.AddItem(BuildComponent(Manager.FindStrategyElementTemplate('CHEngineVersion'), SelfVersion, "Engine", true, true));
-	Comps.AddItem(BuildComponent(Manager.FindStrategyElementTemplate('CHXComGameVersion'), SelfVersion, "XComGame", true, true));
+	CompanionVersionInfoFetcher(SelfVersion);
+	Comps.AddItem(BuildComponent(CompanionVersionInfoFetcher, "X2WOTCCommunityHighlander", SelfVersion));
```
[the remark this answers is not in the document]
robojumper: Because the version struct cannot be `null`. Comparing it against itself in `BuildComponent` always succeeds though.

X2WOTCCommunityHighlander/Src/XComGame/Classes/CHXComGameVersion.uc
robojumper: Technically yes, but the only `Engine` feature we have is the `AudioComponent` WWise/SoundCue fix which I don't think anyone ever queried for.

X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_Components.uc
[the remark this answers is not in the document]
Xymanek: It can be simply omitted though (due to `optional`) and the `if (CompareVersion.MajorVersion > -1)` check will fail

X2WOTCCommunityHighlander/Src/XComGame/Classes/CHXComGameVersion.uc
Xymanek: > I don't think anyone ever 

So far this kind of thought was allowed only in extreme cases, particularly related to new features

X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_Components.uc
```diff
@@ -233,27 +142,22 @@ static function string FormatVersionNew (CHLComponentVersion Version)
...
+		return true; \n\
+	}
 
-	return true;
-}
+`MakeVersionFetcher(DLC2VersionInfoFetcher, CHVersion_DLC2)
+`MakeVersionFetcher(CoreVersionInfoFetcher, CHCoreVersion)
+`MakeVersionFetcher(EngineVersionInfoFetcher, CHEngineVersion)
+`MakeVersionFetcher(XComVersionInfoFetcher, CHXComGameVersionTemplate)
+`MakeVersionFetcher(CompanionVersionInfoFetcher, CHVersion_X2WOTCCH)
```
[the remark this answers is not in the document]
Xymanek: As pointed out by @robojumper on discord, we already have `CHEngineVersion` and `CHXComGameVersionTemplate` which cannot be renamed due to BC, so the only way to try to maintain consistency is via `CH[module]Version` names.

Good thing is that we can still rename `CHVersion_DLC2` and `CHVersion_X2WOTCCH`

robojumper:
Addressed. I would still like to get rid of the Engine template because it's not useful and there is no mod that ever checked that Engine version (it only contained a single bugfix).

X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_Components.uc
```diff
@@ -49,85 +45,27 @@ delegate bool VersionInfoFetcher (out CHLComponentVersion VersionInfo);
 
 static function array<CHLComponent> GetComponentInfo()
 {
-	local X2StrategyElementTemplateManager Manager;
-	local X2StrategyElementTemplate SelfVersion;
-	local CHLComponentVersion SelfVersionNew;
+	local CHLComponentVersion SelfVersion, EmptyVersion;
```
Xymanek:
```suggestion
	local CHLComponentVersion SelfVersion;
```
I don't see where it's used

Xymanek:
I mean, if you are so sure that no one ever checked the engine template, then I guess go ahead....

PR #776, opened 2020-02-16
Prevent crashes from applying facility upgrades if the facility map isn't loaded

### LINKED ISSUE #775: Activating facility upgrades before facility maps are loaded crashes the game
Use case: making "debug strategy start" start with faction(s) met

Culprit: `class'Helpers'.static.SetOnAllActorsInLevelWithTag(none, //...`, called from `XGBase::UpdateFacilityUpgradeVisibilty`

Xymanek:
Fixes #775 

Better compare: https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/compare/e5b0602...d2e6b8f

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGBase.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGBase.uc
@@ -1,0 +1,1051 @@
+//-----------------------------------------------------------
+//
+//-----------------------------------------------------------
+class XGBase extends Actor config(GameData);
+
+var config string BaseMapName;
+var config string BaseCapMapName;
+
+var LevelStreaming AvengerCap_Level;
+var LevelStreaming SkyAndLight_Level;
+var LevelStreaming Avenger_Level;
+var LevelStreaming Terrain_Level;
+var LevelStreaming PostMissionSequence_Level;
+var LevelStreaming PreMissionSequence_Level;
+
+var MaterialInstanceConstant FacilityPreviewMaterial;
+var float FacilityPreviewTransition;
+
+var string CurrentLightEnvironment;
+var string CurrentTerrainMap;
+
+struct AuxLevels
+{
+	var array<LevelStreaming> Levels;
+};
+
+struct PreviewRequest
+{
+	var int MapIndex;
+	var name TemplateName;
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XGBase.uc
```diff
@@ -259,6 +259,11 @@ private function UpdateFacilityUpgradeVisibilty(int RoomIdx)
 	local string PreUpgradeName;
 	local XComGameState_HeadquartersRoom RoomState;
 
+	// Issue #775 (single line)
+	/// HL-Docs: ref:Bugfixes; issue:775
+	/// Prevent crashes from applying facility upgrades if the facility map isn't loaded
```
pledbrook, line 264: Just a little thing: this sounds as if the *crashes* can apply facility upgrades. I understand what is meant, but it's awkwardly put (and this is part of the public documentation). Try:
> Applying/activating/enabling (choose which you prefer!) facility upgrades before the facility map is loaded no longer crashes the game.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGBase.uc
```diff
@@ -259,6 +259,11 @@ private function UpdateFacilityUpgradeVisibilty(int RoomIdx)
...
+	if (m_arrLvlStreaming[RoomIdx] == none) return;
```
pledbrook, line 265: I have no idea what the implications of this are, so don't feel qualified to say yes or no to it. What happens if a player tries to purchase the facility upgrade later? Or is it basically prepurchased?

Xymanek: There is a very small window during strategy init where state is being processed but the visual aspect wasn't loaded yet. For changes to facilities in that window, the same logic as loading from a save will apply - the visual will init to the new state

robojumper (approved):

pledbrook:
Merged with [this commit](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/commit/075ba86aa13284d00ba4f1e81e8f11a39df7b740) (I squashed the second and third commits).

PR #778, opened 2020-02-19
Update risk level on risk chance recalculation (#777)

### LINKED ISSUE #777: Update CA risk level when risk modifiers are recalculated
The [`XCGS_CovertAction.RecalculateRiskChanceToOccurModifiers()`](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc#L973) function doesn't update the level of the risk at the same time as recalculating the risk chance. This means that the risks panel won't be updated with the new information.

pledbrook:
This change simply ensures that the risk level (low, moderate, high) is updated whenever the associated risk's chance to occur is changed via `XCGS_CovertAction.RecalculateRiskChanceToOccurModifiers()`.

Fixes #777.


robojumper:
This isn't a base-game bug, is it? We don't really document bugfixes for HL bugs, so the docs wouldn't be needed.

Also the same function uses spaces instead of tabs a few lines earlier, it'd be great if you could fix that while you're at it.

pledbrook:
I think I've made the changes you wanted @robojumper, but let me know if I misinterpreted your comment.

robojumper (approved):

PR #780, opened 2020-02-19
Fire event to allow overriding CA risk chances (#779)

### LINKED ISSUE #779: Allow mods to override the display text for covert action risks
There is no easy way for mods to override how covert actions risks are displayed (currently it's as "Low", "Moderate" or "High"). We should add an event that allows them to override both the risk labels and values.

For example, a mod may want to show actual percentages for the risk chance rather than a severity band.

pledbrook:
This change introduces the following event that allows mods to change how covert action risk chances are displayed:
```
{
  ID: CovertAction_OverrideRiskStrings
  Data: [
       inout array<string> RiskLabels,
       inout array<string> RiskChanceTexts ]
  Source: self (XCGS_CovertAction)
}
```

Resolves #779.


pledbrook:
Although it *is* possible to increase the number of risk levels and solve the issue that way, at least one bit of code in the base game assumes there are three risk levels. And it's not beyond the realms of possibility that other mods will also assume there are only three risk levels configured in a covert action, which leaves the gate wide open for weird incompatibilities between mods that need this change and others.

Issue #297, opened 2018-10-12
Investigate registering listeners on the shell menu


robojumper:
Currently, X2EventListeners are registered in tactical, and in strategy. The fact that they don't work on the shell screen was a drawback we accepted whenever we decided to use an event hook instead of a DLC hook.

TLE moves some UI to the shell that may be relevant for event hooks (dynamic class icons, for example). We should investigate whether registering on the shell is wanted.

GingerAvalanche:
What are the downsides to using a new boolean RegisterInShell for the CHEventListenerTemplate (or even for X2EventListenerTemplate) and then registering event listeners with that boolean set to true, in OnInit of UIShell or UIFinalShell, and then removing them prior to unloading the shell level?

robojumper:
Update: This seems somewhat difficult, as there are lots of places on the shell menu where the history is reset, clearing all registered listeners. Probably better to wait for a real use case or problem before attempting this...

Iridar:
Well the Issue #783 would be a real use case, I even have the Event done and set up. I'm perfectly content with changing it to a DLC hook, though. I'm not sure if there's something that an Event would accomplish, but a DLC hook would not, in this specific case, since there's all event listeners would have to use ELD_Immediate to take advantage of it anyway.

Xymanek:
Somewhat related: personally, I think we should register strategy listeners before calling `CreateStrategyGameStart` - the setup process may be invoking CHL extension events which are currently getting "lost"

Musashi1584:
I have a new "real use case" with events for the mcm actions in my mcmbuilder that should work in the shell too. @robojumper is this something you would consider to pick up again?

robojumper:
Conceptually, the shell is much closer to strategy than tactical. Are there any drawbacks to just registering strategy listeners on the shell?

Xymanek:
Strategy listeners can expect things like `HQPRES` to exist, which would no longer be the case

Musashi1584:
Just tested calling `class'X2EventListenerTemplateManager'.static.RegisterTacticalListeners();` in a UISL on the shell enables strategy listeners to receive events.

Musashi1584:
> Strategy listeners can expect things like `HQPRES` to exist, which would no longer be the case

wouldn't none checking that be in the error handling responsibility of the listener code?

Xymanek:
Why would you check for existence of global singletons?

Musashi1584:
[...]

PR #784, opened 2020-02-24
New DLC hook ModifyGeneratedUnitAppearance

### LINKED ISSUE #783: Meaningful changes to Character Generation require MCO / OPTC override
Mods like [Clean Soldier Generation](https://steamcommunity.com/sharedfiles/filedetails/?id=1707030670) and [Gender Bias - WOTC](https://steamcommunity.com/sharedfiles/filedetails/?id=1163324651) have to rely on a Mod Class Override in order to make changes to the appearance of generated soldiers.

I propose to add an Event to `XGCharacterGenerator.uc` that will trigger after the soldier appearance is generated so that mods can make arbitrary changes to it at that point.

To make this event fully functional when it comes to resistance faction soldiers will require refactoring them a bit, more details in the upcoming PR for this issue.

Iridar:
Fixes #783

Initially I wanted to address this with an Event, but ran into the issue mentioned in #297; some soldiers (new rookies added to the HQ roster on the campaign start) had their appearance generated before the Event Listener had a chance to register.

So I added a DLC hook instead, which is functionally the same, because the Event Listeners would have to use ELD_Immediate in order to make any changes to the generated appearance anyway.

The hook code runs at the very end of `XGCharacterGenerator::CreateTSoldier()`. 

The Resistance Faction soldiers (Reaper, Skirmisher, Templar) in their own `CreateTSoldier()` do a `super.CreateTSoldier()`, and then modify the country and generate the nickname for the generated soldier, which would mean any changes to these aspects of the soldier appearance in the DLC hook would be disregarded.

Example of the original code:
```
function TSoldier CreateTSoldier(optional name CharacterTemplateName, optional EGender eForceGender, optional name nmCountry = '', optional int iRace = -1, optional name ArmorName)
{
	local X2SoldierClassTemplateManager ClassMgr;
	local X2SoldierClassTemplate ClassTemplate;

	kSoldier = super.CreateTSoldier('ReaperSoldier', eForceGender, nmCountry, iRace, ArmorName);
	SetCountry('Country_Reaper');
	ClassMgr = class'X2SoldierClassTemplateManager'.static.GetSoldierClassTemplateManager();
	ClassTemplate = ClassMgr.FindSoldierClassTemplate('Reaper');
	kSoldier.strNickName = GenerateNickname(ClassTemplate, kSoldier.kAppearance.iGender);

	return kSoldier;
}
```
So in order to facilitate better compatibility with the DLC hook, I have added their character generator classes to the Highlander and refactored them along these lines:
```
function TSoldier CreateTSoldier(optional name CharacterTemplateName, optional EGender eForceGender, optional name nmCountry = '', optional int iRace = -1, optional name ArmorName)
{
	kSoldier = super.CreateTSoldier('ReaperSoldier', eForceGender, nmCountry, iRace, ArmorName);
	return kSoldier;
}

function SetCountry(name nmCountry)
{
	kSoldier.nmCountry = 'Country_Reaper';
	kSoldier.kAppearance.nmFlag = kSoldier.nmCountry; // needs to be copied here for pawns -- jboswell
}

function GenerateName( int iGender, name CountryName, out string strFirst, out string strLast, optional int iRace = -1 )
{
	local X2SoldierClassTemplateManager ClassMgr;
	local X2SoldierClassTemplate ClassTemplate;

	super.GenerateName( kSoldier.kAppearance.iGender, kSoldier.nmCountry, kSoldier.strFirstName, kSoldier.strLastName, kSoldier.kAppearance.iRace );

	ClassMgr = class'X2SoldierClassTemplateManager'.static.GetSoldierClassTemplateManager();
	ClassTemplate = ClassMgr.FindSoldierClassTemplate('Reaper');
	kSoldier.strNickName = GenerateNickname(ClassTemplate, kSoldier.kAppearance.iGender);
}
```
The `SetCountry()` and `GenerateName()` functions will be called by the `super.CreateTSoldier()`. This approach is already used by the SPARK character generator class; it is weird the Firaxis didn't use the same approach for the resistance faction soldiers.

The Skirmisher character generator in particular contained a seemingly completely redundant call for `GenerateName()` which would have already been called in the exact same way by the `super.CreateTSoldier()` called a few moments before; it appears not a lot of care was given to this particular part of the code. There is a lot of potential for further refactoring, but I don't wanna fix what ain't broke without a good reason.


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -674,4 +674,21 @@ final function string GetDisplayName()
...
 }
-/// End Issue #524
\ No newline at end of file
+/// End Issue #524
+
+/// Start Issue #783
+/// <summary>
+/// Called from XGCharacterGenerator:CreateTSoldier
+/// Has no return value, just modify the CharGen.kSoldier directly.
+/// HL-Docs: feature:ModifyGeneratedUnitAppearance; issue:783; tags:Customization
```
robojumper:
```suggestion
/// HL-Docs: feature:ModifyGeneratedUnitAppearance; issue:783; tags:customization
```

This may work on Windows, but fails on Unix due to case sensitive file names.

Iridar:
I've confirmed that my additions function properly with this simple example:

```
static function ModifyGeneratedUnitAppearance(XGCharacterGenerator CharGen, optional name CharacterTemplateName, optional EGender eForceGender, optional name nmCountry = '', optional int iRace = -1, optional name ArmorName, optional XComGameState_Unit UnitState, optional XComGameState UseGameState)
{
	if (CharacterTemplateName == 'Soldier')
	{
		CharGen.kSoldier.kAppearance.nmLeftArmDeco = '';
		CharGen.kSoldier.kAppearance.nmRightArmDeco = '';
		CharGen.kSoldier.kAppearance.nmLeftForearm = '';
		CharGen.kSoldier.kAppearance.nmRightForearm = '';
		CharGen.kSoldier.kAppearance.nmThighs = '';
		CharGen.kSoldier.kAppearance.nmTorsoDeco = '';
		CharGen.kSoldier.kAppearance.nmShins = '';
	}
}```
Once (if) this PR goes live, I'm gonna refactor my Clean Soldier Generation mod so it can work without Mod Class Overrides and in a more predictable and streamlined manner.

File: X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc (hunk 1 above)
Comment (Iridar) [reply X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc]: Mmf I've copied the tag exactly as it was on the documentation website, I'm guessing it gets capitalized later at some point. My version of the highlander didn't have examples of that tag used.

**Comment by MrNiceUK** [comment]
#297 isn't directly applicable, any shell listeners would have already been unregistered as well, the issue was that starting rookies are generated before strategy listeners are registered.

**Comment by Iridar** [comment]
> #297 isn't directly applicable, any shell listeners would have already been unregistered as well, the issue was that starting rookies are generated before strategy listeners are registered.

Didn't think they'd get unregistered. Then yeah.

File: X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc [hunk 2]
```diff
@@ -674,4 +674,21 @@ final function string GetDisplayName()
...
+/// This DLC hook allows mods to make arbitrary changes to unit appearance after it has been generated by the Character Generator.
+/// The main argument of the hook function is the instance of the XGCharacterGenerator. 
+/// It contains the "kSoldier" property, which stores the generated appearance, and can be modified by the hook.
+/// You can also use the methods in the Character Generator if you want to "redo" some parts of the generation process. 
+///	However, it's important *not* to call the CreateTSoldier() method, as this will retrigger the hook, potentially causing an inception loop and crashing the game.
+/// The UnitState and the GameState will be passed with the event only if this CreateTSoldier() function was called from CreateTSoldierFromUnit(), 
+/// which normally happens only in the shell code (TQL / Challenge Mode / Character Pool). 
+/// </summary>
+static function ModifyGeneratedUnitAppearance(XGCharacterGenerator CharGen, optional name CharacterTemplateName, optional EGender eForceGender, optional name nmCountry = '', optional int iRace = -1, optional name ArmorName, optional XComGameState_Unit UnitState, optional XComGameState UseGameState)
```
Musashi1584, line 692: Why are all these parameters optional? especially the unit state and game state. Are there calls to the dlc hook where these are actually none?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator.uc
```diff
@@ -380,6 +407,15 @@ function TSoldier CreateTSoldier( optional name CharacterTemplateName, optional
 
 	BioCountryName = kSoldier.nmCountry;
 
+	// Start issue #783
+	/// HL-Docs: ref:ModifyGeneratedUnitAppearance; issue:783
+	DLCInfos = `ONLINEEVENTMGR.GetDLCInfos(false);
```
Musashi1584, line 412: Preferably move this to a separate method

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator.uc
```diff
@@ -339,6 +362,10 @@ function TSoldier CreateTSoldier( optional name CharacterTemplateName, optional
 	local X2SimpleBodyPartFilter BodyPartFilter;
 	local X2CharacterTemplate CharacterTemplate;
 	local TAppearance DefaultAppearance;
+	// Start variables for issue #783
+	local array<X2DownloadableContentInfo> DLCInfos;
+	local int i;
```
Musashi1584, line 367: Nitpicking alert: not a fan of single char variable names
```suggestion
	local int Index;
```

Iridar:
@Musashi1584 Please review this again when you have a moment.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator.uc
Iridar: Done.

Iridar:
Sorry for the mess, deleted branch by accident.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator.uc
```diff
@@ -380,9 +403,28 @@ function TSoldier CreateTSoldier( optional name CharacterTemplateName, optional
...
 
+	// Start issue #783
+	ModifyGeneratedUnitAppearance(self, CharacterTemplateName, eForceGender, nmCountry, iRace, ArmorName, GenerateAppearanceForUnitState, GenerateAppearanceForGameState);
+	// End issue #783
+
 	return kSoldier;
 }
 
+// Start issue #783
+private static function ModifyGeneratedUnitAppearance(XGCharacterGenerator CharGen, optional name CharacterTemplateName, optional EGender eForceGender, optional name nmCountry = '', optional int iRace = -1, optional name ArmorName, optional XComGameState_Unit GenerateAppearanceForUnitState, optional XComGameState GenerateAppearanceForGameState)
```
pledbrook, line 414: Instead of making this static and passing `self` as the argument, this should probably just be an instance function, since it is bound to an instance of `XGCharacterGenerator`.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator_Reaper.uc
```diff
@@ -0,0 +1,133 @@
...
+	MatchArmorTemplateForTorso = (ArmorName == '') ? 'ReaperArmor' : ArmorName;
+	MatchCharacterTemplateForTorso = 'NoCharacterTemplateName'; //Force the selector to use the armor type to filter torsos
+
+	return class'X2CharacterTemplateManager'.static.GetCharacterTemplateManager().FindCharacterTemplate('ReaperSoldier');
+}
+
+// Start issue #783
+// Normally this function calls the super.CreateTSoldier, and then manually sets the country and nickname.
+// In order to make the DLC hook for this issue more compatible with resistance faction soldiers,
+// this functionality has been moved into SetCounry() and GenerateName() methods which will be called by super.CreateTSoldier.
```
pledbrook, line 32:
```suggestion
// this functionality has been moved into SetCountry() and GenerateName() methods which will be called by super.CreateTSoldier.
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator_Reaper.uc
```diff
@@ -0,0 +1,133 @@
...
+	return kSoldier;
+}
+
+function SetCountry(name nmCountry)
+{
+	kSoldier.nmCountry = 'Country_Reaper';
+	kSoldier.kAppearance.nmFlag = kSoldier.nmCountry; // needs to be copied here for pawns -- jboswell
+}
+
+function GenerateName( int iGender, name CountryName, out string strFirst, out string strLast, optional int iRace = -1 )
```
pledbrook, line 45:
```suggestion
function GenerateName(int iGender, name CountryName, out string strFirst, out string strLast, optional int iRace = -1)
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator_Reaper.uc
```diff
@@ -0,0 +1,133 @@
...
+{
+	local X2SoldierClassTemplateManager ClassMgr;
+	local X2SoldierClassTemplate ClassTemplate;
+
+	super.GenerateName( kSoldier.kAppearance.iGender, kSoldier.nmCountry, kSoldier.strFirstName, kSoldier.strLastName, kSoldier.kAppearance.iRace );
```
pledbrook, line 50:
```suggestion
	super.GenerateName(kSoldier.kAppearance.iGender, kSoldier.nmCountry, kSoldier.strFirstName, kSoldier.strLastName, kSoldier.kAppearance.iRace);
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator_Reaper.uc
pledbrook: I see that this is just inconsistency in the original Firaxis code, so I'm not too bothered about this change.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator_Templar.uc
```diff
@@ -21,19 +21,34 @@ function X2CharacterTemplate SetCharacterTemplate(name CharacterTemplateName, na
 	return class'X2CharacterTemplateManager'.static.GetCharacterTemplateManager().FindCharacterTemplate(CharacterTemplateName);
 }
 
+// Start issue #783
+// Normally this function calls the super.CreateTSoldier, and then manually sets the country and nickname.
+// In order to make the DLC hook for this issue more compatible with resistance faction soldiers,
+// this functionality has been moved into SetCounry() and GenerateName() methods which will be called by super.CreateTSoldier.
```
pledbrook, line 27:
```suggestion
// this functionality has been moved into SetCountry() and GenerateName() methods which will be called by super.CreateTSoldier.
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator_Reaper.uc
```diff
@@ -26,19 +26,34 @@ function X2CharacterTemplate SetCharacterTemplate(name CharacterTemplateName, na
...
+// Normally this function calls the super.CreateTSoldier, and then manually sets the country and nickname.
+// In order to make the DLC hook for this issue more compatible with resistance faction soldiers,
+// this functionality has been moved into SetCounry() and GenerateName() methods which will be called by super.CreateTSoldier.
 function TSoldier CreateTSoldier(optional name CharacterTemplateName, optional EGender eForceGender, optional name nmCountry = '', optional int iRace = -1, optional name ArmorName)
+{
+	kSoldier = super.CreateTSoldier('ReaperSoldier', eForceGender, nmCountry, iRace, ArmorName);
+	return kSoldier;
+}
+
+function SetCountry(name nmCountry)
```
pledbrook, line 39: I don't really understand what this change achieves. Isn't `SetCountry()` typically called after `CreateTSoldier()` anyway, overriding any values set by mods using your new hook?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator_Skirmisher.uc
```diff
@@ -19,20 +19,34 @@ function X2CharacterTemplate SetCharacterTemplate(name CharacterTemplateName, na
 	return class'X2CharacterTemplateManager'.static.GetCharacterTemplateManager().FindCharacterTemplate('SkirmisherSoldier');
 }
 
+// Start issue #783
+// Normally this function calls the super.CreateTSoldier, and then manually sets the country and nickname.
+// In order to make the DLC hook for this issue more compatible with resistance faction soldiers,
+// this functionality has been moved into SetCounry() and GenerateName() methods which will be called by super.CreateTSoldier.
```
pledbrook, line 25:
```suggestion
// this functionality has been moved into SetCountry() and GenerateName() methods which will be called by super.CreateTSoldier.
```

pledbrook (commented):
This seems good to me, although I don't really understand the rationale for the faction soldier refactor. It seems particularly bizarre for `SetCountry()` to completely ignore its argument and just set a hard-coded country.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -674,4 +674,21 @@ final function string GetDisplayName()
 	// Equivalent to empty string if not specified in localization
 	return ModDependency.DisplayName;
 }
-/// End Issue #524
\ No newline at end of file
+/// End Issue #524
+
+/// Start Issue #783
+/// <summary>
```
pledbrook, line 680: We don't need to remove the slashes here, but we do need to remove the one from `</summary>`, so this makes things symmetrical:
```suggestion
// Start Issue #783
// <summary>
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -674,4 +674,21 @@ final function string GetDisplayName()
...
+/// Has no return value, just modify the CharGen.kSoldier directly.
+/// HL-Docs: feature:ModifyGeneratedUnitAppearance; issue:783; tags:customization
+/// This DLC hook allows mods to make arbitrary changes to unit appearance after it has been generated by the Character Generator.
+/// The main argument of the hook function is the instance of the XGCharacterGenerator. 
+/// It contains the "kSoldier" property, which stores the generated appearance, and can be modified by the hook.
+/// You can also use the methods in the Character Generator if you want to "redo" some parts of the generation process. 
+///	However, it's important *not* to call the CreateTSoldier() method, as this will retrigger the hook, potentially causing an inception loop and crashing the game.
+/// The UnitState and the GameState will be passed with the event only if this CreateTSoldier() function was called from CreateTSoldierFromUnit(), 
+/// which normally happens only in the shell code (TQL / Challenge Mode / Character Pool). 
+/// </summary>
```
pledbrook, line 691: We don't want this line appearing in the docs:
```suggestion
// </summary>
```

Iridar: Good catch, thank you.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator_Reaper.uc
Iridar: I don't understand what "this change" in particular you're talking about, but the way the code is set up right now, no, the ModifyGeneratedUnitAppearance hook will run after calling SetCountry(). The vanilla Reaper CharGen did call `SetCountry()` *after* calling `CreateTSoldier()`, but with the refactored code it doesn't happen, no.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator.uc
Iridar: It is done that way so that the DLC Hook gets easy access to the Character Generator class the hook is being called from, in case the hook user wants to reuse some of the methods present in that specific class.

pledbrook: Yes, but you can just pass `self` to the DLC hook. If a member function requires access to `self`, then it really should be an instance function/method rather than static.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator_Reaper.uc
pledbrook: Sorry, I wasn't very clear. "This change" refers to the refactored faction soldier character generators. And my point was that `SetCountry()` is typically called by the same code that invokes `CreateTSoldier()` in the first place.

However, I now realise that code is calling `SetCountry()` on the unit state, not the generator. My bad.

Iridar: I see. Thanks for bothering with this at all.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator.uc
```diff
@@ -380,9 +403,28 @@ function TSoldier CreateTSoldier( optional name CharacterTemplateName, optional
...
+{
+	local array<X2DownloadableContentInfo> DLCInfos;
+	local int i;
+	
+	/// HL-Docs: ref:ModifyGeneratedUnitAppearance; issue:783
+	DLCInfos = `ONLINEEVENTMGR.GetDLCInfos(false);
+	for (i = 0; i < DLCInfos.Length; i++)
+	{
+		DLCInfos[i].ModifyGeneratedUnitAppearance(CharGen, CharacterTemplateName, eForceGender, nmCountry, iRace, ArmorName, GenerateAppearanceForUnitState, GenerateAppearanceForGameState);
```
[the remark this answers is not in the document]
pledbrook: Also, repeating Musashi's question, why are all the arguments optional? The only caller passes all of the arguments anyway, so you're not making uses of the `optional`. I also wonder whether it's reasonable for mods to expect all of these arguments to potentially have default/empty values.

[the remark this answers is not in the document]
Iridar: > Yes, but you can just pass self to the DLC hook. If a member function requires access to self, then it really should be an instance function/method rather than static.

Right, make sense. Sorry it took a while. Changes made.

> why are all the arguments optional?

Because they are `optional` in the CreateTSoldier() function that calls the DLC hook, and they are not always there when CreateTSoldier() is called, for example `UITLE_SkirmishModeMenu.uc` has `Soldier = CharacterGenerator.CreateTSoldier();`.

Iridar:
The status shows "Changes requested", but they have already been applied. I have squashed most of the commits in this PR. If everything is in order, I'm ready to mark this "ready to merge".

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -762,3 +762,20 @@ final function string GetDisplayName()
...
+/// This DLC hook allows mods to make arbitrary changes to unit appearance after it has been generated by the Character Generator.
+/// The main argument of the hook function is the instance of the XGCharacterGenerator. 
+/// It contains the "kSoldier" property, which stores the generated appearance, and can be modified by the hook.
+/// You can also use the methods in the Character Generator if you want to "redo" some parts of the generation process. 
+///	However, it's important *not* to call the CreateTSoldier() method, as this will retrigger the hook, potentially causing an inception loop and crashing the game.
+/// The UnitState and the GameState will be passed to this function only if this CreateTSoldier() function was called from CreateTSoldierFromUnit(), 
+/// which normally happens only in the shell code (TQL / Challenge Mode / Character Pool). 
+// </summary>
+static function ModifyGeneratedUnitAppearance(XGCharacterGenerator CharGen, optional name CharacterTemplateName, optional EGender eForceGender, optional name nmCountry = '', optional int iRace = -1, optional name ArmorName, optional XComGameState_Unit UnitState, optional XComGameState UseGameState)
```
robojumper, line 779: I still think these arguments should not be optional, considering this hook is always called with something in place of the arguments.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator_Reaper.uc
```diff
@@ -0,0 +1,133 @@
...
+// Normally this function calls the super.CreateTSoldier, and then manually sets the country and nickname.
+// In order to make the DLC hook for this issue more compatible with resistance faction soldiers,
+// this functionality has been moved into SetCountry() and GenerateName() methods which will be called by super.CreateTSoldier.
+function TSoldier CreateTSoldier(optional name CharacterTemplateName, optional EGender eForceGender, optional name nmCountry = '', optional int iRace = -1, optional name ArmorName)
+{
+	kSoldier = super.CreateTSoldier('ReaperSoldier', eForceGender, nmCountry, iRace, ArmorName);
+	return kSoldier;
+}
+
+function SetCountry(name nmCountry)
```
robojumper, line 39: It would be great if the high-level idea behind these changes were described in a "compatibility" section. There are a lot of mods with custom CharacterGenerators and some info on how to make these compatible with the new hook in the same way that the existing hero ones were would be useful, especially since we *might* want to revisit this for DLC_3/Sparks.

robojumper (changes requested):
Approach LGTM.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
Iridar: I'm a bit unclear on the policy here.  The "X2DLCInfo::ModifyGeneratedUnitAppearance()` is called by `XGCharacterGenerator::ModifyGeneratedUnitAppearance`, which in turn is called by `XGCharacterGenerator::CreateTSoldier()`, and all of its arguments are `optional`. I would assume that if a function is calling another function, passing its own arguments to it, the `optional` status should be inherited.

Nothing is stopping mods from calling `CreateTSoldier()` with arbitrary arguments to achieve arbitrary things, so I think marking the arguments in the "X2DLCInfo::ModifyGeneratedUnitAppearance()` as `optional` is a good way of communicating that there is a possibility of them being `none`.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator_Reaper.uc
Iridar: How would I do that? Add a `compatibility` tag to HL-Docs, and add some text about "blah blah now you don't have to MCO XGCharacterGenerator"?

robojumper: Not entirely. What I meant is that this hook only works for `XGCharacterGenerator` classes that explicitly received the `Set*` functions rewiring. Custom `XGCharacterGenerator` classes are in the same state as the Reaper/Skirmisher/Templar `XGCharacterGenerator` class before this PR, which would overwrite a bunch of customization properties, overriding the DLC hook.

I'd like this PR to describe the problem and the steps needed to make an `XGCharacterGenerator` work with the DLC hook in a way where it doesn't potentially overwrite DLC hook changes, both so that modders can make their `XGCharacterGenerator`s 100% compatible with the DLC hook and so that we have a reference point for the Spark `XGCharacterGenerator` if we chose to do a DLC_3 HL.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
robojumper: I see the documentation value. My comment was motivated by the fact that `optional` only allows omitting arguments on the caller side, and we as a caller want to ensure we always pass all the arguments we have -- even if their source might be optional (like they are in `CreateTSoldier`).

Additionally, not all arguments are passed verbatim. `CreateTSoldier` does this, for example:

	if (nmCountry == '')
		nmCountry = PickOriginCountry();

so `nmCountry` is (hopefully) never empty.

Iridar:
I have adjusted the documentation, removed `optional` from the arguments, and put `const` in there for most of them just in case.

robojumper (approved):
Thanks!

pledbrook:
@Iridar Would you mind squashing and merging this? I would do it myself, but the commit messages aren't descriptive of the change and I have no idea what to write in the squashed commit.

Iridar:
> @Iridar Would you mind squashing and merging this? I would do it myself, but the commit messages aren't descriptive of the change and I have no idea what to write in the squashed commit.

I squashed and rebased, but I don't have the authoritah to merge.

Issue #356, opened 2018-12-04
Hardcoded Material lists for tinting


MrNiceUK:
In `XComHumanPawn::UpdateMeshMaterials()`, a switch block on the name of the Parent Material(recursive) of the Material is used to decide whether to (attempt) to apply skin color, armor tints, tattoos, etc. This means that these are applied based on hard coded lists of Material names. This doesn't entirely stop mods having tintable custom materials, but they do have to have the same *name* as a material which takes the appropriate tints in the hard coded lists, and rely on the fact that classes from different UPKs can have the same name. However this tops the materials being appropriately named, and encourages bad practice by *knowingly* reusing a class name, both of which can lead to confusion.
The obvious "fix" for this is to change the switch block into a set of `else if`'s, with config defined arrays of appropriate material names which can be appended to by mods. Mods relying on the existing behaviour by reusing existing material names will still work.
Note: Yes, #216 hook could be used, but requires that mods duplicate the material loop, and if multiple mods use the same material, each would have to have the code to not have dependencies.

Issue #216, opened 2018-03-31
More general Material Update hooks for XComHumanPawn


robojumper:
#169 adds a hook for mods to update human pawn mesh materials. The hook has some limitations:

1. It's called for every attachment for every material instance constant
2. It's only called for material instance constants, not for plain old materials.
3. There's not enough context for mods. There's not even an index for mods to identify which material they're actually looking at.

Adding an additional hook that is called once per attachment would be better, because it's much more flexible. We can rewrite the code to deprecate the old hook by making the new function a wrapper for the old function.

Issue #197, opened 2018-02-22
Fix and configuration of Second Wave Options in Debug Strategy Start


ChrisTheThinMint:
Currently, you are unable to select any SWOs for the Debug start, and it also automatically enables Beta Strike, which is likely a leftover from Firaxis' internal testing.

I am going to go ahead and overwrite XGStrategy to include a config array where you can customize the second wave options enabled in the Debug Start.

PR #797, opened 2020-03-04
Even more docs (DebugStartSecondWave, UpdateHumanPawnMeshMaterial & friends)

robojumper:
Documents #197, #216, #169, #356. (Remove needs-docs label when this is merged.)

cc @Iridar I hope this is somewhat correct re. material hooks.


robojumper:
I believed you used the material hook too and wanted to ask whether the documentation is correct (could be wrong, in that case ignore).

Iridar:
> I believed you used the material hook too and wanted to ask whether the documentation is correct (could be wrong, in that case ignore).

Nope, never used it for anything ever.

PR #798, opened 2020-03-04
Input docs (SubscribeToOnInput, SubscribeToOnInputForScreen)

robojumper:
Documents #198, #501. (Remove needs-docs tag when this is merged).

cc @Xymanek

### DIFF
--- a/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_UIScreenListener_ShellSplash.uc
+++ b/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_UIScreenListener_ShellSplash.uc
@@ -13,9 +13,9 @@ event OnInit(UIScreen Screen)
 	// using this input hook is most useful when the HL *isn't* working -- but then,
 	// this input hook doesn't exist. Not much we can do other than ensuring the
 	// user makes it to the main menu screen, sees the error message and checks the log.
-	if (Function'XComGame.UIScreenStack.SubscribeToOnInput' != none)
+	if (Function'XComGame.UIScreenStack.SubscribeToOnInputForScreen' != none)
 	{
-		Screen.Movie.Stack.SubscribeToOnInput(OnInputHook);
+		Screen.Movie.Stack.SubscribeToOnInputForScreen(Screen, OnInputHook);
 	}
 
 	RealizeVersionText(UIShell(Screen));
@@ -31,13 +31,6 @@ event OnReceiveFocus(UIScreen Screen)
 
 event OnRemoved(UIScreen Screen)
 {
-	if(UIShell(Screen) == none || !bEnableVersionDisplay)  // this captures UIShell and UIFinalShell
-		return;
-
-	if (Function'XComGame.UIScreenStack.UnsubscribeFromOnInput' != none)
-	{
-		Screen.Movie.Stack.UnsubscribeFromOnInput(OnInputHook);
-	}
 }
 
 function RealizeVersionText(UIShell ShellScreen)
@@ -208,16 +201,15 @@ function OnHitboxMouseEvent(UIPanel control, int cmd)
 	}
 }
 
-function bool OnInputHook(int iInput, int ActionMask)
+function bool OnInputHook(UIScreen Screen, int iInput, int ActionMask)
 {
 	local UIText TooltipText;
 	local UIBGBox TooltipBG;
 	local UIShell ShellScreen;
 
-	if (iInput == class'UIUtilities_Input'.const.FXS_BUTTON_R3 && (ActionMask & class'UIUtilities_Input'.const.FXS_ACTION_RELEASE) != 0
-		&& `SCREENSTACK.IsCurrentScreen(class'UIShell'.Name))
+	if (iInput == class'UIUtilities_Input'.const.FXS_BUTTON_R3 && (ActionMask & class'UIUtilities_Input'.const.FXS_ACTION_RELEASE) != 0)
 	{
-		ShellScreen = UIShell(`SCREENSTACK.GetFirstInstanceOf(class'UIShell'));
+		ShellScreen = UIShell(Screen);
 
 		TooltipBG = UIBGBox(ShellScreen.GetChildByName('theTooltipBG', false));
 		TooltipText = UIText(ShellScreen.GetChildByName('theTooltipText', false));
--- a/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_UIScreenListener_ShellSplash.uc
+++ b/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_UIScreenListener_ShellSplash.uc
@@ -29,10 +29,6 @@ event OnReceiveFocus(UIScreen Screen)
 	RealizeVersionText(UIShell(Screen));
 }
 
-event OnRemoved(UIScreen Screen)
-{
-}
-
 function RealizeVersionText(UIShell ShellScreen)
 {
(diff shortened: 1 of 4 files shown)


X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_UIScreenListener_ShellSplash.uc
```diff
@@ -31,13 +31,6 @@ event OnReceiveFocus(UIScreen Screen)
...
 event OnRemoved(UIScreen Screen)
 {
-	if(UIShell(Screen) == none || !bEnableVersionDisplay)  // this captures UIShell and UIFinalShell
-		return;
-
-	if (Function'XComGame.UIScreenStack.UnsubscribeFromOnInput' != none)
-	{
-		Screen.Movie.Stack.UnsubscribeFromOnInput(OnInputHook);
-	}
 }
```
Xymanek: Maybe remove `event OnRemoved` entirely?

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -782,6 +782,23 @@ simulated function bool IsNotInStack( class<UIScreen> ScreenClass, optional bool
 }
 
 // start issue #198
+/// HL-Docs: feature:SubscribeToOnInput; issue:198; tags:
+/// Mods may want to intercept mouse/keyboard/controller input and instead run their own code.
```
Xymanek: Afaik mouse events follow a different code path and won't get sent to these delegates

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -782,6 +782,23 @@ simulated function bool IsNotInStack( class<UIScreen> ScreenClass, optional bool
 }
 
 // start issue #198
+/// HL-Docs: feature:SubscribeToOnInput; issue:198; tags:
```
Xymanek: Empty tags?

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -823,6 +840,79 @@ simulated function bool ModOnInput(int iInput, int ActionMask)
 // end issue #198
 
 // Start issue #501
+/// HL-Docs: feature:SubscribeToOnInputForScreen; issue:501; tags:
+/// Mods may want to intercept mouse/keyboard/controller input on certain screens and instead run their own code.
```
Xymanek: Same comments regarding empty tags and mouse

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -823,6 +840,79 @@ simulated function bool ModOnInput(int iInput, int ActionMask)
...
+/// function SubscribeToOnInputForScreen(UIScreen Screen, delegate<CHOnInputDelegateImproved> Callback);
+/// function UnsubscribeFromOnInputForScreen(UIScreen Screen, delegate<CHOnInputDelegateImproved> Callback);
+/// ```
+///
+/// In a nutshell, with `SubscribeToOnInputForScreen` you ask the UIScreenStack
+/// "when screen `Screen` would receive input, ask me first".
+/// This is done with the `CHOnInputDelegateImproved` delegate, which defines the form
+/// of the callback function called when the targeted screen would receive input.
+///
+/// Your function receives the screen that would have received the input (`Screen`), the button (`iInput`),
```
Xymanek: This is slightly confusing

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -823,6 +840,79 @@ simulated function bool ModOnInput(int iInput, int ActionMask)
...
+/// 
+/// 	return false;
+/// }
+/// ```
+///
+/// `CheckInputIsReleaseOrDirectionRepeat` ensures that the button was just released (or, if directional button,
+/// held for a long time), making input behavior more consistent with base game screens.
+///
+/// This feature is a more convenient version of [`SubscribeToOnInput`](./SubscribeToOnInput), which receives
+/// events for any screen and has to be manually unsubscribed. It offers lower-level interaction with the input system,
```
Xymanek: "It" is ambiguous

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
```diff
@@ -823,6 +840,79 @@ simulated function bool ModOnInput(int iInput, int ActionMask)
...
+///
+/// event OnInit (UIScreen Screen)
+/// {
+/// 	local UIMission MissionScreen;
+///
+/// 	MissionScreen = UIMission(Screen);
+/// 	if (MissionScreen == none) return;
+///
+/// 	// Spawn button
+/// 	// ...
```
Xymanek: I don't think this comment needs to be in this example, it's just confusing. Or it needs more meaning/elaboration

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIScreenStack.uc
robojumper: Well, the right mouse click can definitely be intercepted. I clarified this later in the documentation.

Xymanek: Are you sure? FXS lists `FXS_R_MOUSE_DOWN` under mouse section, which is used inside `OnMouseEvent`, which is called from `UIMovie::FlashRaiseMouseEvent` and bypasses the screen stack

```uc
// MOUSE ************************

const FXS_MOUSE_RANGE_BEGIN	= 389;

const FXS_L_MOUSE_DOWN		= 390;
const FXS_L_MOUSE_UP		= 391;
const FXS_L_MOUSE_IN		= 392;
const FXS_L_MOUSE_OUT		= 393;
const FXS_L_MOUSE_DRAG_OVER	= 394;
const FXS_L_MOUSE_DRAG_OUT	= 395;
const FXS_L_MOUSE_RELEASE_OUTSIDE = 396;
const FXS_L_MOUSE_OVER		= 397;
const FXS_L_MOUSE_DOUBLE_UP = 398;
const FXS_L_MOUSE_DOUBLE_DOWN = 399;
// Objects that care about double click events that also perform an action on 
// single click need to listen to this event for their single click action.
const FXS_L_MOUSE_UP_DELAYED = 400;

const FXS_R_MOUSE_DOWN		= 405;
```

Xymanek: Just tested it.

Your initial description was/is correct - mouse events are routed through both flash panel-specific handlers and the "global" ScreenStack pipeline, I tested with `FXS_L_MOUSE_DOWN` but at this point I would assume  that it holds true for all mouse events

robojumper: It's still not a good idea to try and intercept things like button presses, so I'm keeping it this way.

Xymanek: I don't think we should "deny knowledge" based on the fact/opinion that it's a bad solution. A disclaimer (the red "popup" like you posted in discord) would be an appropriate warning though.

Besides, I don't think marking input event as handled in the screen stack pipeline prevents flash from receiving it

robojumper: Fixed.

PR #806, opened 2020-03-16
Add the OverKillDamage unit value #805

### LINKED ISSUE #805: Add an event that allows Abilities to work with Overkill Damage
Similiar to issue #91, But not just exclusive to the bleed out effect. Currently it can't be done because the relevant events only pass the unit states, whose DamageResults arrays don't store overkill damage, just the damage the unit has taken (so if a 3hp unit gets shot for 10 damage it will register only 3).

Grobobobo:
Fixes #805 . After discussing it on discord, we decided that making a seperate event was unnescessary, and just the new unit value would do.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -6154,6 +6154,8 @@ event TakeDamage( XComGameState NewGameState, const int DamageAmount, const int 
 	OverkillDamage = (GetCurrentStat( eStat_HP )) - DmgResult.DamageAmount;
 	if (OverkillDamage <= 0)
 	{
+		SetOverKillUnitValue(OverkillDamage);
+
 		// Issue #202 Start, allow listeners to override killed by explosion
 		KilledByExplosionTuple = new class'XComLWTuple';
 		KilledByExplosionTuple.Id = 'OverrideKilledByExplosion';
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -14682,6 +14684,23 @@ function int GetNumHeavyWeapons(optional XComGameState CheckGameState)
...
 // End Issue #171
 
+//Begin Issue #805
+/// HL-Docs: feature:OverKillDamage; issue:801; tags:tactical
+/// The UnitState's damage results array only holds the actual damage taken by the unit, so the result can't be higher than the unit's HP.
+/// This adds the OverkillDamage Unit value, which is shows how higher the kill damage value was from the standard Unit HP.
+///	One of its use cases is to modify the effects of the abilities that trigger on death, like the trigger chance on Advent Priest's Sustain.
+/// ```unrealscript
+/// 	SetUnitFloatValue('OverKillDamage', -OverkillDamage, eCleanup_BeginTurn);
+/// ```
```
Xymanek: Why is this here?

Grobobobo: To show the function in the documentation? I thought it would be better to include it for additional clarity.

Xymanek: Feels very out-of-place... I do not think that there is any value in showing the implementation code (eg. we don't dump tuple code for event-based hooks into docs), but I think you should specify that the value uses `eCleanup_BeginTurn`

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -14682,6 +14684,23 @@ function int GetNumHeavyWeapons(optional XComGameState CheckGameState)
 }
 // End Issue #171
 
+//Begin Issue #805
+/// HL-Docs: feature:OverKillDamage; issue:801; tags:tactical
```
Xymanek: So which issue number?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -14682,6 +14684,23 @@ function int GetNumHeavyWeapons(optional XComGameState CheckGameState)
...
+/// ```
+/// The OverKillDamage calculated by XCGS_Unit is negative, but the unit value is set to be positive to make using it more intuitive.
+private function SetOverKillUnitValue(int OverKillDamage)
+{
+	SetUnitFloatValue('OverKillDamage', -OverkillDamage, eCleanup_BeginTurn);
+}
+	//	End issue #805
+
+
+
```
Xymanek: Indentation + extraneous space

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -6154,6 +6154,8 @@ event TakeDamage( XComGameState NewGameState, const int DamageAmount, const int
 	OverkillDamage = (GetCurrentStat( eStat_HP )) - DmgResult.DamageAmount;
 	if (OverkillDamage <= 0)
 	{
+		SetOverKillUnitValue(OverkillDamage);
```
Xymanek: Missing `// Issue #805` comment

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
MrNiceUK: Since `LastEffectDamage` uses `eCleanup_BeginTactical`, I think `OverKillDamage` should too for consistency.

PR #813, opened 2020-03-19
Fix covert action report if rewards prevented (#810)

### LINKED ISSUE #810: "Gained XP!" still displays for soldier when a covert action does not give rewards
The implementation for #438 added a 'CovertAction_PreventGiveRewards' event that allows listeners to stop the covert action awarding its rewards to the player. However, this also blocks the awarding of XP. That's not necessarily a problem per se, but then the covert actions report still displays the soldiers as gaining XP and bond cohesion.

pledbrook:
The covert action report will no longer display soldiers as gaining XP or cohesion if the rewards have been prevented from being given. This is because the reward prevention code also prevents the soldiers from gaining XP and cohesion.

I have also added documentation for the CovertAction_PreventGiveRewards event.

Fixes #810.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActionStaffSlot.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActionStaffSlot.uc
@@ -171,7 +171,10 @@ function UpdateData()
 				PromoteLabel = (Unit.ShowPromoteIcon()) ? class'UISquadSelect_ListItem'.default.m_strPromote : "";
 				Value4 = Caps(ActionState.GetStaffRisksAppliedString(StaffIndex)); // Value 4 is automatically red / negative!
 
-				if (!Unit.bCaptured)
+				// Issue #810: Don't display XP and cohesion gain if rewards weren't
+				// given on completion of the covert action (since XP and cohesion are
+				// not granted in that case).
+				if (!Unit.bCaptured && !GetAction().RewardsNotGivenOnCompletion)
 				{
 					Value1 = m_strGainedXP; // Gained Experience
 					CohesionUnitNames = GetCohesionRewardUnits();
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActionStaffSlot.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActionStaffSlot.uc
@@ -174,7 +174,7 @@ function UpdateData()
 				// Issue #810: Don't display XP and cohesion gain if rewards weren't
 				// given on completion of the covert action (since XP and cohesion are
 				// not granted in that case).
-				if (!Unit.bCaptured && !GetAction().RewardsNotGivenOnCompletion)
+				if (!Unit.bCaptured && !ActionState.RewardsNotGivenOnCompletion)
 				{
 					Value1 = m_strGainedXP; // Gained Experience
 					CohesionUnitNames = GetCohesionRewardUnits();
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActionStaffSlot.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActionStaffSlot.uc
@@ -174,6 +174,7 @@ function UpdateData()
 				// Issue #810: Don't display XP and cohesion gain if rewards weren't
 				// given on completion of the covert action (since XP and cohesion are
 				// not granted in that case).
+				/// HL-Docs: ref:CovertAction_PreventGiveRewards; issue:810
 				if (!Unit.bCaptured && !ActionState.RewardsNotGivenOnCompletion)
 				{
 					Value1 = m_strGainedXP; // Gained Experience
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
@@ -723,9 +723,9 @@ function GiveRewards(XComGameState NewGameState)
 /// on the covert action get no XP or cohesion. The results are saved and are used to adjust UI later.
 ///
 /// ```event
-/// EventID: CovertAction_PreventGiveRewards
-/// EventData: Data: [ inout bool PreventGiveRewards ]
-/// EventSource: XComGameState_CovertAction (CAState)
+/// EventID: CovertAction_PreventGiveRewards,
+/// EventData: Data: [ inout bool PreventGiveRewards ],
+/// EventSource: XComGameState_CovertAction (CAState),
 /// NewGameState: no
 /// ```
 private function bool TriggerPreventGiveRewards()
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
@@ -716,16 +716,11 @@ function GiveRewards(XComGameState NewGameState)
 }
 
 // Start Issue #438, #810
-//
-// *Note* This is private to avoid potential problems with side effects
-// and non-idempotent listeners if other parts of the code (or mods) attempt
-// to use it themselves.
-//
 /// HL-Docs: feature:CovertAction_PreventGiveRewards; issue:438; tags:strategy
 /// Fires an event that allows listeners to prevent the covert action from
 /// awarding the action's rewards to the player. Note that if the `PreventGiveRewards`
 /// boolean is `true` then not only are the rewards not given, but the soldiers
-/// on the covert action get no XP or cohesion.
+/// on the covert action get no XP or cohesion. The results are saved and are used to adjust UI later.
 ///
 /// ```unrealscript
 /// EventID: CovertAction_PreventGiveRewards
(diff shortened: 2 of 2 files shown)


pledbrook:
I don't know, it seems more of a bug that the report displays XP gain when the soldiers aren't actually gaining XP.

pledbrook:
Made the trigger function non-static and clarified that it's for internal use only and hence not covered by the backwards compatibility policy.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
```diff
@@ -643,6 +634,41 @@ function GiveRewards(XComGameState NewGameState)
 	}
 }
 
+// Start Issue #438, #810
+//
+// *Note* This is not part of the highlander's public API, so it is not covered
```
Xymanek, line 639:
```suggestion
// *Note* This function (not the event) is not part of the highlander's public API, so it is not covered
```

pledbrook:
@Xymanek I've just realised that the reward popup is shown after the covert action report screen is closed. For example, if a reward was a scientist and the reward was blocked by a listener, the "scientist awarded" popup still appears.

At this point, I'm wondering whether it makes sense to add a "bRewardsNotGiven" boolean to `XCGS_CovertAction`, which is set after the event is fired the first time. Then the other code that depends on whether the reward was given or not can read it from the action state rather than triggering the event again. Thoughts?

Xymanek:
It will definitely prevent issues with non-idempotent listeners. However, adding state variables can always cause issues with in-progress saves, so need to be careful

pledbrook:
I've now switched to using a state variable whose default value is `false`, which should mean it has no impact on existing campaigns.

Iridar (approved):
LGTM, though I don't know much about covert actions, and haven't tested this.

robojumper (commented):
Approach LGTM. Docs should be updated to the "new" format.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActionStaffSlot.uc
```diff
@@ -171,7 +171,10 @@ function UpdateData()
 				PromoteLabel = (Unit.ShowPromoteIcon()) ? class'UISquadSelect_ListItem'.default.m_strPromote : "";
 				Value4 = Caps(ActionState.GetStaffRisksAppliedString(StaffIndex)); // Value 4 is automatically red / negative!
 
-				if (!Unit.bCaptured)
+				// Issue #810: Don't display XP and cohesion gain if rewards weren't
+				// given on completion of the covert action (since XP and cohesion are
+				// not granted in that case).
+				if (!Unit.bCaptured && !GetAction().RewardsNotGivenOnCompletion)
```
Xymanek, line 177:
```suggestion
				if (!Unit.bCaptured && !ActionState.RewardsNotGivenOnCompletion)
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActionStaffSlot.uc
```diff
@@ -171,7 +171,10 @@ function UpdateData()
 				PromoteLabel = (Unit.ShowPromoteIcon()) ? class'UISquadSelect_ListItem'.default.m_strPromote : "";
 				Value4 = Caps(ActionState.GetStaffRisksAppliedString(StaffIndex)); // Value 4 is automatically red / negative!
 
-				if (!Unit.bCaptured)
+				// Issue #810: Don't display XP and cohesion gain if rewards weren't
+				// given on completion of the covert action (since XP and cohesion are
+				// not granted in that case).
```
Xymanek, line 176: Add `HL-Docs: ref:`

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
```diff
@@ -48,6 +48,17 @@ var() config int							BondmateBonusHours; // The number of hours the duration w
...
 
+// Start Issue #810
+//
+// Indicates whether the covert action rewards were given to the player
+// on completion. By default this is `false`, indicating that the rewards
+// *were* given.
+//
+// Note that this variable has no value until the covert action has been
+// completed.
+var bool									RewardsNotGivenOnCompletion;
```
Xymanek, line 59:
```suggestion
var privatewrite bool RewardsNotGivenOnCompletion;
```

To prevent desync, IMO the variable should only be settable by the event's result

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
```diff
@@ -712,6 +715,42 @@ function GiveRewards(XComGameState NewGameState)
 	}
 }
 
+// Start Issue #438, #810
+//
+// *Note* This is private to avoid potential problems with side effects
+// and non-idempotent listeners if other parts of the code (or mods) attempt
+// to use it themselves.
```
Xymanek, line 722: I do not see a purpose for this comment. 

* `Trigger[...]` methods are private all over the CHL and none of them get a dedicated "this is private ..." comment
* If I want to be naughty, I would just copy paste the code and trigger the event manually

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
```diff
@@ -712,6 +715,42 @@ function GiveRewards(XComGameState NewGameState)
...
+//
+/// HL-Docs: feature:CovertAction_PreventGiveRewards; issue:438; tags:strategy
+/// Fires an event that allows listeners to prevent the covert action from
+/// awarding the action's rewards to the player. Note that if the `PreventGiveRewards`
+/// boolean is `true` then not only are the rewards not given, but the soldiers
+/// on the covert action get no XP or cohesion.
```
Xymanek, line 728: I think it makes sense to also mention that "the results are saved and are used to adjust UI later"

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
```diff
@@ -2246,6 +2285,15 @@ simulated public function ActionRewardPopups()
...
 	local int idx;
 
+	// Start Issue #810
+	if (RewardsNotGivenOnCompletion)
+	{
+		// A mod or some other source prevented this reward from being
+		// given to the player, so don't display the popups.
+		return;
+	}
+	// End Issue #810
```
Xymanek, line 2295: I see no reason why this would prevent the `TriggerNextCovertActionPopup` part. IMO this check should be after that block.

Also. add `HL-Docs: ref:`

Xymanek (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
pledbrook: Do you mean you don't see why it *should* prevent the popups? If so, probably agreed, although I'm not sure which popup the `NextCovertActionPopup` is.

pledbrook: Ah right, they're in reverse order. So this is the one that gives the player the option to go straight to the resistance ring to queue up another covert action.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
```diff
@@ -712,6 +715,37 @@ function GiveRewards(XComGameState NewGameState)
...
+/// on the covert action get no XP or cohesion. The results are saved and are used to adjust UI later.
+///
+/// ```unrealscript
+/// EventID: CovertAction_PreventGiveRewards
+/// EventData: XComLWTuple {
+///     Data: [ inout bool PreventGiveRewards ]
+/// }
+/// EventSource: XComGameState_CovertAction
+/// NewGameState: no
+/// ```
```
Xymanek, line 732: Needs update to the current doc format

Xymanek (changes requested):

pledbrook:
Fixed docs (after multiple tries). This should be squash merged once approved and ready for merge.

Xymanek (approved):

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc
pledbrook: And this just broke some code in LWOTC 😆 Now I need to work out a way for a failed Ambush mission to fail the covert action.

Xymanek: Do you want to revert the PR while you figure it out?

pledbrook: No, it's fine. Fortunately the ambush mission completes before the prevent-give-rewards event fires. So I'm just storing the mission outcome in a game state object that my prevent-give-rewards listener reads to determine whether the CA failed or not. A much better solution that modifying that variable directly :smile: So all's Gucci.

Issue #435, opened 2019-02-26
Better action cleanup on end-of-month


Xymanek:
https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/ba117b928f4acae9bac89d90da5abd3fe3d88f57/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc#L931-L941

Add an event, `ShouldCleanupCovertAction` which allows mods to overwrite whether an action should be deleted

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/ba117b928f4acae9bac89d90da5abd3fe3d88f57/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc#L708-L713

Replace `CovertActions.Length = 0` with a loop that checks whether each action is deleted or not. This is actually a base game bug (removing ongoing action from that list) but the vanilla screen somehow accounted for that so it was never visible

PR #820, opened 2020-03-25
Fix covert actions not being cleaned up (#819)

### LINKED ISSUE #819: Covert actions are not being cleaned up properly on month end
Through debugging, I've discovered that [this loop](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blame/master/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc#L979) in `XCGS_ResistanceFaction.CleanupFactionCovertActions()` is broken because the array is being modified (elements removed) while the function is iterating over it. That leads to some of the covert actions been skipped and left in the `CovertActions` property.

pledbrook:
Issue #435 removed the unconditional clearing of a resistance faction's covert actions when new ones were generated. This led to the manifestation of an existing bug in a loop in `CleanupFactionCovertActions()`, which meant that not all of the existing covert actions were being removed.

This change simply replaces the loop with a reverse-traversal one that can handle elements of the `CovertActions` array being removed during iteration over that array.

Fixes #819.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc
@@ -962,9 +962,9 @@ function RemoveCovertAction(StateObjectReference ActionRef)
 function CleanUpFactionCovertActions(XComGameState NewGameState)
 {
 	local XComGameStateHistory History;
+	local int idx;
 	// Issue #435 Start
 	local XComGameState_CovertAction ActionState;
-	local StateObjectReference ActionRef;
 	local XComLWTuple Tuple;
 
 	Tuple = new class'XComLWTuple';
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc
```diff
@@ -976,10 +976,13 @@ function CleanUpFactionCovertActions(XComGameState NewGameState)
 	
 	History = `XCOMHISTORY;
 
-	foreach CovertActions(ActionRef) // Issue #435 Changed Loop type
+	// Issue #819: Replaced the `foreach` with a reverse traversal loop which won't
+	// be affected by elements being removed during the iteration. Also removed the
+	// `ActionRef` local variable.
```
Xymanek:
```suggestion
	// `ActionRef` local variable (introduced in #435).
```

pledbrook:
This should be squash merged

Issue #825, opened 2020-03-27
Trigger an Event in `XComUnitPawn.PlayHitEffects`


Musashi1584:
The event should allow overiding the hardcoded templar fx for `eHit_Parry`, `eHit_Reflect` and `eHit_Deflect` which play for any class using abilities that utilizing these hit results.

Iridar:
PR closing this Issue was merged already, and it already has proper docs. Closing.

PR #826, opened 2020-03-27
Trigger an Event in XComUnitPawn.PlayHitEffects

Musashi1584:
Implements Issue #825

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
@@ -381,19 +381,20 @@ simulated function DamageTypeHitEffectContainer GetDamageTypeHitEffectContainer(
 /// Allows listeners to override the default behavior of XComUnitPawn.PlayHitEffects
 /// This is especially useful for preventing the hardcoded templar fx for 
 /// eHit_Parry, eHit_Reflect and eHit_Deflect which play for any abilities that utilizing these hit results.
+/// If OverrideHitEffect is set to true the PlayHitEffects function will return early and the default behavior is ommited.
 ///
 /// ```unrealscript
 /// EventID: OverrideHitEffects
 /// EventData: XComLWTuple {
 ///     Data: [
 ///       out bool OverrideHitEffect,
-///       inout float Damage
-///       inout Actor InstigatedBy
-///       inout vector HitLocation
-///       inout name DamageTypeName
-///       inout vector Momentum
-///       inout bool bIsUnitRuptured
-///       inout EAbilityHitResult HitResult
+///       inout float Damage,
+///       inout Actor InstigatedBy,
+///       inout vector HitLocation,
+///       inout name DamageTypeName,
+///       inout vector Momentum,
+///       inout bool bIsUnitRuptured,
+///       inout EAbilityHitResult HitResult,
 ///     ]
 /// }
 /// EventSource: self (XComUnitPawn)
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
@@ -377,6 +377,66 @@ simulated function DamageTypeHitEffectContainer GetDamageTypeHitEffectContainer(
 	return DamageEffectContainer;
 }
 
+/// HL-Docs: feature:OverrideHitEffects; issue:825; tags:tactical
+/// Allows listeners to override the default behavior of XComUnitPawn.PlayHitEffects
+/// This is especially useful for preventing the hardcoded templar fx for 
+/// eHit_Parry, eHit_Reflect and eHit_Deflect which play for any abilities that utilizing these hit results.
+///
+/// ```unrealscript
+/// EventID: OverrideHitEffects
+/// EventData: XComLWTuple {
+///     Data: [
+///       out bool OverrideHitEffect,
+///       inout float Damage
+///       inout Actor InstigatedBy
+///       inout vector HitLocation
+///       inout name DamageTypeName
+///       inout vector Momentum
+///       inout bool bIsUnitRuptured
+///       inout EAbilityHitResult HitResult
+///     ]
+/// }
+/// EventSource: self (XComUnitPawn)
+/// NewGameState: no
+/// ```
+simulated function bool TriggerOnOverrideHitEffects(
+	float Damage,
+	Actor InstigatedBy,
+	vector HitLocation,
+	name DamageTypeName,
+	vector Momentum,
+	bool bIsUnitRuptured,
+	EAbilityHitResult HitResult
+)
+{
+	local XComLWTuple Tuple;
+
+	Tuple = new class'XComLWTuple';
+	Tuple.Id = 'OverrideHitEffects';
+	Tuple.Data.Add(8);
+	Tuple.Data[0].kind = XComLWTVBool;
+	Tuple.Data[0].b = false; // Override default hit effects
+	Tuple.Data[1].kind = XComLWTVFloat;
+	Tuple.Data[1].f = Damage;
+	Tuple.Data[2].kind = XComLWTVObject;
+	Tuple.Data[2].o = InstigatedBy;
+	Tuple.Data[3].kind = XComLWTVVector;
+	Tuple.Data[3].v = HitLocation;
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
```diff
@@ -377,6 +377,66 @@ simulated function DamageTypeHitEffectContainer GetDamageTypeHitEffectContainer(
...
+///       inout name DamageTypeName
+///       inout vector Momentum
+///       inout bool bIsUnitRuptured
+///       inout EAbilityHitResult HitResult
+///     ]
+/// }
+/// EventSource: self (XComUnitPawn)
+/// NewGameState: no
+/// ```
+simulated function bool TriggerOnOverrideHitEffects(
```
Xymanek:
```suggestion
simulated private function bool TriggerOnOverrideHitEffects(
```

Musashi1584: We are removing protected and private all over the place. Should we really start adding more if that?

Xymanek: It's not a matter of the modifier itself, it's a matter of what it represents. Private means that something is to be used only internally and not by "outside" code.

When we are removing `private`, we are "approving" the use of something by outside code (which why such changes are accompanied by a use case).

Here, the new function, that is made to abstract away the internals of using XComLWTuple, is not supposed to be used by outside code, hence I'm suggesting (and it's already done in other cases) to mark it as `private`

Musashi1584: What about the use case of a mco that wants to change the event structure?

Xymanek: MCOs should not change event structure, that would break any other mod that uses same event. If an MCO wants to trigger the same event, then they can easily copy and paste the function.

However, if the CHL decided to add another input to the tuple (possible, allowed, happened already) then we suddenly broke an MCO which uses the trigger function. Hence the function (IMO) should not be usable by mods

Musashi1584: That was just an example. Maybe you want to override the function to trigger another event code or whatever. Point is we don’t want to add barriers for modding we are painstakingly removing elsewhere.

Xymanek: Ok. Suppose we do it the non-private way. Then we want add another argument to the tuple (expand context) - **this happened already**. How do we do it in the non-BC-breaking way?

Musashi1584: Not sure what you mean with another argument etc. I am just advocating to never use protected or private in new HL code

Musashi1584: I am even pro adding all files to the HL and remove private and protected from EVERYTHING

Xymanek: > Not sure what you mean with another argument

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/708/files#diff-3fd330f5f74f0a5f020891eb078df6b7

In this case, the event `AllowDarkEventRisk` received a new parameter (`Data[2]`). Because there was no public function that could be called by mods, this was easily done in a non-BC-breaking way. Compare that to the `CanAddItem_*` mess which every time more context was add, a new variant needed to be created **because there was a public function signature which CHL could not break** (due to BC)

Musashi1584: Private can be removed easily and is not enforced at runtime so you don’t know if that did not break anything

Xymanek: > Private can be removed easily and is not enforced at runtime so you don’t know or can’t at least be sure that did not break anything

It's not matter of "anything" - it's a matter of BC policy. If someone unprivates stuff that CHL made explicitly private and CHL later "breaks" that mod, then it's not CHL's fault - CHL took all precautions possible already **to protect the mods from themselves**. The fact that it's not enforced at runtime, for me, is argument **for** using private modifiers - the modder is able to bypass those restrictions, provided they accept the risk of their mod breaking at arbitrary point in future.

On discord you listed a big amount of unprivating you needed to do for a recent mod - fair. However, I will provide you with a counterexample - a case where CHL was able to implement its feature solely because the initial function was private: 

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/45b73702dd72c02748fa1a8e7138a55d117483a9/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_CovertAction.uc#L912-L913

In this way, CHL can act as a "tracker" of sorts of what mods can and cannot touch in XComGame (and other FXS-provided packages) - by both unprivating existing stuff and privating the new code it introduces **for its own use**.

If the CHL was to become abandoned tomorrow, then I would agree with you. However, the chance of that happening is basically non-existent

Xymanek: Additionally, `private` is just a form of marking stuff as not-for-use by mods. There are other forms available and in use, for example

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/45b73702dd72c02748fa1a8e7138a55d117483a9/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc#L62-L68

Would you also argue that that comment should be also be removed because it gives no guarantee that someone didn't use the marker/buffer slot(s) directly (by accident or on purpose)?

Musashi1584: further discussion on this matter should happen in one organised thread here: thttps://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/827/files

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
```diff
@@ -377,6 +377,66 @@ simulated function DamageTypeHitEffectContainer GetDamageTypeHitEffectContainer(
...
+/// HL-Docs: feature:OverrideHitEffects; issue:825; tags:tactical
+/// Allows listeners to override the default behavior of XComUnitPawn.PlayHitEffects
+/// This is especially useful for preventing the hardcoded templar fx for 
+/// eHit_Parry, eHit_Reflect and eHit_Deflect which play for any abilities that utilizing these hit results.
+///
+/// ```unrealscript
+/// EventID: OverrideHitEffects
+/// EventData: XComLWTuple {
+///     Data: [
+///       out bool OverrideHitEffect,
```
pledbrook, line 389: You should clarify in the body of the docs above what impact this value has (so modders don't need to look into the source code to find out). I know it can be inferred, but I can only be confident in the inference because I looked at the code.

A more subtle question is whether changing the event values below affects `PlayHitEffects()` even if `OverrideHitEffect` is `false`. One might assume the values are ignored in that case, but it is still an assumption.

Musashi1584: @pledbrook is that more clear now with that addition?
`/// If OverrideHitEffect is set to true the PlayHitEffects function will return early and the default behavior is ommited.`

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
```diff
@@ -381,19 +381,20 @@ simulated function DamageTypeHitEffectContainer GetDamageTypeHitEffectContainer(
 /// Allows listeners to override the default behavior of XComUnitPawn.PlayHitEffects
 /// This is especially useful for preventing the hardcoded templar fx for 
 /// eHit_Parry, eHit_Reflect and eHit_Deflect which play for any abilities that utilizing these hit results.
+/// If OverrideHitEffect is set to true the PlayHitEffects function will return early and the default behavior is ommited.
```
pledbrook, line 384: I would go with "ignored" or "bypassed" instead of "ommited", but it's fine. It's also "omitted" :wink:

PR #828, opened 2020-03-29
Fix stun action points persisting between missions (#824)

### LINKED ISSUE #824: Units still stunned on tactical end have no action points on next mission
The base game does not clear the stunned action points of any units that are still stunned when a tactical game ends, which leads to them having 0 action points at the start of the next mission they go on.

pledbrook:
At the end of tactical play units' stunned action points are cleared so that they don't persist through to the next mission and cause soldiers to start with no action points.

Fixes #824.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -2468,6 +2468,16 @@ function OnEndTacticalPlay(XComGameState NewGameState)
 		}
 	}
 
+	// Start Issue #824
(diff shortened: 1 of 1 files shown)


Iridar:
I've been asked to do code review for a bit. HLDocs and comments mention Issue #824, but #824 is the number of PR. I thought we are supposed to mark the # of the tracking issue, which is #828.

Currently doing game tests to confirm the bug is there in the basegame and is fixed by this PR.

Iridar:
Confirm the bug is present in the basegame, though affected soldiers eventually unbork themselves in as many turns as necessary, so it's not all that gamebreaking, but obviously still worth fixing.

Confirmed the bug is fixed by this PR. I think it's ready to be merged.

Iridar (approved):
Approved.

PR #745, opened 2020-01-23
Inline documentation

Iridar:
Inline documentation for Issues #281, #397, #457, #676.

PR #834, opened 2020-04-04
Fix docs build

robojumper:
#747 used an unknown tag, but there was no CI check because the PR was based on a commit that didn't have CI set up. This broke the master docs build.

#745 erronously had two `HL-Docs` in one block, which caused the script to emit an error but not terminate with a non-zero exit code, causing the check to succeed. All caught errors now (either deferredly or immediately) terminate the script with exit code 1.

### DIFF
--- a/.scripts/make_docs.py
+++ b/.scripts/make_docs.py
@@ -80,6 +80,7 @@ def make_doc_item(lines: List[str], file: str,
         k, v = pair.strip().split(':')
         if k in item:
             print("%s: error: %s: dupe key `%s`" % (sys.argv[0], file, k))
+            return None
         if k == 'feature' or k == 'ref':
             item[k] = v
         elif k == 'issue':
@@ -89,6 +90,7 @@ def make_doc_item(lines: List[str], file: str,
             item[k] = tags if tags != [''] else []
         else:
             print("%s: error: %s: unknown key `%s`" % (sys.argv[0], file, k))
+            return None
 
     ref = make_ref("\n".join(lines[1:]), file, span, item.get('issue'))
     if "ref" in item:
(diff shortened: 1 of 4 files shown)


.scripts/make_docs.py
```diff
@@ -300,8 +308,14 @@ def render_tag_page(tag: str, items: List[dict], outdir: str):
...
     fname = os.path.join(outdir, tag + ".md")
 
-    with open(fname, 'r'):
-        pass
+    try:
+        with open(fname, 'r'):
+            pass
+    except FileNotFoundError:
+        print(
+            "%s: error: file name %s not found (did you make up the tag `%s`?)"
```
pledbrook, line 316: Instead of "did you make up the tag `...`", how about "`...` is an unknown tag". It makes the problem a bit clearer to me (and I know non-natives often struggle with phrasal verbs).

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -550,9 +550,6 @@ static function bool CanWeaponApplyUpgrade(XComGameState_Item WeaponState, X2Wea
 /// NOTE: To create new sockets from script you need to unconst SocketName and BoneName in SkeletalMeshSocket
 /// </summary>
 /// HL-Docs: feature:DLCAppendWeaponSockets; issue:281; tags:pawns
-/// Allows mods to add, move and rescale sockets on the skeletal mesh of any weapon, which can be used to position visual weapon attachments.
-///	Duplicate sockets are not allowed. If you try to add a socket that already exists, the older socket will be removed.
-/// HL-Docs: feature:DLCAppendWeaponSockets; issue:281; tags:pawns
 /// Allows mods to add sockets to the skeletal mesh of any weapon, which can be used to position visual weapon attachments,
```
pledbrook, line 553: Which is more correct? "Add sockets" or "Add, move and rescale sockets"?

robojumper:
Thanks for the review. I believe I have adressed all the comments.

PR #840, opened 2020-04-12
KismetGameStateMatinee event

### LINKED ISSUE #837: Trigger event in SeqAct_PlayGameStateMatinee
Name suggestion: `KismetGameStateMatinee`

This will allow mods to insert their logic next to the matinee, by using `ELD_OnVisualizationBlockStarted`/`ELD_OnVisualizationBlockCompleted` or `PreBuildVisualizationFn`/`PostBuildVisualizationFn`

Xymanek:
Closes #837

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqAct_PlayGameStateMatinee.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqAct_PlayGameStateMatinee.uc
@@ -1,0 +1,113 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    SeqAct_PlayGameStateMatinee.uc
+//  AUTHOR:  David Burchanowski  --  2/10/2014
+//  PURPOSE: Plays a matinee in a gamestate safe manner
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+
+class SeqAct_PlayGameStateMatinee extends SequenceAction
+	implements(X2KismetSeqOpVisualizer)
+	native;
+
+// This lives here instead of in X2Action_PlayMatinee due to a cyclic dependency
+enum PostMatineeVisibility
+{
+	PostMatineeVisibility_Unchanged,
+	PostMatineeVisibility_Visible,
+	PostMatineeVisibility_Hidden,	
+};
+
+cpptext
+{
+	virtual void PostEditChangeProperty(FPropertyChangedEvent& PropertyChangedEvent);
+
+#if WITH_EDITOR
+	virtual FString GetDisplayTitle() const;
+#endif
+}
+
+// deprecated: hard link to the matinee we want to play. Does not work across maps. Temporarily here
+// so that existing kismet functionality is not broken
+var() editconst SeqAct_Interp Matinee;
+
+// obj comment on the matinee we want to play
+var() string MatineeComment;
+
+// Allows the LDs to specify a world space location for this matinee to play at
+var() string MatineeBaseActorTag;
+var() string MatineeBaseActorSocket;
+
+// If true, leaves game units hidden after playback completes, regardless of what state they were in before
+var() PostMatineeVisibility PostMatineeUnitVisibility;
+
+// finds the matinee with our tag in the loaded maps
+native private function SeqAct_Interp FindMatinee();
+
+function ModifyKismetGameState(out XComGameState GameState)
+{
+	// everything needs to wait for the matinee to complete
+	GameState.GetContext().SetVisualizationFence(true, 40.0f);
+}
+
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqAct_PlayGameStateMatinee.uc
```diff
@@ -49,6 +49,19 @@ function ModifyKismetGameState(out XComGameState GameState)
...
 	// everything needs to wait for the matinee to complete
 	GameState.GetContext().SetVisualizationFence(true, 40.0f);
+
+	/// HL-Docs: feature:KismetGameStateMatinee; issue:837; tags:events
+	/// Allow mods to insert their logic next to matinees triggered by mission kismet,
+	/// by using `ELD_OnVisualizationBlockStarted`/`ELD_OnVisualizationBlockCompleted`
+	/// or `PreBuildVisualizationFn`/`PostBuildVisualizationFn`.
+	/// 
+	/// The triggering `SeqAct_PlayGameStateMatinee` is not passed as event data/source in order to 
+	/// avoid any potential issues with replays. Instead, you can fetch it using the following code:
```
pledbrook, line 59: This should clarify that the event in fact passes *no* data or event source. Plus it should mention that a new game state is being passed. I think it's reasonable to include the standard doc structure for events, i.e.
```
///   EventID: KismetGameStateMatinee
///   EventData: none
///   EventSource: none
///   NewGameState: yes
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqAct_PlayGameStateMatinee.uc
```diff
@@ -0,0 +1,133 @@
...
+	// everything needs to wait for the matinee to complete
+	GameState.GetContext().SetVisualizationFence(true, 40.0f);
+
+	/// HL-Docs: feature:KismetGameStateMatinee; issue:837; tags:events
+	/// Allow mods to insert their logic next to matinees triggered by mission kismet,
+	/// by using `ELD_OnVisualizationBlockStarted`/`ELD_OnVisualizationBlockCompleted`
+	/// or `PreBuildVisualizationFn`/`PostBuildVisualizationFn`.
+	/// 
+	/// The triggering `SeqAct_PlayGameStateMatinee` is not passed as event data/source in order to 
+	/// avoid any potential issues with replays. Instead, you can fetch it using the following code:
```
robojumper, line 59: What are those issues with replays? Is this function even triggered from replays at all?

Xymanek: Good question. I was under the impression that the fully built game states are saved for the replays, while the viz stuff (including the triggered events) is (re-)calculated while the replay is playing.

As such, I believe that the function itself is not called, but the event is still dispatched

pledbrook (approved):
I'm fine with this now. @robojumper, are you OK with the docs? Not sure if you thought they were misleading or you just wanted clarification.

X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqAct_PlayGameStateMatinee.uc
robojumper: My understanding is that this event isn't saved (since `ModifyKismetGameState` is a game state function), but the `PreBuildVisualizationFn`/`PostBuildVisualizationFn` delegates are serialized anyway so mods generally don't need to pay attention to replays (except they need to ensure their delegates don't get garbage collected).

X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqAct_PlayGameStateMatinee.uc
```diff
@@ -0,0 +1,133 @@
...
+	/// 
+	/// ```unrealscript
+	/// XComGameStateContext_Kismet(GameState.GetContext()).FindSequenceOp()
+	/// ```
+	///
+	/// ```unrealscript
```
Iridar, line 65: Perhaps, this should be updated to new style event docs? LGTM otherwise.

X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqAct_PlayGameStateMatinee.uc
robojumper: I think the reference to "potential ssues with replays" is still misleading and should just be removed. I believe the SeqAct can be passed without problems but I wouldn't insist on that.

robojumper (approved):
While I personally don't believe there are any issues with passing the `SeqAct_PlayGameStateMatinee` and replays, I can't prove a negative with native code involved and since there's a documented and concise way to retrieve it, LGTM.

Issue #118, opened 2017-11-29
Add additional inventory slot types and facilitate implementing them


robojumper:
Currently, Inventory Slots are limited to this list:

```
enum EInventorySlot
{
	eInvSlot_Unknown,
	eInvSlot_Armor,
	eInvSlot_PrimaryWeapon,
	eInvSlot_SecondaryWeapon,
	eInvSlot_HeavyWeapon,
	eInvSlot_Utility,
	eInvSlot_Mission,
	eInvSlot_Backpack,
	eInvSlot_Loot,
	eInvSlot_GrenadePocket,
	eInvSlot_CombatSim,
	eInvSlot_AmmoPocket,
        eInvSlot_TertiaryWeapon,
        // ...
	eInvSlot_SeptenaryWeapon,
};
```

Where Mission, Backpack, Loot, Tertiary-Septernary are not available to the user. Adding a mechanism that allows mods to add more slots would certainly add to the moddability, but would also be a rather tricky effort.

### 1. This Enum

Enums are bytes (0-255), where 255 is the implicit `longestCommonPrefix_MAX` value. They are compiled directly into the byte code as constant tokens, and not sensitive to any changes.

Because of the way the Game State and Template system works, it is necessary that any given item type NEVER changes its corresponding slot byte value, be it over the course of the campaign, across different save files, across different machines, independent of other mods being present or not, ...

As such, adding a new slot requires adding a new enum value that does not conflict with any other mods. 
Unfortunately (as far as I know) Byte Properties linked to an enum are serialized using their Enum-Name, setting it to None when saving and _MAX when loading if the entry doesn't exist. This requires us to add new entries to the Highlander.

We have 256 - 1 - 17 = 238 free slots, so that should not be a problem if we are minimally considerate about what types of slots we add.

#### Native Code

[...]

PR #849, opened 2020-04-20
Added four Inventory Slots for Sparks (Issues #118 and #137)

Iridar:
Added inventory slots:

```
eInvSlot_SparkRocket1
eInvSlot_SparkRocket2
eInvSlot_SparkRocket3
eInvSlot_SparkRocket4
```

They will be used by my upcoming SPARK Weapons mod. The slots will be templated and added to the SPARKs and MEC Troopers when an Ordnance Launcher is equipped. I could potentially just add utility slots and control what goes into them via CanAddItemToInventory, but templated slots offer more control and are more compatible with mods that would want SPARKs to have Utility slots for things like SPARK Launcher by RM or ammo.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
@@ -104,10 +104,8 @@ enum EInventorySlot
 	eInvSlot_Pistol,
 	eInvSlot_Wings,
 	eInvSlot_ExtraBackpack,
-	eInvSlot_SparkRocket1,
-	eInvSlot_SparkRocket2,
-	eInvSlot_SparkRocket3,
-	eInvSlot_SparkRocket4,
+	eInvSlot_SparkGrenadePocket,
+	eInvSlot_AuxiliaryWeapon,
 
 	// Marker slot, don't use
 	eInvSlot_END_TEMPLATED_SLOTS,
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
@@ -104,6 +104,10 @@ enum EInventorySlot
 	eInvSlot_Pistol,
 	eInvSlot_Wings,
 	eInvSlot_ExtraBackpack,
+	eInvSlot_SparkRocket1,
+	eInvSlot_SparkRocket2,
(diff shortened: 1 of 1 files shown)


Iridar:
Something so simple doesn't require a review, methinks, but if it matters I did test the slots.

Xymanek:
Why not a single multi-item slot?

Iridar:
I don't know how to work with them, and even if I learned how, it might mean redoing parts of the already overcomplicated rocket launchers mod.

Xymanek:
Well, I'm 👎 on the current approach cuz:

* It wastes (large but) limited space of slots available
* If anyone (including you) ever decides to have the unit equip 5 rockets at the same time, the universe will implode

> I don't know how to work with them

You set the following:

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/e760415df9bc3b64e7b28b7c8c0b35cb5794550c/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHItemSlot.uc#L56-L61

And then work with it like with the utility slot (which is also multi-item)

Iridar:
I don't know how to work with utility slot either... okay, I will give it a shot, but I will still need at least two slots. 

> If anyone (including you) ever decides to have the unit equip 5 rockets at the same time, the universe will implode

I'm 100% fine with that limitation. it's limited by soldier sockets and code anyway.

Iridar:
Looks like a multi item slot can work for me in this case, even though it will require a more complicated logic to set up, so I cut down the number of slots I'm requesting from 4 to 2.

BTW, should push come to shove, and we start running out of inventory slots (which seems highly unlikely), it's always possible to make a "unified slot repository" sort of mod, that would provide slots  to units based on arbitrary conditions. Veehementia's Dedicated Pistol Slot is already very close to that. 

> If anyone (including you) ever decides to have the unit equip 5 rockets at the same time, the universe will implode

To revisit this, universe will not implode, extra rockets just won't be displayed on the soldier's model or during the firing animation. The slots from the Rocket Launcher mod are used to store the PRES-equippable cosmetic copy of the rocket, as the rocket itself is carried in regular utility slots that are not PRES-equippable. Should this ever change, there will be no need for the Rocket Launchers mod to take *any* slots.

Xymanek:
> the rocket itself is carried in regular utility slots that are not PRES-equippable. Should this ever change, there will be no need for the Rocket Launchers mod to take any slots.

#889 😏

Iridar:
> > the rocket itself is carried in regular utility slots that are not PRES-equippable. Should this ever change, there will be no need for the Rocket Launchers mod to take any slots.
> 
> #889 😏

I'll happily relinquish the slots in #137 once #889 is in the stable Highlander.

PR #853, opened 2020-05-04
Add XComTacticalInput.uc for modification

pledbrook:
I'd like to add a hook to override the default camera rotation angle and a separate hook for overriding the ability that gets activated by the R3 controller button.


robojumper (approved):
Easy to approve since this is just an addition for the other two linked PRs (#856 and #857).

PR #856, opened 2020-05-06
Override camera rotation angle

### LINKED ISSUE #854: Allow users and mods to override the camera rotation angle
Most of the camera rotation mods use an MCO for `XComTacticalInput`, but this doesn't sit well with any other mod that wants to override that class for any other reason. And why force players to install a mod if they just want to change the default rotation angle?

pledbrook:
Depends on #853. Adds both an event and a configuration variable that allows mods and/or players to override the default camera rotation angle of 90°.

Note that this passes the ActionMask to the listeners, but that may not be particularly useful when simply overriding the rotation angle. Another approach would be to override the key press/key release behaviour as well. Not sure which is better at this point.

Resolves #854

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
@@ -178,6 +178,9 @@ var config array<name> ClassesAllowPsiPCS;
 // Variable for Issue #724
 var config array<name> ValidReserveAPForUnitFlag;
 
+// Variable for Issue #854
+var config float CameraRotationAngle;
+
 // Start Issue #885
 enum EHLDelegateReturn
 {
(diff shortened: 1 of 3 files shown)


robojumper (changes requested):
As far as I can tell, this change is missing here:

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/3c6eba4a22980af2eada617d0c12cf67c45e1ac9/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_AbilityContainer.uc#L281-L290

Additionally, I'd like a `compatibility` note and tag because several mods (mosly camera mods) override `XComTacticalInput` and can effectively revert the HL changes.

pledbrook:
I've rebased against `master`, updated those `YawCamera()` calls you pointed out and added a compatibility note. This PR should be squash merged, even though I force pushed.

robojumper (approved):

PR #857, opened 2020-05-06
Override controller r3

### LINKED ISSUE #855: Allow mods to override the behaviour of controller R3 button
The R3 button is hard coded to activate the 'PlaceEvacZone' ability, but LWOTC has a different ability that needs activating: 'PlaceDelayedEvacZone'. We should add an event or other hook that will allow mods to rebind the R3 button.

pledbrook:
Depends on #853.

Adds an event that allows mods to override the behaviour of the controller's R3 button, either by simply changing the name of the ability that is activated or by completing overriding the behaviour.

Resolves #855


pledbrook:
I've changed the approach for this pull request: rather than an event, I now use a config var that allows mods to change the name of the ability used for the R3 button. It's simpler, but less flexible. On the other hand, I'm pretty sure it won't interfere with the proposed changes for a command abilities menu/button bar.

pledbrook:
Also note that the previous implementation didn't work as the ability name needed overriding in several different places, some of which I missed.

X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
```diff
@@ -170,6 +170,16 @@ var config int NoiseAlertSoundRange;
 var config array<name> AdditionalAIBTActionPointTypes;
 // End Issue #510
 
+// Start Issue #855
+/// HL-Docs: feature:PlaceEvacZoneAbilityName; issue:855; tags:tactical
+/// Config variable (of type name) that allows mods to override the name of
+/// the ability used for placing/throwing evac zones in tactical missions.
+/// This is because the base game hard codes the ability name used for the
+/// controller's R3 button, which is bad if a mod (like LWOTC) replaces
+/// PlaceEvacZone with a different ability.
```
robojumper, line 179:
```suggestion
/// PlaceEvacZone with a different ability.
///
/// Note that this affects only the input system. The game rules themselves
/// are not aware that the ability changed, so you must handle ability
/// availability, evac zone destruction etc. yourself!
```

pledbrook:
I rebased against `master` and added what I hope is a clarifying comment about what the config var can and can't do.

robojumper (approved):

robojumper:
This now has a merge conflict, presumably due to #856.

PR #858, opened 2020-05-16
Bring back the DLC2 component

### LINKED ISSUE #803: Revisit DLC HL Components
#772 moved the DLC2 HL from component approach to the main CHL mod. Such approach greatly simplifies both developer tooling and the need for player to manually resolve which pieces of CHL they need. The idea was that if nothing specifies `+NonNativePackages=DLC_2` in config, then the package will not be loaded. This assumption is correct, and properly holds under "only CHL should behave same as base game" mantra for both DLC owners and non-owners.

In practice, it turns out, there are more ways to load a script package. If a mod script package references classes from DLC_2.u (even in optional functions), then there is a build-time dependency which the engine dutifully attempts to load when loading the dependency, just like with assets.  

image

Normally, this would silently fail (and crash if the functions with DLC references are called) since there is no `DLC_2.u` to load anywhere, but with #772 merged suddenly there is one. The result is that non-DLC-owners suddenly get the DLC code active, without the assets, localization and even a "formal" DLC identifier, so they get the (empty, cuz no localization) popup about installing the DLC into the existing campaign when loading a save:
image

Clearly this situation is unacceptable. Furthermore, it appears that reverting to the components approach is the only way to resolve this, despite all the associated downsides.

When implementing the components, I see two main areas to be considered:

1) Developer tooling
2) Player-perceptive dependency mess

Here are my thoughts:

1) Developer tooling:
  * Workshop IDs for main and beta versions
  * Build tasks for DLC2HL (debug/release/workshop stable release)
  * Build script for DLC2HL to include the main CHL on top of SrcOrig
  * Build tasks for main CHL to include DLC2HL for version checker
  * (CHL-dependant mods) Build script to include DLC2HL source into SrcOrig
2) Player-perceptive dependency mess
  * **Require** the DLC2HL when both main CHL and DLC2 are installed
    * This way, individual mods don't need to include "also requires DLC2HL if you own DLC2" and players are guaranteed to receive the fixes made in DLC2HL
  * **Require** the main CHL when the DLC2HL is installed (otherwise crash due to use of `XComLWTuple` and potentially other main CHL things)

Also I suggest merging #790 and #793 before reworking to components approach (to prevent the need to rebase and rework theses PRs)

cc @robojumper what do you think?
cc @cannonfodder1 one of affected non-DLC2-owners

This needs to be fixed before 1.20 release to prevent BC issues

Xymanek:
Closes #803 (+ context)

Possible things to do (feedback please):

* Popup when the main CHL is loaded without the DLC2HL (but the DLC2 is active) saying "since CHL vX.Y.Z you also need DLC2HL"
* Expand the troubleshooting wiki page
* Standardize on naming - "main CHL", "DLC2HL", etc

cc @robojumper @pledbrook


Components/DLC2CommunityHighlander/.scripts/build.ps1
```diff
@@ -0,0 +1,342 @@
+Param(
```
robojumper, line 1: I'm not so keen on duplicating the entire build script. In particular, we duplicate the whole `Invoke-Make`, compiler invocations, `.x2proj` handling and so on and it'd be good to avoid that. Is it possible to customize one script to work for both the main HL and generic mods? Or extract the common parts into a common script that is included in both?

Other than that, LGTM.

Xymanek: The problem is that there are some parts that are very specific (e.g. cooking, workshop ID swap). I would love to build a more generic script that can handle all scenarios; unfortunately, I do not have time for that right now, especially considering that the current situation works

Xymanek:
When the DLC2 Component is missing, the below gets shown:

image

image

```
[0022.25] X2WOTCCommunityHighlander: Companion script package loaded
[0022.28] X2WOTCCommunityHighlander: Components:
[0022.28] X2WOTCCommunityHighlander:   X2WOTCCommunityHighlander -- v1.20.0 (9a52d6)
[0022.28] X2WOTCCommunityHighlander:   Core -- v1.20.0 (9a52d6)
[0022.28] X2WOTCCommunityHighlander:   Engine -- v1.20.0 (9a52d6)
[0022.28] X2WOTCCommunityHighlander:   XComGame -- v1.20.0 (9a52d6)
[0022.28] X2WOTCCommunityHighlander:   DLC_2 -- NOT LOADED - REQUIRED
[0022.28] X2WOTCCommunityHighlander: Required packages were not loaded. Please consult https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/wiki/Troubleshooting for further information
```

Components/DLC2CommunityHighlander/.scripts/build.ps1
robojumper: From discussion on Discord -- even if a single parametrized script isn't possible, we should move the common parts to a different, [sourced](https://stackoverflow.com/a/6217858) file.

Xymanek:
Superseded by #999

PR #860, opened 2020-05-16
Black Market changes

### LINKED ISSUE #859: Open BlackMarket data up to mods
#473 discusses buy prices and sell data being manipulated by mods, but the list of items the black market buys from you are reset later by `UIBlackMarket_Sell.PopulateData()`. Thus, any new states added to the BuyPrices will be removed at that point.

`XComGameState_BlackMarket.UpdateBuyPrices()` could fire an event `BlackMarketBuyPricesUpdated` that allows mods to add new buy prices without the threat of getting removed.

`UIBlackMarket_Sell.PopulateItemCard()` could fire an event `BlackMarketItemCardPopulated` with a LWTuple containing the display data, which mods could handle before display. This is important for mods that handle state data because the vanilla `PopulateItemCard()` ignores state data in favor of template data.

For the record, `UIBlackMarket_Sell.OnConfirmButtonClicked()` already fires the event `BlackMarketGoodsSold` with the new GameState, so mods are already able to handle items that are sold by retrieving that GameState and diffing the XComHQ.Inventory against the one from the most recent history frame. However, it may be simpler in the long run to compile a list of sold items and include them in the event, though this is not strictly necessary so I'd like opinions.

GingerAvalanche:
Implements #859 changes

Closes #859 
(Testing to see if this does what I think it does)

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_Sell.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_Sell.uc
@@ -1,0 +1,415 @@
+//---------------------------------------------------------------------------------------
+//  *********   FIRAXIS SOURCE CODE   ******************
+//  FILE:    UIBlackMarket_Sell.uc
+//  AUTHOR:  Samuel Batista
+//  PURPOSE: Screen where players can exchange items for supplies.
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+
+class UIBlackMarket_Sell extends UIScreen;
+
+var int SaleAmount;
+//var UIText TotalValue;
+var StateObjectReference BlackMarketReference;
+var UIButton ConfirmButton;
+
+var localized string m_strConfirmButtonLabel;
+var localized string m_strInventoryLabel;
+var localized string m_strSellLabel;
+var localized string m_strTotalLabel;
+var localized string m_strCostLabel;
+var localized string m_strInterestedLabel;
+var localized string m_strQuantityLabel;
+var localized String	m_strSellConfirmTitle;
+var localized String	m_strSellConfirmText;
+var localized string m_strAddRemoveItems;
+
+var UIList List;
+var UIPanel ListBG;
+
+var XComGameStateHistory History;
+var XComGameState_HeadquartersXCom XComHQ;
+
+
+simulated function InitScreen(XComPlayerController InitController, UIMovie InitMovie, optional name InitName)
+{
+	super.InitScreen(InitController, InitMovie, InitName);
+
+	History = `XCOMHISTORY;
+	XComHQ = class'UIUtilities_Strategy'.static.GetXComHQ();
+
+	BuildScreen();
+	MC.FunctionVoid("AnimateIn");
+	Show();
+}
+
+simulated function BuildScreen()
+{
+	History = `XCOMHISTORY;
+	XComHQ = class'UIUtilities_Strategy'.static.GetXComHQ();
+
+	ListBG = Spawn(class'UIPanel', self);
+	ListBG.InitPanel('InventoryListBG');
+	ListBG.Show();
+
+	List = Spawn(class'UIList', self).InitList('inventoryListMC');
+	List.bStickyHighlight = true;
+	List.OnSelectionChanged = SelectedItemChanged;
+
+	Navigator.SetSelected(List);
+
+	// send mouse scroll events to the list
+	ListBG.ProcessMouseEvents(List.OnChildMouseEvent);
+
+	UpdateNavHelp();
+
+	MC.BeginFunctionOp("SetGreeble");
+	MC.QueueString(class'UIAlert'.default.m_strBlackMarketFooterLeft);
+	MC.QueueString(class'UIAlert'.default.m_strBlackMarketFooterRight);
+	MC.QueueString(class'UIAlert'.default.m_strBlackMarketLogoString);
+	MC.EndOp();
+
+	//---------------------
+	
+	// Move and resize list to accommodate label
+	List.SetHeight(class'UIBlackMarket_SellItem'.default.Height * 16);
+
+	UpdateSellInfo();
+
+	ConfirmButton = Spawn(class'UIButton', self).InitButton('ConfirmButton', m_strConfirmButtonLabel, OnConfirmButtonClicked, eUIButtonStyle_HOTLINK_BUTTON);
+	ConfirmButton.SetGamepadIcon(class'UIUtilities_Input'.static.GetGamepadIconPrefix() $ class'UIUtilities_Input'.const.ICON_X_SQUARE);
+
+	PopulateData();
+	UpdateTotalValue();
+	List.SetSelectedIndex(0, true);
+}
+
+simulated function PopulateData()
+{
+	local XComGameState NewGameState;
+	local BlackMarketItemPrice Item;
+	local XComGameState_Item InventoryItem;
+	local array<BlackMarketItemPrice> Items;
+	local XComGameState_BlackMarket BlackMarketState;
+	local UIBlackMarket_SellItem ListItem;
+	local BlackMarketItemPrice PrevItem;
+
+	ListItem = UIBlackMarket_SellItem(List.GetSelectedItem());
+	if(ListItem != None)
+		PrevItem = ListItem.ItemPrice;
+
+	// override behavior in child classes
+	List.ClearItems();
+
+	NewGameState = class'XComGameStateContext_ChangeContainer'.static.CreateChangeState("Update Black Market Price List");
+	BlackMarketState = XComGameState_BlackMarket(`XCOMHISTORY.GetGameStateForObjectID(BlackMarketReference.ObjectID));
+
+	// Update Black Markets prices if we need to
+	BlackMarketState = XComGameState_BlackMarket(NewGameState.ModifyStateObject(class'XComGameState_BlackMarket', BlackMarketState.ObjectID));
+	if(BlackMarketState.UpdateBuyPrices())
+	{
+		`XCOMGAME.GameRuleset.SubmitGameState(NewGameState);
+	}
+	else
+	{
+		NewGameState.PurgeGameStateForObjectID(BlackMarketState.ObjectID);
+		`XCOMHISTORY.CleanupPendingGameState(NewGameState);
+	}
+
+	BlackMarketState = XComGameState_BlackMarket(`XCOMHISTORY.GetGameStateForObjectID(BlackMarketReference.ObjectID));
+	Items = BlackMarketState.BuyPrices;
+	Items.Sort(SortByInterest);
+	
+	foreach Items(Item)
+	{
+		// Don't display if none in your inventory to sell
+		InventoryItem = XComGameState_Item(`XCOMHISTORY.GetGameStateForObjectID(Item.ItemRef.ObjectID));
+		if( InventoryItem.Quantity > 0 )
+		{
+			Spawn(class'UIBlackMarket_SellItem', List.itemContainer).InitListItem(Item);
+			if(Item == PrevItem)
+				List.SetSelectedIndex(List.GetItemCount() - 1);
+		}
+	}
+
+	if(List.ItemCount > 0)
+	{
+		ListItem = UIBlackMarket_SellItem(List.GetItem(0));
+		PopulateItemCard(ListItem.ItemTemplate, ListItem.ItemRef, string(ListItem.Price));
+		if(List.SelectedIndex < 0)
+			List.SetSelectedIndex(0);
+	}
+	else
+	{
+		ClearItemCard();
+	}
+
+	if (List.ItemCount <= 0)
+		CloseScreen(); //bsg-crobinson (5.22.17): Force player out if they run out of items to sell
+}
+
+function int SortByInterest(BlackMarketItemPrice BuyPriceA, BlackMarketItemPrice BuyPriceB)
+{
+	local XComGameState_BlackMarket BlackMarketState;
+	local XComGameState_Item ItemA, ItemB;
+
+	History = `XCOMHISTORY;
+	BlackMarketState = XComGameState_BlackMarket(History.GetGameStateForObjectID(BlackMarketReference.ObjectID));
+	ItemA = XComGameState_Item(History.GetGameStateForObjectID(BuyPriceA.ItemRef.ObjectID));
+	ItemB = XComGameState_Item(History.GetGameStateForObjectID(BuyPriceB.ItemRef.ObjectID));
+
+	if(BlackMarketState.InterestTemplates.Find(ItemB.GetMyTemplateName()) != INDEX_NONE &&
+	   BlackMarketState.InterestTemplates.Find(ItemA.GetMyTemplateName()) == INDEX_NONE)
+	{
+		return -1;
+	}
+	
+	return 0;
+}
+
+simulated function UpdateTotalValue(optional int Delta = 0)
+{
+	SaleAmount += Delta;
+
+	ConfirmButton.SetDisabled(SaleAmount == 0);
+
+	MC.BeginFunctionOp("UpdateSellTotal");
+	MC.QueueString(class'UIUtilities_Strategy'.default.m_strCreditsPrefix  $ string(SaleAmount));
+	MC.EndOp();
+	UpdateNavHelp();
+}
+
+simulated function OnConfirmButtonClicked(UIButton Button)
+{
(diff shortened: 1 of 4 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_Sell.uc
```diff
@@ -0,0 +1,456 @@
...
+/// ** strImage - the image to display for the item
+/// ** strTitle - the title of the item card (typically the item's name)
+/// ** m_strCostLabel - the localized string for the English word "COST"
+/// ** ItemPrice - the string denoting the item's price (if you want to edit prices,
+///        you should use BlackMarketGoodsReset or BlackMarketBuyPricesUpdated)
+/// ** strInterest - the string denoting whether or not the black market is interested.
+///        By default, the localized string for the English words "VERY INTERESTED" or ""
+/// ** strSummary - a summary of the item. By default, the item's BriefSummary
+/// * An int, the item's ObjectID
+/// * An object, the item's X2ItemTemplate
```
robojumper, line 282: We don't really have a strict schema for Tuple contents documentation but this will be unreadable in the rendered document.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_BlackMarket.uc
```diff
@@ -495,6 +495,17 @@ function UpdateForSaleItemDiscount()
 }
 
 //---------------------------------------------------------------------------------------
+// issue #859
+/// HL-Docs: feature:BlackMarketBuyPricesUpdated; issue:859; tags:strategy,events
+/// This is an event that allows mods to manipulate black market BuyPrices directly,
+/// including adding new BuyPrices that won't be reset.
+///
+/// EventData is an XComLWTuple with two values:
+/// * An array of ints, the ObjectIDs for states with buy prices
```
robojumper, line 504: Same here, this at least needs a newline before the bullet point list.

robojumper (changes requested):
Looks sound to me, but docs will not render correctly.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_Sell.uc
GingerAvalanche: Markup’s always gotta look different on every platform, doesn’t it? :P

Is there a way to see what the docs will look like before I upload? I’m used to single asterisk being bullet point and double asterisk being indented bullet point. Not really sure how to resolve that if I can’t see what I’m doing.

robojumper: Locally: See https://x2communitycore.github.io/X2WOTCCommunityHighlander/#documentation-for-the-documentation-tool

You can also push another commit to re-run the checks, then download the artifact by clicking the green checkmark next to the commit, choosing "details", clicking "Artifacts" and downloading the HTML bundle. Unless you run it on a webserver, the stylesheets will be missing, but the HTML markup will be visible.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_BlackMarket.uc
GingerAvalanche: Resolved.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_Sell.uc
GingerAvalanche: I wish the docs would support nested bullet points, though I realize that would be very difficult, given that spaces at the beginning of newlines would have ambiguous purpose, between differently-indented bullet points and just plain old clean comment formatting. Doesn't stop me from hating it, though.

GingerAvalanche:
Original commit split into file-add and changes, subsequent commits squashed into first edit commit for readability.

ready-for-review tag, can't change, blah blah blah.

GingerAvalanche:
Realized I left an unedited file out of the unedited file commit split, so I went back and redid that one. Now it should be good to go.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_Sell.uc
```diff
@@ -273,17 +301,35 @@ simulated function PopulateItemCard(X2ItemTemplate ItemTemplate, StateObjectRefe
...
+	Tuple.Data[0].as.AddItem(strImage);
+	Tuple.Data[0].as.AddItem(strTitle);
+	Tuple.Data[0].as.AddItem(m_strCostLabel);
+	Tuple.Data[0].as.AddItem(ItemPrice);
+	Tuple.Data[0].as.AddItem(strInterest);
+	Tuple.Data[0].as.AddItem(ItemTemplate.GetItemBriefSummary(ItemRef.ObjectID));
+	Tuple.Data[1].kind = XComLWTVInt;
+	Tuple.Data[1].i = ItemRef.ObjectID;
+	Tuple.Data[2].kind = XComLWTVObject;
+	Tuple.Data[2].o = ItemTemplate;
```
pledbrook, line 316: There's a significant mismatch between the docs and this code. The docs imply this:
```
Tuple.Data[0].kind = XComLWTVString;
Tuple.Data[0].s = strImage;
Tuple.Data[1].kind = XComLWTVString;
Tuple.Data[1].s = strTitle;
Tuple.Data[2].s = m_strCostLabel;
Tuple.Data[3].s = ItemPrice;
Tuple.Data[4].s = strInterest;
Tuple.Data[5].s = ItemTemplate.GetItemBriefSummary(ItemRef.ObjectID);
```
Note that `Item.ObjectID` and `ItemTemplate` are not documented.

In addition, `.as` is not intended just to group different properties of the same type together. It's more for properties that *are* arrays of strings.

One last thing: do you need to pass the template? Can't a listener get that from the item state object if it needs it?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_BlackMarket.uc
```diff
@@ -531,6 +545,28 @@ function bool UpdateBuyPrices()
...
+	Tuple.Data[0].kind = XComLWTVArrayInts;
+	Tuple.Data[1].kind = XComLWTVArrayInts;
+	
+	foreach BuyPrices(BuyPrice)
+	{
+		Tuple.Data[0].ai.AddItem(BuyPrice.ItemRef.ObjectID);
+		Tuple.Data[1].ai.AddItem(BuyPrice.Price);
+	}
+
+	`XEVENTMGR.TriggerEvent('BlackMarketBuyPricesUpdated', Tuple, History);
```
pledbrook, line 559: There is no need to pass `XComGameStateHistory` as the event source. This can be retrieved on demand from anywhere. Plus, is the history really the event source logically speaking? It might make sense to pass the Black Market state object as the event source.

pledbrook: Also, as before, I really dislike polluting base game code with plumbing like this. I would prefer to see it factored out into a separate function.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_BlackMarket.uc
```diff
@@ -495,6 +495,18 @@ function UpdateForSaleItemDiscount()
...
 //---------------------------------------------------------------------------------------
+// issue #859
+/// HL-Docs: feature:BlackMarketBuyPricesUpdated; issue:859; tags:strategy,events
+/// This is an event that allows mods to manipulate black market BuyPrices directly,
+/// including adding new BuyPrices that won't be reset.
+///
+/// EventData is an XComLWTuple with two values:
+/// 
+/// * An array of ints, the ObjectIDs for states with buy prices
+/// * An array of ints, the prices
```
pledbrook, line 506: This should follow the standard format for event definition documentation, just as you used for the other event.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_SellItem.uc
```diff
@@ -84,7 +84,21 @@ simulated function PopulateData(BlackMarketItemPrice BuyPrice)
 		Price = Min(Price, ItemCost);
 	}
 
-	ItemName = class'UIUtilities_Text'.static.GetColoredText(ItemTemplate.GetItemFriendlyName(), eUIState_Normal, 24);
+	//issue #859 changes
+	Tuple.Data.Add(3)
+	Tuple.Data[0].kind = XComLWTVArrayStrings;
+	Tuple.Data[0].as.Add(1);
+	Tuple.Data[0].as.AddItem(ItemTemplate.GetItemFriendlyName());
+	Tuple.Data[0].as.Add(4);
```
pledbrook, line 92: This is very confusing. You scale the array to accept one item. You then use `AddItem()`, which automatically scales the dynamic array to fit in the new item, and then you scale it to 4 items without adding any more.

As before, I really don't think you should be using `.as` here. Just a normal string tuple field works fine and is simpler.

pledbrook (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_Sell.uc
```diff
@@ -260,9 +260,37 @@ simulated function SelectedItemChanged(UIList ContainerList, int ItemIndex)
...
+/// black market. As PopulateItemCard() normally ignores state data in favor of
+/// template data, mods that add items for black market sale whose state data is
+/// important should use this event to make sure the item cards use the correct
+/// data.
+///
+/// ```unrealscript
+/// EventID: BlackMarketItemCardPopulated
+/// EventData: XComLWTuple {
+///     Data: [
+///       inout string strImage - the image to display for the item
```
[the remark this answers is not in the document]
robojumper: The way #908 is implemented, this won't parse, both because that PR switches from `XComLWTuple { Data: [ ... ] }` to simply `[ ... ]` and because comments aren't supported. I'm willing to accept this for now though and rebase #908 after this is merged and change this in that PR.

Feedback about a comment column being useful hereby noted.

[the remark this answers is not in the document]
GingerAvalanche: I settled on this because I saw it was done this way in another, existing doc and I was tired after committing about a half-dozen changes just to see the docs fail to render properly. There was no way to do indented bullet points, so I just gave up. If there's another way to handle this that will actually render properly, I'll gladly do it that way, instead.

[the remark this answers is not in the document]
robojumper: This is fine this way, no need to hold up the PR because another, not yet merged PR changes the format again.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_Sell.uc
GingerAvalanche: Docs mismatch is wholly due to the fact that I was using an unfamiliar format to document, because, as mentioned on the doc comment, the docs don't support indented bullet points. I'll go back and try to understand that format in another way, or see if I can find a way to make the readable format render properly, but trying to explain nested collections with our doc format has proven incredibly frustrating so far.

That being said, it sounds like the tuple format I inadvertently put into the docs would be the preferred format. In which case, I can simply refactor the code to match the docs, instead of the other way around.

No. I do not need to pass the template.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_Sell.uc
```diff
@@ -273,17 +301,35 @@ simulated function PopulateItemCard(X2ItemTemplate ItemTemplate, StateObjectRefe
...
-	MC.QueueString(strImage);
-	MC.QueueString(strTitle);
-	MC.QueueString(m_strCostLabel);
-	MC.QueueString(ItemPrice);
-	MC.QueueString(strInterest);
+	MC.QueueString(Tuple.Data[0].as[0]);
+	MC.QueueString(Tuple.Data[0].as[1]);
+	MC.QueueString(Tuple.Data[0].as[2]);
+	MC.QueueString(Tuple.Data[0].as[3]);
+	MC.QueueString(Tuple.Data[0].as[4]);
```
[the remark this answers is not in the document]
GingerAvalanche: Can you give me an example of "as done elsewhere"? I followed the examples of previous issues when I set it up in this manner, so I'm not sure where I should be looking.

Also, I have no idea how I'm supposed to set this up to use named variables without manually assigning return values from the separate function to those named variables, which, wouldn't that just be replacing the plumbing with slightly less plumbing? Maybe I'll understand once I can see what you mean.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_BlackMarket.uc
GingerAvalanche: The last review I received on the other event stated there was no standard format. Is there a place where these formats are documented?

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_SellItem.uc
GingerAvalanche: I did this because it's firing the same event as `UIBlackMarket_Sell:PopulateItemCard()` so that one listener can respond to events from both functions, as what data is used in this function is also used in that one. It cleans up the code on the mod end of things. I leave those indices empty because this function doesn't care about the data that the other function puts into those indices.

I didn't feel the need to include this function in the documentation because this function's purpose is incidental to `PopulateItemCard()`'s - an author would expect that when that function performs a job, this one will perform its job in tandem. This firing of the same event is simply to uphold that expectation.

This also means that those empty strings would/will still be necessary if I turn the `.as` into five `.s`'s.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_Sell.uc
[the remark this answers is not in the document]
pledbrook: Here's one simple example: https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAdventOperations.uc#L252

And another one here: https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIResistanceReport.uc#L162

Hope that helps. If you have any more questions, let me know. A large chunk of these functions begin `TriggerOverride`, so you could also search for that in the code base to see more examples.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_BlackMarket.uc
pledbrook: Well, there's a pre-#908 standard that the doc generation uses, and the #908 standard. I think the idea is to go back and change the old style to the new in previous commits/pull requests. I think it might help that transition if you use the same form as you did earlier in this PR for the `BlackMarketItemCardPopulated` event.

Alternatively, just add the line
```
//(/) FIXME: Update to new event specification (#908)
```
to the comment block somewhere so that it can be identified easily.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_SellItem.uc
pledbrook: OK, thanks for the explanation. I'll need to think some more on this. I'm not really keen on the idea of padding event data just so the same listener can be used for both events. One approach would be to simply make ItemFriendlyName the first element of the `Data` array and then listeners can just not bother with the other elements if `Data` only has length 1.

Anyway, let me think about it.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIBlackMarket_Sell.uc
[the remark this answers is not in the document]
GingerAvalanche: Aha! Many thanks, I've gotten so used to Python recently that I forgot you can specify primitive-type arguments as `out` and it'll pass a reference to the argument instead of the literal.

Iridar:
Per discussion on discord, leaving this PR for @pledbrook to handle.

Iridar:
It's dead, Jim.

PR #863, opened 2020-05-18
UIStaffSlot fixes

### LINKED ISSUE #861: UIStaffSlot loses visual focus after personnel selection - with solution
**NOTE** : I am nearly certain this is a WotC bug which, in theory, applies to mouse & keyboard and controller users, but in practise, only applies to controller users. Additionally, this problem only pertains to conventional UIStaffSlots; subclasses of UIStaffSlot do not share this problem.

**PROBLEM** : When a controller user clicks a normal UIStaffSlot, a drop down menu is created, if necessary, and shown. At this point in time, both the slot, and the top menu list item, are highlighted. Now, part of the drop down menu setup code is :
m_kPersonnelDropDown.onSelectedDelegate = StaffSlot.OnPersonnelSelected; in other words, once the drop down menu is clicked, call the staff slot's OnPersonnelSelected() function. 

The problem with OnPersonnelSelected() is that its 1st call is to : OnLoseFocus(). So, once you select someone for the slot, including the empty slot possibility, the drop down menu hides itself and the slot becomes unhighlighted; the staff slot is no longer focused. Or so we think ! In reality, navigator selection has not been modified, so the slot actually remains focused; consequently, it should have remained highlighted.

This is very easy to verify with a controller.
1.] Select a normal UIStaffSlot
2.] Click on the slot with the A button, at which point a drop down menu shows up. Remember that the staff slot remains highlighted.
3.] Select an individual for the slot; the drop down menu hides and the slot loses its highlighting.
4.] Click the A button one more time. The slot's drop down menu suddenly shows up again !! This shouldn't happen if the slot has lost focus since the A button should be sent to the selected UI item. The same things happens if this is the 1st slot in the staff container, the 2nd, the 3rd ... etc.

**THE SOLUTION** :

Within UIStaffSlot --> OnPersonnelSelected() either :

1.] Remove the call to OnLoseFocus(). This will fix the issue for mouse & keyboard users as well as controller users; however, like mentioned above, this doesn't really affect mouse & keyboard users. The reason is that, once mouse & keyboard users move their mouse down to the drop down list, the staff slot automatically loses focus; this is the way it should be, since mouse & keyboard focus is based on where the cursor is currently pointing. Since the cursor is no longer pointing at the staff slot, the staff slot shouldn't be focused anymore. Unfortunately, controllers are different; they rely much more greatly upon visual highlighting to signify focus.

2.] Replace the call to OnLoseFocus() with :

```
if (!`ISCONTROLLERACTIVE)
{
	OnLoseFocus();
}
```

In this case, mouse & keyboard code doesn't change, but controller code does. Now, when a controller user selects personnel from the drop down list, the staff slot retains visual focus, which is consistent with the navigator focus it still has. Additionally, we don't run into situations in which all slots appear unhighlighted, and the user becomes confused as to what, exactly, has focus.

### LINKED ISSUE #862: When an unfocused, disabled UIStaffSlot loses focus again, it blinks on then off - WotC Bug with solution
**PROBLEM** : When an unfocused, disabled UIStaffSlot loses focus again, it blinks on then off. This is a War of the Chosen Bug !

**SOLUTION** : Follow the precedent set in UIPanel. Only call OnLoseFocus() if the staff slot is focused, and only call OnReceiveFocus() if the staff slot is not focused.

1.] UIStaffSlot --> OnLoseFocus() should become the following :

```
if (bIsFocused)
{
	bIsFocused = false;
	if (`ISCONTROLLERACTIVE)
	{
		MC.FunctionVoid("mouseOut");
	}
}
```

2.] UIStaffSlot --> OnReceiveFocus() should become the following :

```
if (!bIsFocused)
{
	bIsFocused = true;
	if (`ISCONTROLLERACTIVE)
	{
		MC.FunctionVoid("mouseIn");
	}
}
```

**EXPLANATION** : Each time OnLoseFocus() is called on a UIStaffSlot, it calls MC.FunctionVoid("mouseOut"), regardless of bIsFocused; in other words, even if a staff slot has already lost focus, it still loses focus all over again. 

I traced the issue to an ActionScript call : this.realizeState(), which would then call Utilities.RealizeBGState(this.bg,this.isDisabled,this.isFocus,this.mouseDown). The function RealizeBGState() determines the state of the flash element, such as its disabled status and focus status, then calls the function gotoAndPlay().

**Here is where the problem is** :
1.] A focused, disabled staff slot performs : gotoAndPlay("_DisabledOver"). Now, when I sent in the command : MC.ChildFunctionString("bg", "gotoAndStop", "_DisabledOver"), I ended up with a highlighted button.
2.] An unfocused, disabled staff slot performs : gotoAndPlay("_DisabledUp"). Now, when I sent in the command : MC.ChildFunctionString("bg", "gotoAndStop", "_DisabledUp"), I also ended up with a highlighted button.
3.] When I sent in the command : MC.ChildFunctionString("bg", "gotoAndPlay", "_DisabledUp"), I ended up with a flashing button which went : highlighted --> unhighlighted.
4.] **The conclusion** is that gotoAndPlay("_DisabledUp") goes through a highlighted state, possibly but not necessarily _DisabledOver, then finishes in an unhighlighted state.

**How to Reproduce with a controller** : Go to any facility with a disabled staff slot, then press the right or left bumper to access the Avenger shortcuts. Each time you hit the left or right bumper, the disabled staff slot blinks.

This is because every time you hit the left or right bumper, the UIFacility calls ClearFocusOffofFacilityButtons(). From here :
1.] The UIStaffContainer calls OnLoseFocus(), which then calls OnLoseFocus() on all of its staff slots.
2.] ClearFocusOffofFacilityButtons(), itself, calls OnLoseFocus() on all of the staff slots.

Basically, it is guaranteed that OnLoseFocus() is being called on unfocused staff slots.

GingerAvalanche:
Fixes bugs described by #861 and #862

Closes #861 
Closes #862

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIStaffSlot.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIStaffSlot.uc
@@ -1,0 +1,415 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    UIStaffSlot.uc
+//  AUTHOR:  Joe Weinhoffer
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+
+class UIStaffSlot extends UIPanel
+	dependson(UIPersonnel);
+
+var public StateObjectReference StaffSlotRef;
+
+var UIText m_kStaffName;
+var UIText m_kStaffStatus;
+var UIBGBox m_kBG;
+
+var localized string m_strOpenSlot;
+var localized string m_strCannotEmptySlotTitle;
+var localized string m_strCannotEmptySlotText;
+var localized string m_strCannotMoveStaffTitle;
+var localized string m_strCannotMoveStaffText;
+
+var UIStaffContainer StaffContainer;
+var StaffUnitInfo m_PendingStaff; // used when reassigning staff
+var bool m_QueuedDropDown;
+var bool bSizeRealized;
+var bool  IsDisabled;
+
+var public delegate<OnStaffUpdated> onStaffUpdatedDelegate;
+
+delegate OnStaffUpdated();
+
+//-----------------------------------------------------------------------------
+
+simulated function UIStaffSlot InitStaffSlot(UIStaffContainer OwningContainer, StateObjectReference LocationRef, int SlotIndex, delegate<OnStaffUpdated> onStaffUpdatedDel)
+{
+	// Subclasses need to initialize StaffSlotRef
+
+	InitPanel();
+	
+	StaffContainer = OwningContainer;
+	onStaffUpdatedDelegate = onStaffUpdatedDel;
+
+	ProcessMouseEvents(OnClickStaffSlot);
+
+	UpdateData();
+
+	return self;
+}
+
+simulated function OnCommand(string cmd, string arg)
+{
+	local array<string> sizeData;
+	if (cmd == "RealizeDimensions")
+	{
+		sizeData = SplitString(arg, ",");
+		X = float(sizeData[0]);
+		Y = float(sizeData[1]);
+		Width = float(sizeData[2]);
+		Height = float(sizeData[3]);
+		bSizeRealized = true;
+
+		// update location of dropdown that might be attached to this Staff Slot
+		if(m_QueuedDropDown || (StaffContainer.m_kPersonnelDropDown != none && StaffContainer.m_kPersonnelDropDown.bIsVisible &&  StaffContainer.m_kPersonnelDropDown.SlotRef == StaffSlotRef))
+		{
+			ShowDropDown();
+			m_QueuedDropDown = false;
+		}
+	}
+}
+
+simulated function QueueDropDownDisplay()
+{
+	m_QueuedDropDown = true;
+}
+
+simulated function UIStaffSlot SetDisabled(bool disabled)
+{
+	if(IsDisabled != disabled)
+	{
+		IsDisabled = disabled;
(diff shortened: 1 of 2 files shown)


robojumper (commented):
The comments never mention that it fixes UIStaffSlot. The rendered Bugfixes page will mention the UIStaffSlot source file, but I'd like the comment text to also reference it in some capacity.

pledbrook:
Could you also split the first commit into two: one that adds the file and a second commit with the actual changes to that file? Unless that's how it is right now. The commit message implies it's not.

GingerAvalanche:
Both concerns have been addressed, plus the change commits have been squashed so you can view all edits in a single commit. Github was only letting me select 5 commits to view the changes for at a time, which was annoying because I wanted to check 6 commits worth of diff. Figured you might end up with that issue too.

Still can't change labels, or I'd change it to waiting-for-review.

pledbrook:
Thanks for splitting the file addition out. The implementation looks good to me. I hate to ask, but would you mind improving the second commit message? It doesn't need to contain any information about the multiple commits that existed before. It should just describe what changed and why. So perhaps something like:

> Fix behavior of staff slots with a controller (#861 & #862)
>
> This improves the behavior of staff slots for players that use a controller by
> ensuring that the slot gains and loses focus more appropriately than before.

In the meantime, I'll test the changes in game. I'm assuming that everything is working fine, but just want to be sure.

GingerAvalanche:
Done. Sorry about the commit messages, I first learned collaborative git by using this very repo and still don't have enough experience with the collaboration aspect to have good git documentation habits.

pledbrook (approved):
Looks good, thanks

pledbrook:
@robojumper Do the latest changes satisfy your request on the documentation side?

robojumper (approved):

PR #867, opened 2020-05-19
UIStaffContainer fix

### LINKED ISSUE #866: Hook within UIStaffContainer --> OnReceiveFocus() which allows for its custom logic to be ignored
**General** : A facility, represented by UIFacility, generally speaking, consists of a staff slot container and a room container. The staff slot container, located on the top of the screen, stores the staff slots; the room container, near the bottom of the screen, stores the room buttons. For example, the Guerilla Tactics School has 1 staff slot at the top with which to "Train Rookies", and 1 room button labelled "New Combat Tactics" whereby you can increase the squad size ... etc.

**Explanation** : The description below assumes a controller is being used, since this is the situation I am concerned with.

When a staff slot, UIStaffSlot, is clicked, a drop down menu is diplayed via a call to ShowDropDown() on the staff slot container. Now, at the end of ShowDropDown(), the variable SelectedStaffSlot is set to the staff slot which was just clicked. Basically, when you click on a staff slot, its container stores it as the selected staff slot, via the variable SelectedStaffSlot.

If the staff slot container loses focus at some point, which would happen if you navigate down to the room button, then gains focus again, focus is then automatically transferred through to the SelectedStaffSlot. This is done via UIStaffContainer --> OnReceiveFocus(), which calls the following code :

```
if(SelectedStaffSlot != none)
	SelectedStaffSlot.OnReceiveFocus();
```
	
**Problem** : Say you have 2 staff slots and 1 room button; this is the case with the Proving Ground in LWotC. 
1.] Assume you click on the 1st staff slot with the A button, a drop down menu appears, then you cancel the selection with the B button; at this point SelectedStaffSlot is set to the 1st staff slot. 
2.] Now, assume you press D-Pad right, at which point the navigator within the staff slot container points to the 2nd staff slot. 
3.] You then press D-Pad down to select the room container and room button; the staff slot container loses focus.
4.] You press D-Pad up; the room container loses focus and the staff slot container receives focus. The problem is that SelectedStaffSlot points to the 1st staff slot while the staff slot container's navigator points to the 2nd staff slot. UIStaffContainer --> OnReceiveFocus() places focus on SelectedStaffSlot (1st staff slot); however, an A button press will take into account the navigator (2nd staff slot).

**Request** : Within UIStaffContainer --> OnReceiveFocus() I would like a hook, if that's the right word for it, that allows me to exit the function before any of its custom logic is called. In other words, I would like it to call super.OnReceiveFocus(), but nothing else. This way I can write my own logic for navigation and not have to worry about the UIStaffContainer overriding some of what I have done.

GingerAvalanche:
Implements #866 fix

Closes #866

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIStaffContainer.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIStaffContainer.uc
@@ -1,0 +1,203 @@
+//----------------------------------------------------------------------------
+//  *********   FIRAXIS SOURCE CODE   ******************
+//  FILE:    UIStaffContainer.uc
+//  AUTHOR:  Brit Steiner 
+//  PURPOSE: Staff container that will load in and format staff items. 
+//----------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//----------------------------------------------------------------------------
+
+class UIStaffContainer extends UIPanel
+	dependson(UIStaffSlot);
+
+var string Title;
+var string Message;
+var string Icon;
+var string Skill;
+var array<UIStaffSlot> StaffSlots; 
+
+var UIPersonnel_DropDown m_kPersonnelDropDown;
+
+var localized string DefaultStaffTitle; 
+
+var UIStaffSlot SelectedStaffSlot;
+simulated function UIStaffContainer InitStaffContainer(optional name InitName, optional string NewTitle = DefaultStaffTitle)
+{
+	InitPanel(InitName);
+	
+	SetTitle(NewTitle);
+
+	return self;
+}
+simulated function bool IsEmpty()
+{
+	local int i;
+
+	for (i = 0; i < StaffSlots.Length; ++i)
+	{
+		if (StaffSlots[i].IsSlotFilled())
+			return false;
+	}
+	return true;
+}
+
+
+simulated function OnLoseFocus()
+{
+	local int i;
+
+	super.OnLoseFocus();
+
+	if(m_kPersonnelDropDown != none)
+		m_kPersonnelDropDown.Hide();
+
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UIStaffContainer.uc
```diff
@@ -0,0 +1,217 @@
...
+	/// HL-Docs: ref:Bugfixes; issue:866
+	/// SelectedStaffSlot is only set during ShowDropDown() and only used during OnReceiveFocus(),
+	/// the usage of which will preclude the call to Navigator.GetSelected() which is a more
+	/// comprehensive solution for deciding which staff slot to focus on. It takes longer to run,
+	/// but given that it runs every time staff slots are switched without a dropdown being shown
+	/// anyway (plus it's on a UIPanel on a staff slot - not exactly code that's going to be run
+	/// every frame), SelectedStaffSlot seems to be a solution to a problem that never existed.
+	///
+	/// Setting it to none during OnLoseFocus() ensure that OnReceiveFocus() will always call
+	/// Navigator.GetSelected(), which should always focus the correct UIPanel.
```
robojumper, line 67: Bugfixes should have a comparatively short documentation, see https://x2communitycore.github.io/X2WOTCCommunityHighlander/Bugfixes/ .

Feel free to keep the detailed explanations as a regular comment.

robojumper (changes requested):

GingerAvalanche:
Resolved requested changes and split the commit into two for easier detailing of my actual changes.

pledbrook (approved):
Seems to work just fine, thanks. I would prefer a commit message that doesn't rely on someone having to follow a link just to find out the what and why of the commit, but I can live with it.

robojumper (approved):

PR #870, opened 2020-05-23
Loading saves hook

### LINKED ISSUE #808: Provide a hook to run code on every save load with the mod enabled
It's possible to achieve this already [using a fake DLCInfo with `OnLoadedSavedGame`](https://github.com/WOTCStrategyOverhaul/CovertInfiltration/blob/8fbbc658aa2a7d84acd57dd54bea4ef33ee8b97f/CovertInfiltration/Src/CovertInfiltration/Classes/X2DownloadableContentInfo_CovertInfiltration_Fake.uc) but this creates a "waste" for every DLCInfo call for every mod that uses such method.

We should investigate providing a centralized hook via CHL to avoid the "waste".

Xymanek:
Closes #808

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
@@ -754,3 +754,59 @@ static function UnitPawnPostInitAnimTree(XComGameState_Unit UnitState, XComUnitP
 static function ModifyGeneratedUnitAppearance(XGCharacterGenerator CharGen, const name CharacterTemplateName, const EGender eForceGender, const name nmCountry, const int iRace, const name ArmorName, XComGameState_Unit UnitState, XComGameState UseGameState)
 {}
 /// End Issue #783
+
+// Start issue #808
+/// HL-Docs: feature:OnLoadedSavedGameWithDLCExisting; issue:808; tags:
+/// When loading a save the game makes a distinction between "existing" and "new" mods/DLCs.
+/// The list of the "existing" mods is stored inside the save and is used as "source of truth" during the loading process.
+/// 
+/// First, the game checks if any of the "existing" mods are currently not active. If such exist, the player gets
+/// the "missing mods" popup.
+/// 
+/// Then, the game checks whether any of the currently active mods are not listed as "existing". Such
+/// mods are considered "new" and the `OnLoadedSavedGame` hook is called on their DLCInfos.
+/// 
+/// Finally, the "new" mods are marked as "existing" to prevent the previous step from occurring again the next
+/// time the same campaign is loaded and to facilitate the popup, should any of them be removed.
+/// 
+/// The above process misses an important aspect - what happens if the mod is "existing" but wants to make
+/// state changes before the save is loaded? An example use case would be adjusting existing campaigns due to
+/// updates in the mod code. `OnLoadedSavedGameWithDLCExisting` fills that gap - it is called on the "existing"
+/// mods every time a save is loaded.
+///
+/// Important note 1: `OnLoadedSavedGameWithDLCExisting` is exclusive with `OnLoadedSavedGame`. If the mod was just
+/// added (it is "new") then only `OnLoadedSavedGame` will be called. On subsequent loads of saves from that
+/// campaign (the mod is now "existing") only `OnLoadedSavedGameWithDLCExisting` will be called.
+/// 
+/// Important note 2: this (and the base game OnLoadedSavedGame) is called before the ruleset of the save is initialized.
+/// This is great as any state changes done here will be picked up automatically (no need to refresh anything anywhere),
+/// however it imposes several limitations:
+///
+/// 1. You cannot use `SubmitGameStateContext`/`SubmitGameState`. Use `XComGameStateHistory::AddGameStateToHistory` instead
+/// 2. Event listener templates should be assumed as not registered
+/// 3. In fact, due to (1), only `ELD_Immediate` listeners (that are registered on state objects) will be triggered.
+///    Therefore, you are advised to not trigger any events at all.
+///
+/// If the above is too limiting for your use case, consider using `OnLoadedSavedGameToStrategy`/`OnLoadedSavedGameToTactical`,
+/// which are called after the relevant listener templates are registered and most of the presentation has loaded.
+///
+/// Important note 3: in case the save is loaded mid-tactical, some strategy state objects will be **missing** from the history.
+/// This is intended behaviour and you must account for it when using this hook. You can read more about it here: 
+/// https://robojumper.github.io/too-real/history/#archived
+///
+/// Important note 4: the list of "existing" mods is not cleared when the mod is removed (in order to facilitate the popup).
+/// This means that add -> save -> remove -> load -> save -> add -> load will trigger `OnLoadedSavedGameWithDLCExisting`
+/// as the mod will be considered "existing". However, any state objects which are instances of class(es) added by the mod
+/// **will be gone** as they will fail to deserialize when the save is loaded without the mod active. The only exception to
+/// this is the "Remove Missing Mods" mod which removes the missing mods from the list of the "existing" ones. In that case,
+/// the mod will be considered "new" (again).
+///
+/// *While any mod can potentially manipulate that list, the "Remove Missing Mods" mod is currently the only known way of
+/// removing entries from said list*
+///
+/// Important note 5: the decision to consider a mod either "new" or "existing" is made using its `DLCName`.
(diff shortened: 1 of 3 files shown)


robojumper (approved):

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -762,3 +762,26 @@ final function string GetDisplayName()
 	return ModDependency.DisplayName;
 }
 /// End Issue #524
+
+// Start issue #808
+/// HL-Docs: feature:OnLoadedSavedGameWithDLCExisting; issue:808; tags:
+/// Base game provides the OnLoadedSavedGame hook, which is called only when the mod is added to an in-progress campaign for the first time.
+/// However, there are cases when the mods need to react to save loads for existing campaigns.
+/// An example would be making state changes due to updates in the mod.
```
pledbrook, line 770: I personally find this description ambiguous and had to look at the code to understand what was meant. I suggest an alternative:

```suggestion
/// The base game already provides the OnLoadedSavedGame hook, which is called when a mod is added to
/// XCOM 2 after the saved game that's being loaded was saved. In other words, when a game is saved, it
/// records which mods are active at that time. This hook then notifies a mod if a saved game that doesn't
/// have a record of that mod is loaded.
///
/// However, there are cases when mods need to react when saved games that do already have references
/// to them are loaded. In other words, the mods were active when the game was saved, but they still want
/// to handle situations when that saved game is loaded again. For example, they may want to make state
/// changes due to updates in the mod itself.
```

It still sounds awkward to me, but it's somewhat clearer.

Xymanek:
Note to self: also add a note regrading cases where in-tactical saves are loaded

Iridar:
I've read through the docs carefully, and I still do not understand the purpose of this PR. What can it do what cannot be done by using `OnLoadedSavedGameToStrategy()` and `OnLoadedSavedGameToTactical()` together? 

Judging by the code, assuming I read it correctly, this is meant to be an opposite to `OnLoadedSavedGame()`, which runs if the mod wasn't active when the save was made, while `OnLoadedSavedGameWithDLCExisting()` runs *only* if the mod was active when the save was made? 

If my understanding is correct, perhaps the docs should mention this fact *before* going into great details about the game's logic, as well as why would anyone need it *over* other existing tools.

I have particular gripes with these lines:

> what happens if the mod is "existing" but wants to make state changes before the save is loaded?

The mod uses `OnLoadedSavedGameToStrategy()` and `OnLoadedSavedGameToTactical()`.

> `OnLoadedSavedGameWithDLCExisting` fills that gap - it is called every time a save is loaded

Eeeeeexcept the case if the mod wasn't present when the save was made. Highly misleading, because this made me think this was some sort of `AnySaveGameLoaded()` hook, so, like Peter, I had to look at the code.

robojumper (changes requested):
So one reason this *should* exist in some form is that the existing hooks happen after some of presentation layer stuff etc. has already loaded. The new hook and `OnLoadedSavedGame` run immediately after deserializing the History, which is super important for doing migration work on the save file.

But yes, the one doc line that says "every time a save is loaded" needs to be changed.

Xymanek:
@robojumper the docs have been significantly expanded but I forgot to update the "better compare" link which is why you were probably looking at the old version.

I updated the link - please take a look again

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -762,3 +762,52 @@ final function string GetDisplayName()
...
+/// mods are considered "new" and the `OnLoadedSavedGame` hook is called on their DLCInfos.
+/// 
+/// Finally, the "new" mods are marked as "existing" to prevent the previous step from occurring again the next
+/// time the same campaign is loaded and to facilitate the popup, should any of them be removed.
+/// 
+/// The above process misses an important aspect - what happens if the mod is "existing" but wants to make
+/// state changes before the save is loaded? An example use case would be adjusting existing campaigns due to
+/// updates in the mod code. `OnLoadedSavedGameWithDLCExisting` fills that gap - it is called every time a save is loaded
+///
+/// Important note 1: this will not be called when the base game OnLoadedSavedGame is called ("new" mods).
```
robojumper, line 784: Please don't use "important note" to contradict something said in a paragraph earlier. Directly say that `OnLoadedSavedGameWithDLCExisting` is only called when loading a save file where this mod was previously recorded as "existing" and omit the note.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -762,3 +762,52 @@ final function string GetDisplayName()
...
+/// Important note 2: this (and the base game OnLoadedSavedGame) is called before the ruleset of the save is initialized.
+/// This is great as any state changes done here will be picked up automatically (no need to refresh anything anywhere),
+/// however it imposes several limitations:
+///
+/// 1. You cannot use `SubmitGameStateContext`/`SubmitGameState`. Use `XComGameStateHistory::AddGameStateToHistory` instead
+/// 2. Event listener templates should be assumed as not registered
+/// 3. In fact, due to (1), only `ELD_Immediate` listeners (that are registered on state objects) will be triggered.
+///    Therefore, you are advised to not trigger any events at all.
+///
+/// If the above is too limiting for your use case, consider using `OnLoadedSavedGameToStrategy`/`OnLoadedSavedGameToTactical`.
```
robojumper, line 795: "... which are called after the relevant listener templates are registered and most of the presentation has loaded."

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -762,3 +762,52 @@ final function string GetDisplayName()
...
+///
+/// Important note 3: in case the save is loaded mid-tactical, some strategy state objects will be **missing** from the history.
+/// This is intended behaviour and you must account for it when using this hook. You can read more about it here: 
+/// https://robojumper.github.io/too-real/history/#archived
+///
+/// Important note 4: the list of "existing" mods is not cleared when the mod is removed (in order to facilitate the popup).
```
robojumper, line 801: Maybe mention that this list of existing mods is maintained and checked on a DLCName (`*.XComMod` file) basis.

robojumper (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
Xymanek: > contradict something said in a paragraph earlier

I fail to see any contradiction - could you please be more specific about which part you were referring to?

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
Xymanek: I would prefer to have a centralized resource that explains DLCIndentifiers and how they relate to the one set in DLCInfo's config and the common mistake that happens there. If we have one now, I could link to it, but I see no reason to block this PR on this point.

I know that #965 mentions this stuff, but I don't think it's a good idea to link to it since it's not merged yet

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
pledbrook: > `OnLoadedSavedGameWithDLCExisting` fills that gap - it is called every time a save is loaded

followed by:

> this will not be called when the base game OnLoadedSavedGame is called ("new" mods)

I am assuming that @robojumper is requesting a change to the first sentence so that it doesn't say that the hook is called every time, since it's not. I agree with that as it's a bit confusing.

Xymanek:
Squashed down to 2 commits, addressed the comments regarding docs (I believe) and rebased to get rid of the merge conflict

Xymanek:
I think it doesn't make sense to hold this PR for #1003. In fact, when #1003 is implemented, it can directly change all 3 references, rather than requiring "cooperation" from this PR

Iridar (approved):
Looks like you adjusted the docs since the last time I looked. It's much clearer now, although I'm still not sure what kind of use case this hook could have that is not filled by `OnLoadedSavedGameToStrategy`/`OnLoadedSavedGameToTactical`, but I assume you have a good reason for it, so I'm not gonna be the one to wall it. Approved.

pledbrook (approved):

Issue #741, opened 2020-01-21
Trigger events for adding items  to or removing items from the inventory


Musashi1584:
For book keeping of mods that need to sync generated weapons like Primary Secondaries it would help if there would be events in
`XComGameState_HeadquartersXCom.AddItemToHQInventory`,
`XComGameState_HeadquartersXCom.GetItemFromInventory`,
`XComGameState_HeadquartersXCom.UpgradeItems` and
`XComGameState_HeadquartersXComRemoveItemFromInventory`

whenever an item is added or removed from the inventory.

Iridar:
Could use these events in some potential mods, so I do hope this gets implemented, eventually.

Musashi1584:
@Iridar ^

Iridar:
> @Iridar ^

Awwww thanks

Iridar:
Since the modern solution to the outlined use case is using weapons themselves in any slot you want rather than adding additional versions of the same item that can be used in different slot, closing until there's a different use case for the requested events.

PR #888, opened 2020-06-22
Implements #741 - Trigger HQ Events for item add/remove

Musashi1584:
Trigger events for adding items to or removing items from the inventory


Iridar:
Archiving this due to lack of progress over three years.

PR #900, opened 2020-08-06
Update X2TacticalGameRulesetDataStructures.uc

HudsonSavage:
Adds six entries see the discussion on this issue for further information. [https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/issues/137]

(hope I did this correctly)

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
@@ -109,7 +109,6 @@ enum EInventorySlot
 	eInvSlot_AModMedical,
 	eInvSlot_ArmorMod,
 
-
 	// Marker slot, don't use
 	eInvSlot_END_TEMPLATED_SLOTS,
 };
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
@@ -106,6 +106,12 @@ enum EInventorySlot
 	eInvSlot_ExtraBackpack,
 	eInvSlot_SparkGrenadePocket,
 	eInvSlot_AuxiliaryWeapon,
+	eInvSlot_AModHeavy,
+	eInvSlot_AModMedical,
+	eInvSlot_AModUtility,
+	eInvSlot_AModWebbing,
+	eInvSlot_ArmorMod,
+	eInvSlot_DaemonArmor,	
 
 	// Marker slot, don't use
 	eInvSlot_END_TEMPLATED_SLOTS,
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
@@ -106,12 +106,9 @@ enum EInventorySlot
 	eInvSlot_ExtraBackpack,
 	eInvSlot_SparkGrenadePocket,
 	eInvSlot_AuxiliaryWeapon,
-	eInvSlot_AModHeavy,
 	eInvSlot_AModMedical,
-	eInvSlot_AModUtility,
-	eInvSlot_AModWebbing,
 	eInvSlot_ArmorMod,
-	eInvSlot_DaemonArmor,	
+
 
 	// Marker slot, don't use
 	eInvSlot_END_TEMPLATED_SLOTS,


Iridar:
Do you actually need all those six slots? Could you please provide more details about that ArmorMod system you have plans for? Because I suspect you could get away with much fewer slots than you have requested, since it's possible to template a Highlander slot to present itself differently based on arbitrary conditions. The total number of slots that can be given away is limited, so it's important to request only as many slots as absolutely necessary.

HudsonSavage:
I think I can cut at least one, since it`s only for a planned feature,
nothing I`ll get to soon. One other pair of slots do something very similar
to standard utility slots, I`ll get the details, later and perhaps someone
could suggest a way for me to keep the functionality without using two (or
any) additional slots.
Please stand by for a detailed report, sometime this weekend.

HudsonSavage:
The requested changes are for a very extensive utility items mod I've been working on since shortly after WOTC released. A large part of the mod is the Armour Mod system.

Armour mods (AMod) are a class of item that provide a variety of effects through abilities on the item, to the user (some of these unlock further slots.). By default everyone gets the Standard Issue AMod and all but a handful of the AMods grant an additional utility type slot (see eInvSlot_AModUtility)

**NEED**
eInvSlot_ArmorMod,		- Adds a slot to all Xcom units, the slot only accepts the AMod item type, this is fundamental to the mod. 
**PROBABLY NEED**
eInvSlot_AModHeavy,		- Available on one specific AMod, grants an additional Heavy Weapon slot allowing any unit to have a heavy weapon slot and also for units that already have a HWeps slot to get another.
eInvSlot_AModMedical,	- Works like the AMod Heavy slot but for Medical items, I'll be adding a few new medical category items by release time, this allows two different medical class items to be carried. I prefer this to the previous +1 charge per equipped functionality since it allows for a bit more versatility.
**MAY NOT NEED**
eInvSlot_AModUtility,	- Most AMods unlock this slot, it is set up to accept pretty much anything a normal utility slot does. This could be redundant if there is a better way of adding an additional utility slot that has the same unlock path (skills on items)
eInvSlot_AModWebbing,	- One specific AMod adds this slot, it is the same as AMod Utility except for its unlock criteria being skills from one specific item. This could be redundant if there is a better way of adding an additional utility slot that has the same unlock path (skills on items)
**REMOVE**
eInvSlot_DaemonArmor,	- Remove this, it was future proofing for a planned feature, but at this point I might never get arround to it.
	
This is my assessment of the requested slots, I'd be interested in your thoughts on the **MAY NOT NEED** category. I'd love to hear of a cleaner way of doing the same thing, if it doesn't change how they work from a game play standpoint. If you want further information on this please ask.

Iridar:
Perfect, this is exactly what I needed.

`eInvSlot_ArmorMod, - Adds a slot to all Xcom units, the slot only accepts the AMod item type, this is fundamental to the mod.`
This indeed seems required, unless you're willing to make your AMods an upgrade for the armor equipped by the player. 
Or adding your AMods as Vests, in which case your mod could depend on the Musashi's Vest Slot mod, but it's understandable if you wish to keep your mod separate and independent. 

`eInvSlot_AModHeavy` 
Not needed, you can grant additional Heavy Weapon slots based on arbitrary conditions using the `GetNumHeavyWeaponSlotsOverride` DLC hook in the highlander, for example, the code below will grant an additional heavy weapon slot to units with configured character template names if they have a secondary weapon with the configured weapon category.
```
static function GetNumHeavyWeaponSlotsOverride(out int NumHeavySlots, XComGameState_Unit UnitState, XComGameState CheckGameState)
{
	local XComGameState_Item ItemState;

	if (default.SparkCharacterTemplates.Find(UnitState.GetMyTemplateName()) != INDEX_NONE)
	{
		ItemState = UnitState.GetItemInSlot(eInvSlot_SecondaryWeapon, CheckGameState);
		if (ItemState != none && default.WeaponCategoriesAddHeavyWeaponSlot.Find(ItemState.GetWeaponCategory()) != INDEX_NONE)
		{
			NumHeavySlots++;
		}
	}
}
```
`eInvSlot_AModMedical, - Works like the AMod Heavy slot but for Medical items`
You'll need this, yeah. 

`eInvSlot_AModUtility`
Not needed, you there's a similar DLC hook for utility slots. For example, code below removes a Utility Slot from the Unit if they have more than one and they have a Disposable Rocket Launcher equipped.
```
static function GetNumUtilitySlotsOverride(out int NumUtilitySlots, XComGameState_Item EquippedArmor, XComGameState_Unit UnitState, XComGameState CheckGameState)
{
	if (default.Utility_DRL_Occupies_Two_Slots && NumUtilitySlots > 1 && UnitHasUtilityDRLEquipped(UnitState))
	{
		NumUtilitySlots--;
	}
}
```
`eInvSlot_AModWebbing`
Indeed, this is redundant, extra utility slots can be easily added with the hook above.

HudsonSavage:
Superb, go ahead and remove the ones I don't need and I'll implement them the way you suggest. I hope you don't mind my pinging you on Discord if I run in to trouble?

Iridar:
Please remove the unneeded ones yourself. You can ping me.

HudsonSavage:
Will do, thanks for your assistance.

HudsonSavage:
Made the requested changes, only two additional entries now.

Iridar (approved):

PR #902, opened 2020-08-14
Refactor and test docs script

robojumper:
... in which we use a slightly more modern string interpolation syntax.

We also add a test suite for all the ways in which the docs script can go wrong.


Xymanek (approved):
Since we do not have another python-fluent member in the CHL team who could review this properly, I'll approve this based on the quality of the tests

PR #908, opened 2020-09-05
Docs: Parse, validate event trigger specification, generate listener template

robojumper:
Basically, turns this

```unrealscript
/// ```event
/// EventID: OverrideHasGrenadePocket,
/// EventData: [ inout bool bHasGrenadePocket ],
/// EventSource: XComGameState_Unit (SourceUnit),
/// NewGameState: no
/// ```
```
into this
grafik


Iridar:
This is pretty wizard

Xymanek:
Small note: I've stopped filling `Tuple.Id` in my latest events. Reasons:

* They are already identified by the event id
* If you have listeners being called by different events (without the intention of doing so), then you have far more fundamental problems
* If you have listeners being called by different events (and it's intended) and you want to differentiate, the `EventId` is passed to listeners already
* Some of the events we introduce do **not** use a tuple (instead they pass state objects, UI panels, viz actors, etc, or even nothing) and in such cases there is no equivalent mechanism to checking `Tuple.Id != '...'`

Xymanek (commented):
Note: the `Listener template` from the screenshot in the description does not appear in the test output

robojumper:
Some bikeshedding:

* `EventID`/`EventSource`/`EventData`/`NewGameState` vs `ID`/`Source`/`Data`/`GameState`?
* Allow omitting parameters?
* Allow `[inout bool bHasGrenadePocket]` as a shorthand notation for `XComLWTuple { Data: [inout bool bHasGrenadePocket] }`?

X2WOTCCommunityHighlander/X2WOTCCommunityHighlander.x2proj
```diff
@@ -60,6 +60,9 @@
     <Content Include="Src\Engine\Classes\SkeletalMeshSocket.uc">
       <SubType>Content</SubType>
     </Content>
+    <Content Include="Src\X2WOTCCommunityHighlander\Classes\CHL_Event_Compiletest.uc">
+      <SubType>Content</SubType>
+    </Content>
```
robojumper, line 65: N.b. this actually adds a somewhat test-only file to the project, but it's still compiled and shipped with the `X2WOTCCommunityHighlander` package. Is this a good idea?

X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/CHL_Event_Compiletest.uc
```diff
@@ -0,0 +1,688 @@
...
+{
+	local XComLWTuple Tuple;
+	local bool bShow;
+	local string strHeader;
+	local string strValue;
+	local string strFooter;
+
+	Tuple = XComLWTuple(EventData);
+
+	if (Tuple == None || Tuple.Id != 'OverrideNextRetaliationDisplay') return ELR_NoInterrupt;
```
Xymanek, line 315: This check will never pass (see my comment above about `Tuple.Id`)

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/428f2288c0e9d2c1d4f82876ca5641c7f7521325/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAdventOperations.uc#L269-L290

robojumper: I did catch that one and added the ID in this PR already. In any case, I kind of agree and removed the ID check.

robojumper:
cc @pledbrook -- I believe it was you who started documenting event tuples in a similar format, even before we had HL docs.

docs_src/docs/index.md
```diff
@@ -124,3 +124,59 @@ for example:
...
+{
+	local XComHQPresentationLayer Pres;
+	local XComLWTuple Tuple;
+	local CHLPromotionScreenType PromotionScreenType;
+	local class<UIArmory_Promotion> PromotionUIClass;
+
+	Pres = XComHQPresentationLayer(EventSource);
+	Tuple = XComLWTuple(EventData);
+
+	if (Tuple == None || Tuple.Id != 'OverridePromotionUIClass') return ELR_NoInterrupt;
```
robojumper, line 171: Note to self: Update this

X2WOTCCommunityHighlander/X2WOTCCommunityHighlander.x2proj
Xymanek: Could we exclude it in the build script somehow? Possibly in non-debug builds (since the goal is to check that the file compiles). Or we could even make a new flag + ModBuddy build config

robojumper: New flag + build config it is, the "unused variable" spam would get awful otherwise.

robojumper:
Deleted the verbose tuple data syntax (`EventData: XComLWTuple { Data: [ ... ] }`) in favor of the shorter `EventData: [ ... ]`.

Two unresolved questions:

* Should we allow omitting `EventData`/`EventSource`/`NewGameState` and use default values of `none`/`none`/`no`?
* Naming: Should the keys have the `Event` prefix or is `ID`/`Data`/`Source` enough? Note that the immediately preceding line will be \`\`\`event in any case.

robojumper:
Decided on requiring all four of EventID/EventData/EventSource/NewGameState with their full name, and renamed the `no` case of `NewGameState` to `none` to be consistent with the rest. I think this is good to go now.

docs_src/docs/index.md
```diff
@@ -176,3 +176,9 @@ static function EventListenerReturn OnOverridePromotionUIClass(Object EventData,
 	return ELR_NoInterrupt;
 }
 ```
+
+All generated listeners will be dumped into [CHL_Event_Compiletest.uc](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/CHL_Event_Compiletest.uc),
+which will be excluded from normal compilation and only kept in "compiletest" tasks. If you documented a listener (or changed an existing listener documentation)
+but did not run `makeDocs`, CI (automatic scripts running on GitHub servers when you open or update a pull request) will note that `CHL_Event_Compiletest` is
+out of date and the `check-event-templates` check will fail! To fix, run `makeDocs` locally and commit the changes to `CHL_Event_Compiletest`.  
+Future possibility: generate + upload patch file as artifact?
```
Xymanek, line 184: We can potentially write a GH action that will commit the updated file on comment (e.g. `!update-compile-test`)

https://github.com/marketplace/actions/pull-request-comment-trigger

robojumper: Opened #912 because setting this action up will take some experimenting with the PR (test opened, closed PRs, PRs with/without the "allow maintainers to push" checkbox etc.

Xymanek (approved):
While I'm not suited to review the python code, the output looks solid

Xymanek (approved):
My previous review and caveat remains.

Furthermore, I suggest merging this ASAP so that it can be used (and tested more)

robojumper:
Preliminary testing in #931 suggests this is working as intended.

Issue #896, opened 2020-07-21
Stocks and Laser Sights do not benefit from Inside Knowledge


Iridar:
```
[ResCard_InsideKnowledge X2StrategyCardTemplate]
DisplayName="Inside Knowledge"
SummaryText="The effect of all weapon modifications is increased."
```

There is a mod that claims the bug exists and claims to fix it: https://steamcommunity.com/sharedfiles/filedetails/?id=1124909099

I propose to fix this in Highlander instead. However, the first step would be confirming the bug is actually there, because from a quick look at X2Item_DefaultUpgrades, it seems there are measures in place to take the UpgradesEmpowered bonus into account.

Iridar:
Confirmed the bug with Stocks and Laser Sights by gameplay testing, investigating the reason.

For Laser Sights, the bonus amount for Empowered Upgrades is taken into account here:
```
static function int GetCritBonusAmount(X2WeaponUpgradeTemplate UpgradeTemplate)
{
	return UpgradeTemplate.CritBonus + (AreUpgradesEmpowered() ? default.CRIT_UPGRADE_EMPOWER_BONUS : 0);
}
```
However, this function is never called by `X2Effect_LaserSight`, which just gives the hardcoded amount of crit.

Iridar:
Issue with Stocks is a bit different; there's much more noodle code involved, but the gist is the same: The
```
static function int GetDamageBonusAmount(X2WeaponUpgradeTemplate UpgradeTemplate)
{
	return UpgradeTemplate.BonusDamage.Damage + (AreUpgradesEmpowered() ? default.MISS_DAMAGE_EMPOWER_BONUS : 0);
}
``` 
is never taken into account.

Iridar:
For Laser Sights, the natural place to fix this would be the `X2Effect_LaserSight` itself, but there's a caveat. Judging by this line of code in `X2Effect_LaserSight:`, it seems the weapon itself ends up as being the source item for the Laser Sight effect:
```
if (SourceWeapon != none && SourceWeapon.ObjectID == EffectState.ApplyEffectParameters.ItemStateObjectRef.ObjectID)
```
So I don't see an easy way to reliably get the Laser Sight's weapon upgrade item state and its template from inside the Laser Sight effect, so that the `AddCritChanceModifierFn` or `GetBonusAmountFn` on it can be called. 

[...]

PR #914, opened 2020-09-19
Fix Stocks and Laser Sights not benefitting from empowered upgrades (Issue #896)

Iridar:
Fixes Issue #896

`X2Effect_ApplyWeaponDamage` now calls `WeaponUpgradeTemplate.GetBonusAmountFn()`, if it exists, rather than accessing `WeaponUpgradeTemplate.BonusDamage` directly. 

`X2Effect_LaserSight` now adds `class'X2Item_DefaultUpgrades'.default.CRIT_UPGRADE_EMPOWER_BONUS` to the bonus crit chance amount if empowered upgrades are available. 

Both of these implementations fix the bug, as I confirmed with testing, but in both cases it might potentially change the behavior of some obscure mods that have been relying on the things working as they did. So I would appreciate some input here. Feel free to check the associated issue for more context.

Note that this bug is successfully fixed by this mod: 
Insider Knowledge Fix
https://steamcommunity.com/sharedfiles/filedetails/?id=1124909099

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
@@ -467,7 +467,19 @@ simulated function GetDamagePreview(StateObjectReference TargetRef, XComGameStat
 
 					ModifyDamageValue(UpgradeTemplateBonusDamage, TargetUnit, AppliedDamageTypes);
 
-					UpgradeDamageValue.Damage += UpgradeTemplateBonusDamage.Damage;
+					//	Start Issue #896
+					/// HL-Docs: ref:Bugfixes; issue:896
+					///	Call the (hopefully) relevant delegate, if this weapon upgrade has it.
+					/// This makes the guaranteed damage on missed shots added by Stocks properly benefit from Insider Knowledge
+					if (WeaponUpgradeTemplate.GetBonusAmountFn != none)
+					{
+						UpgradeDamageValue.Damage += WeaponUpgradeTemplate.GetBonusAmountFn(WeaponUpgradeTemplate);
+					}
+					else
+					{
+						UpgradeDamageValue.Damage += UpgradeTemplateBonusDamage.Damage;
+					}
+					//	End Issue #896
 					UpgradeDamageValue.Spread += UpgradeTemplateBonusDamage.Spread;
 					UpgradeDamageValue.Crit += UpgradeTemplateBonusDamage.Crit;
 					UpgradeDamageValue.Pierce += UpgradeTemplateBonusDamage.Pierce;
@@ -804,7 +816,19 @@ simulated function int CalculateDamageAmount(const out EffectAppliedData ApplyEf
 					if (UpgradeTemplateBonusDamage.Damage > 0) bHadAnyDamage = true;
 					bWasImmune = bWasImmune && ModifyDamageValue(UpgradeTemplateBonusDamage, kTarget, AppliedDamageTypes);
 
-					UpgradeDamageValue.Damage += UpgradeTemplateBonusDamage.Damage;
+					//	Start Issue #896
+					/// HL-Docs: ref:Bugfixes; issue:896
+					///	Call the (hopefully) relevant delegate, if this weapon upgrade has it.
+					/// This makes the guaranteed damage on missed shots added by Stocks properly benefit from Insider Knowledge
+					if (WeaponUpgradeTemplate.GetBonusAmountFn != none)
+					{
+						UpgradeDamageValue.Damage += WeaponUpgradeTemplate.GetBonusAmountFn(WeaponUpgradeTemplate);
+					}
+					else
+					{
+						UpgradeDamageValue.Damage += UpgradeTemplateBonusDamage.Damage;
+					}
+					//	End Issue #896
 					UpgradeDamageValue.Spread += UpgradeTemplateBonusDamage.Spread;
 					UpgradeDamageValue.Crit += UpgradeTemplateBonusDamage.Crit;
 					UpgradeDamageValue.Pierce += UpgradeTemplateBonusDamage.Pierce;
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_LaserSight.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_LaserSight.uc
@@ -1,0 +1,37 @@
+class X2Effect_LaserSight extends X2Effect_Persistent config(GameData_SoldierSkills);
+
+var config array<int> CritBoostArray;
+
+var int CritBonus;
+
+function GetToHitModifiers(XComGameState_Effect EffectState, XComGameState_Unit Attacker, XComGameState_Unit Target, XComGameState_Ability AbilityState, class<X2AbilityToHitCalc> ToHitType, bool bMelee, bool bFlanking, bool bIndirectFire, out array<ShotModifierInfo> ShotModifiers)
+{
+	local int Tiles;
+	local XComGameState_Item SourceWeapon;
+	local ShotModifierInfo ShotInfo;
+
+	SourceWeapon = AbilityState.GetSourceWeapon();
+	if (SourceWeapon != none && SourceWeapon.ObjectID == EffectState.ApplyEffectParameters.ItemStateObjectRef.ObjectID)
+	{
+		Tiles = Attacker.TileDistanceBetween(Target);
+		if (CritBoostArray.Length > 0)
+		{
+			if (Tiles < CritBoostArray.Length)
+				ShotInfo.Value = CritBoostArray[Tiles];
+			else  //  if this tile is not configured, use the last configured tile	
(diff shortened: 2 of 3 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_LaserSight.uc
```diff
@@ -21,7 +21,10 @@ function GetToHitModifiers(XComGameState_Effect EffectState, XComGameState_Unit
 			else  //  if this tile is not configured, use the last configured tile	
 				ShotInfo.Value = CritBoostArray[CritBoostArray.Length - 1];
 
-			ShotInfo.Value += CritBonus;
+			// Single line for Issue #896
+			/// HL-Docs: ref:Bugfixes; issue:896
+			///	Add bonus crit if Insider Knowledge is present.
+			ShotInfo.Value += CritBonus + AreUpgradesEmpowered() ? class'X2Item_DefaultUpgrades'.default.CRIT_UPGRADE_EMPOWER_BONUS : 0;
```
Xymanek, line 27:
```suggestion
			ShotInfo.Value += CritBonus + `XCOMHQ.bEmpoweredUpgrades ? class'X2Item_DefaultUpgrades'.default.CRIT_UPGRADE_EMPOWER_BONUS : 0;
```

Also remove the `AreUpgradesEmpowered` function

Xymanek (commented):
Code-quality wise this PR is 👍 besides the small suggested improvement.

Regarding any potential side effects - I do not consider myself knowledgeable enough in this area of the game to approve

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_LaserSight.uc
Iridar: Good idea, done.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
```diff
@@ -467,7 +467,19 @@ simulated function GetDamagePreview(StateObjectReference TargetRef, XComGameStat
...
 					ModifyDamageValue(UpgradeTemplateBonusDamage, TargetUnit, AppliedDamageTypes);
 
-					UpgradeDamageValue.Damage += UpgradeTemplateBonusDamage.Damage;
+					//	Start Issue #896
+					/// HL-Docs: ref:Bugfixes; issue:896
+					///	Call the (hopefully) relevant delegate, if this weapon upgrade has it.
+					/// This makes the guaranteed damage on missed shots added by Stocks properly benefit from Insider Knowledge
+					if (WeaponUpgradeTemplate.GetBonusAmountFn != none)
+					{
+						UpgradeDamageValue.Damage += WeaponUpgradeTemplate.GetBonusAmountFn(WeaponUpgradeTemplate);
```
pledbrook, line 476: This feels risky to me. We have no idea what `GetBonusAmountFn` does. Perhaps it does calculate a bonus damage value. But what if it calculates a bonus aim, or bonus crit damage, or something else?

pledbrook: On top of that, CHL appears to have added a `CHBonusDamage` property to weapon upgrade templates. Not sure what that's for. Allegedly a more general bonus damage than `BonusDamage` which is for "miss damage only". Honestly, this feels like a minefield.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_LaserSight.uc
```diff
@@ -0,0 +1,40 @@
...
+		{
+			if (Tiles < CritBoostArray.Length)
+				ShotInfo.Value = CritBoostArray[Tiles];
+			else  //  if this tile is not configured, use the last configured tile	
+				ShotInfo.Value = CritBoostArray[CritBoostArray.Length - 1];
+
+			// Single line for Issue #896
+			/// HL-Docs: ref:Bugfixes; issue:896
+			///	Add bonus crit if Insider Knowledge is present.
+			ShotInfo.Value += CritBonus + `XCOMHQ.bEmpoweredUpgrades ? class'X2Item_DefaultUpgrades'.default.CRIT_UPGRADE_EMPOWER_BONUS : 0;
```
pledbrook, line 27: The issue raises a good question: this is supposed to be taken into account by `X2AbilityToHitCalc_StandardAim.GetHitChance()`, so why is it in `X2Effect_LaserSight`? I'm wondering whether the effect is for units that don't have moddable weapons, such as ADVENT.

I'm a little reluctant to bless this change without understanding why `GetHitChance()` isn't taking it into account.

pledbrook: OK, so `GetHitChance()` will only work with bonus-crit weapon upgrade templates that are configured to do so. Basically, there's a config var array that contains the name of templates that *don't* use effects to add the bonus crit. Since laser sight does use an effect, it's not handled directly by `GetHitChance()`.

One new concern I have is that `X2Effect_LaserSight` could be used by mods to grant enemies the laser sight bonus, and in doing so, they will get the bonus from Insider Knowledge.

Iridar: > One new concern I have is that X2Effect_LaserSight could be used by mods to grant enemies the laser sight bonus, and in doing so, they will get the bonus from Insider Knowledge.

I know it's not exactly representative, but none of the ~800 mods in my reference library use X2Effect_LaserSight for anything. I agree that this is a potential problem, but I don't see any more elegant ways to fix the bug. I suppose we could grant the empowered upgrade bonus only if the unit is present in the XCOM HQ crew?..

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
Iridar: The bonus damage is added only if the `if (WeaponUpgradeTemplate.BonusDamage.Tag == DamageTag)` check passes, which *should* ensure the purpose of the weapon upgrade template is to provide bonus damage.

Iridar:
robojumper said they're not qualified to review this, so somebody else will have to pick up the mantle.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_LaserSight.uc
pledbrook: How about we just add a boolean property that determines whether the Insider Knowledge bonus takes effect? By default it does, but if a mod uses `X2Effect_LaserSight` for some ability that shouldn't take into account Insider Knowledge, they can flip the switch on that boolean.

Iridar: If a mod wants to use X2Effect_LaserSight for something *other* than *exactly* what X2Effect_LaserSight should be doing, they should just make a new X2Effect, potentially extending X2Effect_LaserSight if inheritance is important for some reason.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
pledbrook: That's a fair point. If something trips up on this, we could reasonably claim an abuse of the API. I kind of expect it to happen, but we'll deal with it then. Far more folks will benefit from this change, and hopefully it won't be too hard to work out what's going wrong if/when it does go wrong for someone.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_LaserSight.uc
pledbrook: All Laser Sight is doing is providing a flat crit chance bonus and a range-based one. Is it so unreasonable for a mod to want to apply that effect to a soldier ability for example, rather than a weapon mod? And Insider Knowledge should not apply to the soldier ability. I would certainly prefer a flag for flexibility.

Iridar: Alright, fair point, I'll do so.

Iridar:
This wouldn't build, not sure if due to issues on my end or not, so I rebased it onto current master and everything is fine now. This PR should be good to go.

pledbrook (approved):
LGTM

pledbrook:
Please resolve conflicts.

Iridar:
> Please resolve conflicts.

Done.

pledbrook:
@Iridar I would ask for one more change if that's OK: the second commit's message is pretty bare bones. Would you mind fleshing it out with a description of the change(s)?

Iridar:
> @Iridar I would ask for one more change if that's OK: the second commit's message is pretty bare bones. Would you mind fleshing it out with a description of the change(s)?

No problem at all. Done.

PR #916, opened 2020-10-22
Deprivatise functions related to soldier AP (#915)

### LINKED ISSUE #915: AP calculation functions for soldiers are private
In LWOTC, we would like to integrate combat intelligence into Not Created Equal. WOTC makes that awkward because the functions in `XComGameState_Unit` that return the AP amount for normal and faction soldiers are private, so we have to copy the code instead.

Since both these functions have no side effects, they can safely be made public. And quite frankly they should be public, since they represent important data about soldiers.

The functions in question are `XComGameState_Unit.GetResistanceHeroAPAmount()` and `XComGameState_Unit.GetBaseSoldierAPAmount()`.

pledbrook:
This change deprivatises the functions:

 * `XComGameState_Unit.GetResistanceHeroAPAmount()`
 * `XComGameState_Unit.GetBaseSoldierAPAmount()`

These two functions have no side effects -- they don't change any state -- and they provide useful data to mods, particularly those that want to reconstruct soldier AP amounts after changes to combat intelligence for example.

Resolves #915.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -2993,7 +2993,9 @@ function ImproveCombatIntelligence()
 	AbilityPoints += Round(APIncrease);
 }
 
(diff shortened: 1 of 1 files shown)


Iridar (approved):
Looks good to me.

PR #918, opened 2020-10-29
Custom ability target styles v2

### LINKED ISSUE #763: Implement framework for custom multi targeting styles
Currently it's not possible to "easily" make a custom `X2AbilityMultiTargetStyle` , since the `XComGameState_Ability::GatherAbilityTargets` is native, even though luckily the UnrealScript implementation of its code was included by Firaxis.

That code is calling `AbilityTemplate.AbilityMultiTargetStyle.GetMultiTargetOptions`, which is also native. 

Correct me if I'm wrong, but I think the problem here is that since both functions are native, the native code will talk directly to other piece of native code, bypassing any UnrealScript override functions that you were to put into your `X2AbilityMultiTargetStyle` subclass.

So in essence, you can make a custom Multi Target Style all you want, but its code will never get called by the game. 

robojumper had this to say on this issue:

> My recommendation would be to create a separate subclass of XCGS_Ability, scan for non-native target styles in X2AbilityTemplate and instantiate the subclass when needed

This is something I'd be glad to do if I had a better than vague idea of what this means.

robojumper:
Fixes #763, resubmission of #762.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Ability_CH.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Ability_CH.uc
@@ -24,10 +24,10 @@
 /// ## Compatibility
 ///
 /// In order to selectively override the `XComGameState_Ability` class, we modify
-/// `X2AbilityTemplate:CreateInstanceFromTemplate`. If your mod overrides `CreateInstanceFromTemplate` in a subclass,
+/// `X2AbilityTemplate:CreateInstanceFromTemplate`. If your mod overrides `CreateInstanceFromTemplate` in a subclass
 /// of `X2AbilityTemplate`, the Highlander will not be able to apply this enhancement to that ability. On the other
 /// hand, if you override `CreateInstanceFromTemplate` precisely to instantiate your own ability state object with
-/// `GatherAbilityTargets` de-nativized for custom target style, your abilities will keep working the same way.
+/// `GatherAbilityTargets` de-nativized for a custom target style, your abilities will keep working the same way.
 
 class XComGameState_Ability_CH extends XComGameState_Ability;
 
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Ability_CH.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Ability_CH.uc
@@ -1,0 +1,130 @@
+/// HL-Docs: feature:CustomTargetStyles; issue:763; tags:tactical,compatibility
+/// Target styles are responsible for the initial target selection for tactical abilities.
+/// The `AbilityTargetStyle` builds an initial target list (or none, for cursor-targeted abilities).
+/// The `AbilityMultiTargetStyle` adds other targets to every target built by the `AbilityTargetStyle`
+/// or the cursor target.
+///
+/// Even though the results are cached, this code is extremely performance-critical:
+/// Every time a game state is added to the History, all cached results are invalidated.
+/// Whenever the AI evaluates a unit, the player selects a unit, or a reaction ability checks for
+/// valid targets, this code is run for every ability on the unit!
+///
+/// As a result, all of this code is native: The target styles themselves are native, and their functions
+/// are called by the native `GatherAbilityTargets` function. This native code simply cannot see UnrealScript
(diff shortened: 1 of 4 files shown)


robojumper:
I personally have no use for this, but #762 in limbo bothered me, cc @Xymanek. #762 says LWotC needs this in some form, cc @pledbrook.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Ability_CH.uc
```diff
@@ -0,0 +1,130 @@
...
+/// the game state object for an ability template.
+///
+/// TL;DR Custom ability target/multi-target styles magically start working.
+///
+/// ## Compatibility
+///
+/// In order to selectively override the `XComGameState_Ability` class, we modify `X2AbilityTemplate:CreateInstanceFromTemplate`.
+/// If your mod overrides `CreateInstanceFromTemplate` in a subclass, the Highlander will not be able to apply this enhancement
+/// to that ability. On the other hand, if you override `CreateInstanceFromTemplate` precisely to instantiate your own
+/// ability state object, things will keep working.
```
Xymanek, line 29: But only if the custom state class also un-natives `GatherAbilityTargets`, no?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Ability_CH.uc
```diff
@@ -0,0 +1,130 @@
...
+					Targets.Remove(i, 1);
+			}
+		}
+
+		//The Multi-target style may have deemed some primary targets invalid in calls to CheckFilteredMultiTargets - so CheckFilteredPrimaryTargets must come afterwards.
+		AvailableCode = m_Template.AbilityTargetStyle.CheckFilteredPrimaryTargets(self, Targets);
+		if (AvailableCode != 'AA_Success')
+			return AvailableCode;
+
+		Targets.Sort(SortAvailableTargets);
```
Xymanek, line 86: Could we use `CH_MAKE_QUICKSORT` from #809 since this is so performance critical?

robojumper: Needs profiling. My intuition is that anything but `.Sort()` will worsen performance. How many *available* targets do we expect? Note that #809 currently already cuts off at 15 elements to use InsertionSort.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Ability_CH.uc
robojumper: That's what I meant. Clarified.

pledbrook:
I take it this has nothing to do with targeting methods?

pledbrook:
LWOTC has a `X2AbilityMultiTarget_Cone_LWFlamethrower` class that is a custom, UC targeting style. It seems to work as it is. Should I change something to start making use of this change?

robojumper:
> I take it this has nothing to do with targeting methods?

For cursor-targeted abilities, the multi-target style builds the multi targets for the cursor target created by the targeting method. For everything else, the target style builds the initial target list, the multi-target style builds the multi-targets for every target, and the targeting method only selects a particular target.

robojumper:
`X2AbilityMultiTarget_Cone_LWFlamethrower` should work as is already and doesn't benefit from any change in this PR. since it's cursor-targeted. If I were to suggest a change, it's this TODO:

https://github.com/long-war-2/lwotc/blob/db7ff997a5882cb4950a07861026f8e90317bd24/LongWarOfTheChosen/Src/LW_Overhaul/Classes/X2AbilityMultiTarget_Cone_LWFlamethrower.uc#L78-L83

`UpdateParameters` basically modifies the template, so it's a bit of a hack.

Iridar:
Wait, so it's all done and ready? Just needs a review? Imma get into that soon as I have time.

pledbrook (approved):
I mean, it looks fine to me. I think it just requires some testing to ensure it doesn't break anything. Could use it for the next LWOTC release so that it does get some level of playtesting.

Iridar (approved):
I have tested this a while ago, but forgot to report in. It works perfectly fine and is sufficient to do what I wanted to do, which is to make a version of Faceoff that can target variable number of targets based on pistol's current ammo.

PR #925, opened 2020-12-11
Add extra damage modifier hooks for effects (#923)

### LINKED ISSUE #923: Make damage-multiplier effects more consistent
It is impossible at the moment (without MCOing `X2Effect_ApplyWeaponDamage`) to reliably multiply ability damage because all modifiers are applied in the order that their effect was applied. This means that additive and muliplicative modifiers are intermingled, and since those two operations are not commutative, the final damage calculation depends on the order that they're applied.

`X2Effect_PersistentStatChange` is informative because it recognises this issue and allows you to apply stat changes as either multipliers, additions, or post multipliers (coming after other multipliers and additions). We should add something similar to `X2Effect_ApplyWeaponDamage`.

pledbrook:
This adds 4 extra hooks to `X2Effect_Persistent` that allow mods better control over the order in which damage modifiers are applied. These hooks occur for both attacking and defending modifiers just before the existing modifiers are applied and just after. This primarily helps mods that want to multiply weapon damage by some factor without running into issues with the order effects are applied (since most effects add or substract flat damage and those are not commutative with multipliers).

Resolves #923

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Persistent.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Persistent.uc
@@ -794,7 +794,20 @@ event string GetSpecialDamageMessageName() { return FriendlyName; }
 /// your effect has, so if you want your effects to work with older versions
 /// of the Community Highlander and setups without any Community Highlander at
 /// all, then you should guard the old damage modifier functions with a
-/// [check for Community Highlander version](../../misc/ComponentVersions/).
+/// check for these new functions, e.g.
+/// ```unrealscript
+/// function int GetAttackingDamageModifier(...)
+/// {
+///     if (Function'XComGame.X2Effect_Persistent.GetPreDefaultAttackingDamageModifier_CH' != none)
+///     {
+///         // Using the new hooks, so skip the old implementation
+///         return;
+///     }
+///
+///     // Continue with old implementation here
+///     ...
+/// }
+/// ```
 function float GetPreDefaultAttackingDamageModifier_CH(XComGameState_Effect EffectState, XComGameState_Unit SourceUnit, Damageable Target, XComGameState_Ability AbilityState, const out EffectAppliedData ApplyEffectParameters, float WeaponDamage, XComGameState NewGameState) { return 0.0; }
 function float GetPreDefaultDefendingDamageModifier_CH(XComGameState_Effect EffectState, XComGameState_Unit SourceUnit, XComGameState_Unit TargetUnit, XComGameState_Ability AbilityState, const out EffectAppliedData ApplyEffectParameters, float WeaponDamage, XComGameState NewGameState) { return 0.0; }
 function float GetPostDefaultAttackingDamageModifier_CH(XComGameState_Effect EffectState, XComGameState_Unit SourceUnit, Damageable Target, XComGameState_Ability AbilityState, const out EffectAppliedData ApplyEffectParameters, float WeaponDamage, XComGameState NewGameState) { return 0.0; }
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Persistent.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Persistent.uc
@@ -808,10 +808,10 @@ event string GetSpecialDamageMessageName() { return FriendlyName; }
 ///     ...
 /// }
 /// ```
-function float GetPreDefaultAttackingDamageModifier_CH(XComGameState_Effect EffectState, XComGameState_Unit SourceUnit, Damageable Target, XComGameState_Ability AbilityState, const out EffectAppliedData ApplyEffectParameters, float WeaponDamage, XComGameState NewGameState) { return 0.0; }
-function float GetPreDefaultDefendingDamageModifier_CH(XComGameState_Effect EffectState, XComGameState_Unit SourceUnit, XComGameState_Unit TargetUnit, XComGameState_Ability AbilityState, const out EffectAppliedData ApplyEffectParameters, float WeaponDamage, XComGameState NewGameState) { return 0.0; }
-function float GetPostDefaultAttackingDamageModifier_CH(XComGameState_Effect EffectState, XComGameState_Unit SourceUnit, Damageable Target, XComGameState_Ability AbilityState, const out EffectAppliedData ApplyEffectParameters, float WeaponDamage, XComGameState NewGameState) { return 0.0; }
-function float GetPostDefaultDefendingDamageModifier_CH(XComGameState_Effect EffectState, XComGameState_Unit SourceUnit, XComGameState_Unit TargetUnit, XComGameState_Ability AbilityState, const out EffectAppliedData ApplyEffectParameters, float WeaponDamage, XComGameState NewGameState) { return 0.0; }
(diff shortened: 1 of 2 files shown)


pledbrook:
Having looked at its implementation, the Cover-based Damage Reduction would definitely benefit from this change. It currently MCOs `X2Effect_ApplyWeaponDamage` and `X2Effect_Shredder` to ensure that its damage reduction applies *after* all the other damage modifiers. But doing this means that it doesn't work for the preview.

Using these hooks would allow it to keep the same behaviour while also showing its effect in the damage preview (and without using MCOs of course).

MrNiceUK:
Your `DamageMods` won't add up to the actual total extra damage once fractions are involved, since the `ModifierInfo.Value = EffectDmg;` assigment will truncate. Also means you can get zero entries since the check to avoid that is before that rounding. May be instructive to see #313 for how I solve a similar issue when fixing the broken (for other reasons) `GetStatModifiers()` method.

Grobobobo:
He has taken that into account with this:

`EffectDmg = Round(CurrDamage) - Round(OldDamage);`

Quote from him:
>  Let's say that the first damage mod is +1.2, which gets rounded to +1.
> Then the next one is +1.2 again. That gets rounded to +1. So now the total damage mod is +2.4, with the sum of rounded values being +2.
> If we kept rounding down, then the next +1.2 would result in +3.6 and +3. But, the logic should mean that the third +1.2 gets rounded up to +2 to compensate.
> It's worth verifying that is what actually happens, though

pledbrook:
So to clarify, the `ModifierInfo.Value = EffectDmg;` won't truncate because `EffectDmg` is already an integer via the line:
```
EffectDmg = Round(CurrDamage) - Round(OldDamage);
```
The current implementation means that the individual shot modifiers aren't necessarily accurate because of the rounding, but the *sum* of them should be accurate. I'm just distributing the rounding numbers to specific shot modifiers when needed.

pledbrook:
Based on a discussion with @MrNiceUK, it probably makes sense to truncate the final value for consistency with how XCOM 2 generally handles fractional damage values. Note that I'm only suggesting truncation of the *final* value, not the intermediate ones.

pledbrook:
This is now ready for review in case anyone wants to take a look. I've added docs and switched to truncating the final damage value rather than using `Round()`.

Iridar:
I'll take a look when I can.

Iridar (approved):
A cursory look did not identify any issues, everything seems to be neat and tidy. Do you want me to test these as well?

pledbrook:
They're tested in LWOTC, but not in the base game. So 🤷 Any extra testing is always welcome of course.

On another note, I want to update the docs based on Mr Nice's feedback.

pledbrook:
Ready for review again. Should be squashed and merged when ready.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Persistent.uc
```diff
@@ -692,6 +692,128 @@ function bool AreMovesVisible() { return true; }
...
+/// of the Community Highlander and setups without any Community Highlander at
+/// all, then you should guard the old damage modifier functions with a
+/// check for these new functions, e.g.
+/// ```unrealscript
+/// function int GetAttackingDamageModifier(...)
+/// {
+///     if (Function'XComGame.X2Effect_Persistent.GetPreDefaultAttackingDamageModifier_CH' != none)
+///     {
+///         // Using the new hooks, so skip the old implementation
+///         return;
```
pledbrook, line 804:
```suggestion
///         return 0;
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Persistent.uc
```diff
@@ -692,6 +692,128 @@ function bool AreMovesVisible() { return true; }
...
+/// ```
+/// The arguments are largely self explanatory, but note that `WeaponDamage`
+/// is a float, as is the return value. `WeaponDamage` is the current total
+/// damage, which is the base damage with any preceding damage modifiers
+/// already applied.
+///
+/// Your implementations of these functions should return the amount of
+/// damage your effect is adding or subtracting, **not** the overall damage.
+/// The value you return will be added to `WeaponDamage`.
+///
```
Iridar, line 748: Nitpick: while it's great that there is such a comment, I'm not sure it's necessary to mention it twice in an already somewhat lengthy docs page.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Persistent.uc
```diff
@@ -692,6 +692,128 @@ function bool AreMovesVisible() { return true; }
...
+/// The following hook is used for effects that are applied to the
+/// attacker:
+/// ```unrealscript
+/// float GetPreDefaultAttackingDamageModifier_CH(
+///         XComGameState_Effect EffectState,
+///         XComGameState_Unit SourceUnit,
+///         Damageable Target,
+///         XComGameState_Ability AbilityState,
+///         const out EffectAppliedData ApplyEffectParameters,
+///         float WeaponDamage,
```
Iridar, line 725: Nitpick: not sure I agree with using the `WeaponDamage` name for the "total damage so far" kind of variable.

Also, I'm a bit confused. If a soldier has multiple effects that use, say, `GetPreDefaultDefendingDamageModifier_CH`, won't there be an exact same issue with the final damage numbers depending on the order of these effects? I've looked at the code, and that was my impression.

If that is the case, I'd argue that this implementation is not much of an improvement over vanilla, and I would suggest to reimplement it so that rather than "total damage so far", each hook gets the "damage before modifiers" as an `in` argument, while X2Effect_ApplyWeaponDamage stores the accumulated bonus damage in some other variable.

So that each X2Effect_Persistent gets exact same input arguments regardless of the order.

Iridar (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Persistent.uc
pledbrook: > Also, I'm a bit confused. If a soldier has multiple effects that use, say, GetPreDefaultDefendingDamageModifier_CH, won't there be an exact same issue with the final damage numbers depending on the order of these effects? I've looked at the code, and that was my impression.

If you stick to commutative operations, such as multiplication, then there is no issue. The problem with the default implementation is that it mixes additive and multiplicative modifiers, which are not commutative. This is why I preferred to name these functions to indicate that they should be used for percentage/fraction-based multipliers.

> So that each X2Effect_Persistent gets exact same input arguments regardless of the order

I'm pretty sure that would defeat the purpose. Let's say you have a 50% damage reduction, a 25% damage increase and an 80% damage increase. With the current implementation you get:

  Final Damage = Current Damage * 0.5 * 1.25 * 1.8

Or in a different order:

  Final Damage = Current Damage * 1.8 * 0.5 * 1.25

Your suggestion would be:

  Final Damage = (Current Damage * -0.5) + (Current Damage * 0.25) + (Current Damage * 0.8)

Each of those terms would indeed be commutative, but you have a different result.

pledbrook:
Updated docs based on feedback and replaced `WeaponDamage` with `CurrentDamage`.

Iridar (approved):

PR #928, opened 2020-12-23
Support campaign-start event listeners (#297)

### LINKED ISSUE #297: Investigate registering listeners on the shell menu

### LINKED ISSUE #869: Events during campaign setup are "lost" regarding listener templates
When a campaign is first launched, the initial set of state creation is done in the shell map. Shell has no listener templates (see #297) and as such, mods cannot react to the events (hooks) triggered during that phase. Below are the events that get triggered when using just CHL (based on CI's latest development version, 3c3bdd565e9bef45e0f261ac74f33b7fe0257bf8):

```
UnitRandomizedStats
OverrideClipSize
OverrideRandomizeAppearance
OverrideHasHeavyWeapon
OverrideItemMinEquipped
OverrideHasGrenadePocket
NewCrewNotification
UnitRankUp
OverrideAddChosenTacticalTagsToMission
OverrideScienceScore
OverrideCurrentDoom
SquaddieItemStateApplied
```

Source behind generating this list: 5fc965c20a3fc7eaace430c37faffcda5833b4b6
Full output + parser: https://gist.github.com/Xymanek/b8ac7021900d17ba417d2f7759151525
Note that the list above potentially does not include all events triggered, in case some events are triggered in response to other events (I had to practically disable all listeners to investigate)

An easy solution would be to register strategy listener templates before starting the process, since these are mostly strategy events. Downside is that the listeners might not expect to be called in not-ready-campaign state and might not expect to be called while a start state is opened.

An opt-in solution would be adding `RegisterInStrategySetup` to `CHEventListenerTemplate`; obviously that would require mod authors to manually mark their listeners, so some bugs resulting from the above behaviour might remain.

cc @Musashi1584 @pledbrook since I know some of these events are yours

pledbrook:
This change adds an extra `RegisterInCampaignStart` boolean to CHL event listener templates (on top of the existing `RegisterInTactical` and `RegisterInStrategy` ones). Listeners registered for campaign start will now receive events that are fired during campaign initialization, which strategy listeners don't receive.

Resolves #869.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2EventListenerTemplateManager.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2EventListenerTemplateManager.uc
@@ -1,0 +1,77 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2EventListenerTemplateManager.uc
+//  AUTHOR:  David Burchanowsk
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+class X2EventListenerTemplateManager extends X2DataTemplateManager
+	native(Core);
+
+static function native X2EventListenerTemplateManager GetEventListenerTemplateManager();
+
+event X2EventListenerTemplate FindEventListenerTemplate(name TemplateName)
+{
+	local X2EventListenerTemplate Template;
+
+	Template = X2EventListenerTemplate(FindDataTemplate(TemplateName));
+	if(Template == none)
+	{
+		`Redscreen("Could not find X2EventListenerTemplate " $ TemplateName);
+	}
+
+	return Template;
+}
+
+static function RegisterTacticalListeners()
+{
+	local X2EventListenerTemplate ListenerTemplate;
+	local X2DataTemplate Template;
+
+	// unregister any previously registered templates
+	UnRegisterAllListeners();
+
+	foreach GetEventListenerTemplateManager().IterateTemplates(Template)
+	{
+		ListenerTemplate = X2EventListenerTemplate(Template);
+		if(ListenerTemplate.RegisterInTactical)
+		{
+			ListenerTemplate.RegisterForEvents();
+		}
+	}
+}
+
+static function RegisterStrategyListeners()
+{
+	local X2EventListenerTemplate ListenerTemplate;
+	local X2DataTemplate Template;
+
+	// unregister any previously registered templates
+	UnRegisterAllListeners();
+
+	foreach GetEventListenerTemplateManager().IterateTemplates(Template)
+	{
+		ListenerTemplate = X2EventListenerTemplate(Template);
+		if(ListenerTemplate.RegisterInStrategy)
+		{
+			ListenerTemplate.RegisterForEvents();
+		}
+	}
+}
+
+static function UnRegisterAllListeners()
+{
+	local X2EventListenerTemplate ListenerTemplate;
+	local X2DataTemplate Template;
+
+	foreach GetEventListenerTemplateManager().IterateTemplates(Template)
+	{
+		ListenerTemplate = X2EventListenerTemplate(Template);
+		ListenerTemplate.UnRegisterFromEvents();
+	}
+}
+
+defaultproperties
+{
+	TemplateDefinitionClass=class'X2EventListener'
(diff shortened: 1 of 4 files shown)


Xymanek:
I suggest using #869 as the tracking issue instead. #297 is a more general issue regrading events in the shell and this PR doesn't cover all use cases (e.g. UI events in the shell)

pledbrook:
I linked to #297 to prompt a discussion on that particular issue. Should this solution actually be registering in UIShell instead, as suggested by #297? Does this solution satisfy all the currently known use cases for #297 instead?

Xymanek:
> Does this solution satisfy all the currently known use cases for #297 instead?

No, it does not, e.g. UI events (Musashi's comment)

> Should this solution actually be registering in UIShell instead, as suggested by #297? 

No, as the concerns raised there are not addressed (e.g. the history reset) and are not really relevant for the current use case 

> I linked to #297 to prompt a discussion on that particular issue

I mean.... issues can be linked without being used as the tracking issue (for purposes of docs)

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2EventListenerTemplateManager.uc
```diff
@@ -59,6 +59,26 @@ static function RegisterStrategyListeners()
...
+// Start Issue #297
+//
+// Registers CHL event listeners that are configured to listen for
+// events during a new campaign start.
+static function RegisterCampaignStartListeners()
+{
+	local CHEventListenerTemplate ListenerTemplate;
+	local X2DataTemplate Template;
+
+	foreach GetEventListenerTemplateManager().IterateTemplates(Template)
```
Xymanek, line 71: `RegisterTacticalListeners`/`RegisterStrategyListeners` call `UnRegisterAllListeners` first, perhaps the same should be done here?

Xymanek (commented):
Searching SrcOrig for `CreateStrategyGameStart(` reveals 9 calls to it, some in rather weird places, e.g. `XComMPLobbyGRI::CreateMPStartState`, `UITLE_LadderModeMenu::CreateNewLadder`, `UIMPShell_Lobby::SetupStartState`.

I'm concerned for cases where these listeners would linger past the intended lifetime, so perhaps they could be unregistered at the end of `CreateStrategyGameStart`?

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2EventListenerTemplateManager.uc
pledbrook: Yeah, you're probably right. I was thinking that it was the start of the campaign, so there wouldn't be any listeners. But leaving a campaign and starting a new one within the same session may leave listeners around. As may other stuff, so seems like a good idea to unregister first.

pledbrook:
> Searching SrcOrig for `CreateStrategyGameStart(` reveals 9 calls to it, some in rather weird places, e.g. `XComMPLobbyGRI::CreateMPStartState`, `UITLE_LadderModeMenu::CreateNewLadder`, `UIMPShell_Lobby::SetupStartState`.
> 
> I'm concerned for cases where these listeners would linger past the intended lifetime, so perhaps they could be unregistered at the end of `CreateStrategyGameStart`?

How about I move the listener registration to `UIShellDifficulty` [here](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIShellDifficulty.uc#L895). That would cover both new campaigns and new campaigns with tutorial, but none of the other cases.

Iridar:
I'd like to test or review this, but I'm not sure what are the use cases for something like this? That don't fit in `InstallNewCampaign()`.

pledbrook:
@Iridar See [the region links PR](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/929). `InstallNewCampaign` runs too late for that. You also mentioned issue #783, but that probably doesn't need such an early hook.

Iridar (approved):
For what it's worth, everything here seems solid to me.

robojumper (approved):

PR #930, opened 2020-12-25
Fix smoke and poison grenade targeting and effects

### LINKED ISSUE #669: Fix smoke and poison targeting
This would port the LW2 changes to the highlander so that smoke and gas grenade targeting would match the tiles that are actually affected. It would remove the need for Reliable Smoke and its MCO.

pledbrook:
Grenades that apply effects to the world, such as smoke grenades, are unreliable in their targeting because those effects are not always applied to tiles that are highlighted by targeting. This change adds two boolean config vars that, when true, fix that so that smoke and/or poison are applied to all the highlighted tiles.

In addition, there are some instances where the targeting will highlight units without highlighting the tiles that they're on. This gives the impression that those units will be affected by the world effects (like smoke) when they won't. This changes adds a config array of grenade names, where any grenade in that array will not highlight units unless they are on highlighted tiles. Smoke grenades and bombs are the main candidates for this array, but mods may add other types.

Resolves #669

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TargetingMethod_Grenade.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TargetingMethod_Grenade.uc
@@ -1,0 +1,285 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2TargetingMethod_Grenade.uc
+//  AUTHOR:  David Burchanowski  --  8/04/2014
+//  PURPOSE: Targeting method for throwing grenades and other such bouncy objects
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+
+class X2TargetingMethod_Grenade extends X2TargetingMethod
+	native(Core);
+
+var protected XCom3DCursor Cursor;
+var protected XComPrecomputedPath GrenadePath;
+var protected transient XComEmitter ExplosionEmitter;
+var protected bool bRestrictToSquadsightRange;
+var protected XComGameState_Player AssociatedPlayerState;
+
+var bool SnapToTile;
+
+function Init(AvailableAction InAction, int NewTargetIndex)
+{
+	local XComGameStateHistory History;
+	local XComWeapon WeaponEntity;
+	local PrecomputedPathData WeaponPrecomputedPathData;
+	local float TargetingRange;
+	local X2AbilityTarget_Cursor CursorTarget;
+	local X2AbilityTemplate AbilityTemplate;
+
+	super.Init(InAction, NewTargetIndex);
+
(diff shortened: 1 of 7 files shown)


pledbrook:
I don't envisage any issues with Reliable Smoke mod, because it MCOs `X2Effect_SmokeGrenade` and `X2Effect_ApplySmokeGrenadeToWorld`. Yes, it overrides the CHL settings when it does that, but only to do the same thing. It should even benefit from the `GrenadesRequiringUnitsOnTargetedTiles` setting if the player adds grenades to it.

pledbrook:
Save file that demonstrates the issue, if anyone wants it for testing:

[save_BUG Smoke targeting.zip](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/files/5742932/save_BUG.Smoke.targeting.zip)

Just try the targeting with the smoke bomb on the Specialist(?), aiming at various points amid the three soldiers that are close to one another (which includes the Specialist).

To test the fix, add these lines to user config *XComGame.ini*:
```
[XComGame.CHHelpers]
DisableExtraLOSCheckForSmoke=true
DisableExtraLOSCheckForPoison=true
GrenadesRequiringUnitsOnTargetedTiles=SmokeGrenade
GrenadesRequiringUnitsOnTargetedTiles=SmokeGrenadeMk2
GrenadesRequiringUnitsOnTargetedTiles=DenseSmokeGrenade
GrenadesRequiringUnitsOnTargetedTiles=DenseSmokeGrenadeMk2
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TargetingMethod_Grenade.uc
```diff
@@ -216,6 +217,59 @@ function Update(float DeltaTime)
...
+		TargetUnit = XGUnit(TargetActor);
+		if (TargetUnit != none )
+		{
+			kUnitState = TargetUnit.GetVisualizedGameState();
+			if (kUnitState != none)
+			{
+				bFoundTile = false;
+				foreach Tiles(CheckTile)
+				{
+					if (CheckTile == kUnitState.TileLocation)
```
robojumper, line 253: Tiles code is consistently performance-intensive enough for Firaxis to have created the following function in `Helpers.uc` (and a few more):

```
static native function int FindTileInList(TTile Tile, const out array<TTile> List);
```

I think this can definitely come in useful here.

robojumper (changes requested):
Config changes are definitely OK, `CheckForTargetsOnTiles` seems fine with a caveat.

robojumper:
Also this has merge conflicts.

pledbrook:
I've merged @Grobobobo 's changes into the main commit and rebased against `master` to help with a clean merge.

robojumper (approved):

pledbrook:
Can confirm that the fix still seems to be working after the switch to using the native function.

PR #933, opened 2021-01-02
WIP: Build Script Rewrite

robojumper:
Based on #858. Extracts common parts of build scripts to a shared utility class,
updates the script with some recent innovations (utf-8 loc files, cached shader cache),
and uses it for the CHL and DLC2HL builds.


robojumper:
Needs testing (I didn't actually test the ModShaderCache thing yet). If this works out, we should also consider hosting `build_common.ps1` somewhere else so that it would be possible for StratOverhaul/LWotC to use the same build script instead of diverging with every new requirement.

cc @Xymanek

robojumper:
TODO:

- [ ] `extra_globals.uci`
- [ ] Clean for mod dependencies

robojumper:
This is being worked on at https://github.com/X2CommunityCore/X2ModBuildCommon and #991.

PR #939, opened 2021-01-08
Reimplement issue #694

### LINKED ISSUE #694: Add Event OnItemEquipped & OnItemUnEquipped

Iridar:
Implements issue #694, the PR for which was either accidentally reverted or never merged in the first place. 
Documented and tested to be working at least as far as logging equipped/unequipped items on units. 

I got confused a bit, `XCGS_Unit::RemoveItemFromInventory()` marks XComGameState as `optional`, even though I don't see anywhere in the code this function being called without giving it a game state, and I don't particularly understand how it would work anyway. Still, I marked the Game State in the event docs as `optional`, since it technically is.

Closes #694

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,7 +114,6 @@ RunPriorityGroup=RUN_STANDARD
   meet the vanilla game's conditions for it (#633)
 - Triggers the event `OverrideNextRetaliationDisplay` to allow mods to customize and/or enable/disable "next retaliation"
   display in `UIAdventOperations` (#667)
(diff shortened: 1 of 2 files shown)


Iridar:
@robojumper docs / build seems to fail for this PR, and I don't understand why, can you please take a look?

robojumper:
The error is coming from the `ref`. Seems like I forgot that `ref` can also have event documentation in it. I'm going to merge a fix real quick and then try a re-run.

robojumper (approved):

PR #734, opened 2020-01-21
Changes to item upgrade system

GingerAvalanche:
Retry of #732, this time without being a 2 year old branch.

Implements Issue #93 changes
Implements Issue #289 changes

Allows any item type to be upgraded, and allows for upgrade slot handling on a per-state basis.

PR #910, opened 2020-09-13
Add 'none' checks and other fixes to prevent log spam added by TLP (Issue #295)

Iridar:
Props to @robojumper for posting the detailed list of log warnings:

Marked by "x" are the ones that I fixed.
Marked by "?" are the ones I fixed, but my confidence level about added changes is low, and I would appreciate code review on them.
I have not attempted to fix the unmarked ones due to not knowing how to approach the existing code, and I'm probably not gonna fix them, unless I get some directions.

 * [x]  `Accessed None 'SCREENSTACK' XComGame.XComGameState_Item:OnCreation:0120`
 * [x]  `Accessed None 'Ability' XComGame.XGUnit:CreateVisualizer:05B8`
 * [x]  `Accessed None 'FactionState' XComGame.UIPersonnel_SoldierListItem:UpdateData:0BA1`
 * [x]  `Accessed None 'FactionState' XComGame.UISoldierHeader:???:???`
 * [x]  `Accessed None 'EmptyRef' XComGame.X2StrategyElement_DefaultStaffSlots:GetLocationDisplayStringDefault:0022`
 * [x]  `Accessed None 'StateObject_NewState' XComGame.X2Action:AddActionToVisualizationTree:01DF `
 * [x]  `Accessed None 'm_kEntity' XComGame.XGWeapon:Destroyed:000A`
 * [x]  `Accessed None 'ExistingItem' XComGame.XComGameState_LadderProgress:SwapItem:0102`
 * [x]  `Accessed None 'replacementItem' XComGame.UILadderUpgradeScreen:PopulateUpgradePanel:06AA `
 * [ ]  `Accessed None 'MoveBeginAction' XComGame.X2Action:CheckMoveActionForMatch:01FD`
 * [x]  `Accessed array 'Targets' out of bounds (0/0)  XComGame.UITacticalHUD:TargetHighestHitChanceEnemy:0377`
 * [x]  `Accessed array 'UITacticalHUD_Enemies_1.m_arrTargets' out of bounds (0/0) XComGame.UITacticalHUD_Enemies:GetSelectedEnemyStateObjectRef:003D`
 * [x]  `ScriptWarning: Accessed None 'BG' XComGame.UIImageChip:SetWidth:001D`
 * [x]  `Accessed None 'NavHelp' XComGame.UITLE_ChallengeModeMenu:UpdateNavHelp:0247`
 * [ ]  `Accessed None 'ChildPanels' XComGame.UITLE_ChallengeModeMenu:UpdateList:007D`
[...]

Issue #393, opened 2019-01-15
Issue #393 - Add Event Hook to GetClipSize


shiremct:
Add an event hook in XComGameState_Item to allow effects to dynamically modify weapon ammo count.

Iridar:
The PR that was supposed to close this issue has been merged to the live version of the Highlander. It remains open because it needs docs. Please add some :)

PR #941, opened 2021-01-09
Changes to item upgrade system (reupload)

### LINKED ISSUE #93: Streamline NumUpgradeSlots for X2ItemTemplates
To extend upgrade slots to items other then weapons and make them accessible to modders
add `GetNumWeaponUpgradeSlots()` to `X2ItemTemplate`
and use it in all places that currently access `X2WeaponTemplate(Template).NumUpgradeSlots`.

`GetNumWeaponUpgradeSlots` returns  `NumUpgradeSlots` if set (in case of `X2WeaponTemplate`) or reads the value from a helper config array.

Also add a function `GetNumWeaponUpgradeSlots` in `XComGameState_Item` that considers breakthroughs, and adds a way for mods to modify the output.

Pitfalls: Lookout for code that relies on weapons be the only items that use attachements.

### LINKED ISSUE #289: Add an event to UpgradeItems to allow mods to handle upgraded items
Grimy's Loot Mod currently overrides the UpgradeItems function in order to transfer per-gamestate data added by the mod from the old state to the new state. Since I'm updating the mod to use the highlander and stop overriding classes, an event in UpgradeItems, ItemUpgraded, could be added that would allow mods like the loot mod to perform arbitrary handling of items on upgrade. Grimy's Loot Mod reads the old and new state IDs, but it would good for forward compatibility to allow full-scale handling of the items themselves.

Will trigger the event with the old state as the EventSource, and the new state as the EventData. There are code paths within UpgradeItems that do not have an old state, so any mods that wish to use it should remember to check the objects to ensure they exist.

Iridar:
This is a reupload of #734, which was abandoned by the original author. 

Closes #93
Closes #289

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UILadderUpgradeScreen.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UILadderUpgradeScreen.uc
@@ -207,7 +207,8 @@ simulated function PopulateUpgradePanel( int Index, X2LadderUpgradeTemplate Upgr
 					}
 					WeaponTemplate = X2WeaponTemplate(primaryWeaponItem.GetMyTemplate());
 
-					bAddLine = WeaponTemplate != none && WeaponTemplate.NumUpgradeSlots > 0 && ItemUpgradeTemplate.CanApplyUpgradeToWeapon(primaryWeaponItem);
+					// Single line for Issue #93
+					bAddLine = WeaponTemplate != none && primaryWeaponItem.GetNumUpgradeSlots() > 0 && ItemUpgradeTemplate.CanApplyUpgradeToWeapon(primaryWeaponItem);
 
 					//
 					ParamTag.StrValue0 = ClassTemplate.DisplayName;
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2ItemTemplate.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2ItemTemplate.uc
@@ -362,6 +362,41 @@ function bool CanBeLootedByUnit(XComGameState_Item LootItem, XComGameState_Unit 
 	return true;
 }
 
+// Start Issue #93
+/// HL-Docs: feature:NonWeaponUpgradeSlots; issue:93; tags:strategy
+/// Mods can add upgrade slots to items that are not based on X2WeaponTemplate
+/// by adding entries to `NonWeaponUpgradeSlots` config array in `XComGame.ini`, for example:
+/// ```unrealscript
+/// [XComGame.CHHelpers]
+/// +NonWeaponUpgradeSlots = (TemplateName = "KevlarArmor", NumUpgradeSlots = 1)
+/// +NonWeaponUpgradeSlots = (TemplateName = "PlatedArmor", NumUpgradeSlots = 2)
+/// +NonWeaponUpgradeSlots = (TemplateName = "PoweredArmor", NumUpgradeSlots = 3)
+/// ```
+/// Note that if several mods add config entries for the same item template, only the first one one will take effect.
+/// Which one counts as first will depend on the config load order.
+function int GetNumUpgradeSlots()
+{
+	local X2WeaponTemplate SelfAsWeaponTemplate;
+	local int idx;
+
+	SelfAsWeaponTemplate = X2WeaponTemplate(self);
+
+	if (SelfAsWeaponTemplate != none)
+	{
+		return SelfAsWeaponTemplate.NumUpgradeSlots;
+	}
+
+	idx = class'CHHelpers'.default.NonWeaponUpgradeSlots.Find('TemplateName', DataName);
+
+	if (idx != INDEX_NONE)
+	{
+		return class'CHHelpers'.default.NonWeaponUpgradeSlots[idx].NumUpgradeSlots;
+	}
+
+	return 0;
+}
+// End Issue #93
+
 DefaultProperties
 {
 	iItemSize=1;
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
@@ -769,8 +769,8 @@ simulated function bool HasBeenModified()
 		
 	WeaponTemplate = X2WeaponTemplate( m_ItemTemplate );
 
-	// Single line for Issue #306
-	if ((WeaponTemplate != none) && (WeaponTemplate.NumUpgradeSlots > 0) && (GetMyWeaponUpgradeCount() > 0))
+	// Single line for Issues #93 and #306
+	if ((WeaponTemplate != none) && (GetNumUpgradeSlots() > 0) && (GetMyWeaponUpgradeCount() > 0))
 		return true;
 
 	return false;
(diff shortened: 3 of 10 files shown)


Iridar:
This took a while to rebase properly. I kept losing the separate commit for adding new files to the Highlander. To deal with it I had to checkout current master into a separate branch, then add new files to it, in their unchanged form from the base game src.

Then I checked out Ginger's original branch, and rebased it onto the new master branch with the new files.

Iridar:
I have briefly tested these changes and they seem perfectly fine. The original PR was also reviewed by @Musashi1584 and @pledbrook, so I assume this should be good to go.

Iridar:
Person responsible for merging this should probably take a quick look at the new docs to make sure everything is up to snuff.

robojumper:
This PR has merge conflicts, presumably due to #910 adding `UILadderUpgradeScreen.uc` for modification.

Iridar:
I've resolved the conflict and rebased the PR to the latest master branch.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2ItemTemplate.uc
```diff
@@ -362,6 +362,41 @@ function bool CanBeLootedByUnit(XComGameState_Item LootItem, XComGameState_Unit
...
+/// HL-Docs: feature:NonWeaponUpgradeSlots; issue:93; tags:strategy
+/// Mods can add upgrade slots to items that are not based on X2WeaponTemplate
+/// by adding entries to `NonWeaponUpgradeSlots` config array in `XComGame.ini`, for example:
+/// ```unrealscript
+/// [XComGame.CHHelpers]
+/// +NonWeaponUpgradeSlots = (TemplateName = "KevlarArmor", NumUpgradeSlots = 1)
+/// +NonWeaponUpgradeSlots = (TemplateName = "PlatedArmor", NumUpgradeSlots = 2)
+/// +NonWeaponUpgradeSlots = (TemplateName = "PoweredArmor", NumUpgradeSlots = 3)
+/// ```
+/// Note that if several mods add config entries for the same item template, only the first one one will take effect.
```
pledbrook, line 375: Since there's a potential for conflict between mods, should we not provide a means of resolution? My usual go to is the event system, since listener priorities can control the order. The alternative is for one mod to configure itself to run after the other and then modify the config array directly. Although perhaps OPTC is too late for that?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
```diff
@@ -2408,9 +2408,43 @@ function bool IsMissionObjectiveItem()
...
+/// HL-Docs: feature:OverrideNumUpgradeSlots; issue:93; tags:strategy
+/// This event allows mods to modify the number of upgrade slots for each individual item state.
+/// 
+/// ```event
+/// EventID: OverrideNumUpgradeSlots,
+/// EventData: [inout int NumUpgradeSlots],
+///	EventSource: XComGameState_Item (ItemState),
+/// NewGameState: none
+/// ```
+function int GetNumUpgradeSlots()
```
pledbrook, line 2421: Isn't there the potential for the template being out of sync with the item state in this case? Wondering whether template + item state should both have event overrides.

pledbrook (commented):
Not very familiar with weapon mod code, but I've left some questions/comments. In particular, I'm worried that there could easily be discrepancies between a template's number of upgrade slots and an item state's.

X2WOTCCommunityHighlander/X2WOTCCommunityHighlander.x2proj
```diff
@@ -217,6 +217,9 @@
     <Content Include="Src\XComGame\Classes\UIInventory_LootRecovered.uc">
       <SubType>Content</SubType>
     </Content>
+    <Content Include="Src\XComGame\Classes\UILadderUpgradeScreen.uc">
+      <SubType>Content</SubType>
+    </Content>
```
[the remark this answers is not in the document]
Iridar: This is probably because `UILadderUpgradeScreen.uc` that was supposed to be included by this PR has been included by another PR that was merged ahead of this one, which caused a conflict and I messed up fixing that conflict. I'll take a look and get this fixed, good catch.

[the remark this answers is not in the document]
Iridar: Done.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2ItemTemplate.uc
Iridar: Yeah if different mods want the same items to have a different number of upgrade slots there's gonna be a conflict either way, so it's a matter of having tools to deal with it. 

`GetNumUpgradeSlots()` is called at runtime, so `class'CHHelpers'.default.NonWeaponUpgradeSlots` can be edited at runtime by any mod at any time and see the effect immediately, though I'd still use `OnPreCreateTemplates()` to run it as early as possible just in case.

So an event is not absolutely necessary here to help resolve conflicts, and DLC Run Order Control can be used to arbitrate things if several mods want to do it at the same time, though I don't expect that to happen often.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
Iridar: > Isn't there the potential for the template being out of sync with the item state in this case?

Yes, that's kinda the whole point - being able to control upgrade slots on a specific instance without affecting all other items using that template.

There's a [precedent of that kind of thing](https://github.com/shiremct/X2WOTCCommunityHighlander/blob/b538661293230bdc44a9009a2126d570b0fa1385/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc#L930-L942) for #393, and in general that kind of flexibility is what Highlander is all about.

pledbrook: My concern was that most code seems to check the template for number of item slots, not the item state.

pledbrook: OK, just to clarify in my head what's going on, since there are two different things being added in this PR it seems. The number of weapon upgrade slots is tied to the weapon template still, correct? Only if the item is something other than a weapon does the item state get to override the number of upgrade slots? If that's the case, then yes, this seems fine. I can't reasonably say whether all code that fetches the number of upgrade slots has been updated, though. Are we sure there aren't any instances in native?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
```diff
@@ -769,8 +769,8 @@ simulated function bool HasBeenModified()
 		
 	WeaponTemplate = X2WeaponTemplate( m_ItemTemplate );
 
-	// Single line for Issue #306
-	if ((WeaponTemplate != none) && (WeaponTemplate.NumUpgradeSlots > 0) && (GetMyWeaponUpgradeCount() > 0))
+	// Single line for Issues #93 and #306
+	if ((WeaponTemplate != none) && (GetNumUpgradeSlots() > 0) && (GetMyWeaponUpgradeCount() > 0))
```
pledbrook, line 773: Is this change necessary since if this is a weapon, then the template's `NumUpgradeSlots` is used anyway? If the intention is that the implementation of `GetNumUpgradeSlots()` might change, then shouldn't the ladder-related classes in this PR use `Item.GetNumUpgradeSlots()` rather than `WeaponTemplate.GetNumUpgradeSlots()`?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
Iridar: By default, weapon templates will use number of weapon upgrade slot written in the template.

Non-weapon templates will draw the default value of upgrade slots from config.

XCGS_Item instances of both weapons and non-weapons can additionally use the override event to conditionally change the number of upgrade slots. 

As such, `.GetNumUpgradeSlots()` should be used from the Item State whenever one is available.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
pledbrook: Gah, getting confused between the version of the function on the template and this one.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_LadderProgress.uc
```diff
@@ -1120,7 +1120,7 @@ private static function UpdateUnitState( XComGameState StartState, XComGameState
 	PrimaryWeapon = UnitState.GetPrimaryWeapon( );
 	PrimaryWeapon = XComGameState_Item( StartState.GetGameStateForObjectID( PrimaryWeapon.ObjectID ) );
 	WeaponTemplate = X2WeaponTemplate( PrimaryWeapon.GetMyTemplate( ) );
-	if (WeaponTemplate.NumUpgradeSlots > 0)
+	if (WeaponTemplate.GetNumUpgradeSlots() > 0) // Single line for Issue #93
```
pledbrook, line 1123: This is the last remaining are of puzzlement for me: why is this not using `XCGS_Item.GetNumUpgradeSlots()`, since the item state is available?

X2WOTCCommunityHighlander/Src/XComGame/Classes/UILadderUpgradeScreen.uc
```diff
@@ -207,7 +207,8 @@ simulated function PopulateUpgradePanel( int Index, X2LadderUpgradeTemplate Upgr
 					}
 					WeaponTemplate = X2WeaponTemplate(primaryWeaponItem.GetMyTemplate());
 
-					bAddLine = WeaponTemplate != none && WeaponTemplate.NumUpgradeSlots > 0 && ItemUpgradeTemplate.CanApplyUpgradeToWeapon(primaryWeaponItem);
+					// Single line for Issue #93
+					bAddLine = WeaponTemplate != none && WeaponTemplate.GetNumUpgradeSlots() > 0 && ItemUpgradeTemplate.CanApplyUpgradeToWeapon(primaryWeaponItem);
```
pledbrook, line 211: Also not sure why this isn't using `primaryWeaponItem.GetNumUpgradeSlots()`.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_LadderProgress.uc
Iridar: Good question. What does Ladder Progress file even do?

Iridar:
Updated the PR to address @pledbrook's concerns.

pledbrook (approved):

Iridar:
@GingerAvalanche 
We did, sempai.

Issue #420, opened 2019-02-11
Add UISquadSelect_NavHelpUpdate event


Xymanek:
To allow to gracefully update nav help by other mods. Eg. add buttons

rj's UISS version: robojumper/robojumperSquadSelect#8

Iridar:
Beats using a UISL. I'll get on this.

PR #942, opened 2021-01-10
Add UISquadSelect_NavHelpUpdate Event

Iridar:
Closes issue #420

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UISquadSelect.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UISquadSelect.uc
@@ -909,6 +909,21 @@ simulated function UpdateNavHelp()
 		}
 		`endif
 
+		// Start Issue #420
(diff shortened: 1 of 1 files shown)


Iridar:
Briefly tested this with Prototype Armory, which listens to this event (expecting it from RJSS) to add a "remove weapon upgrades" button. Seems to work fine.

pledbrook (approved):
Looks good to me!

Iridar:
One thing I'm not entirely sure about is that I have purposefully put the event trigger before `NavHelp.Show();`, so that any changes by mods are seamless. However, [in RJSS](https://github.com/Xymanek/robojumperSquadSelect/blob/5e066086e3ea3a054dc0d13a91df5c9f2d18cba9/robojumperSquadSelect/Src/robojumperSquadSelect/Classes/robojumper_UISquadSelect.uc#L852) there is no line like that in `UpdateNavHelp()` at all. So not sure what to make of it.

robojumper (approved):
Order of event and `Show()` shouldn't really matter; flash calls are queued and flushed at the end of the frame anyway.

PR #947, opened 2021-01-12
Add X2Condition mapping decorator and inverse condition

### LINKED ISSUE #699: Add Inverse Boolean to X2Condition
When Inverse is True any Value that would have been returned will be Inverted first.

So "AA_Success" would become "AA_AbilityUnavailable" and vice versa.

Extremely useful when working with Abilities.

Probably would have to be an intermediate for X2Condition.MeetsCondition and whatever native stuff goes between it and CallMeetsCondition

I don't know anything about the native stuff tho.

Iridar:
Closes #699

Adds an `X2Condition_MappingDecorator` class that stores an `X2Condition` and remaps its return codes according to specified remapping data.

Adds `X2Condition_ReverseCondition`, which is an extension of the previous class, preconfigured for simply reversing the return codes of the inlaid condition.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_MappingDecorator.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_MappingDecorator.uc
@@ -35,12 +35,18 @@ private function name RemapCondition(const name OriginalCode)
 {
 	local int Index;
 
+	`LOG("Remapping Condition:" @ InnerCondition.Class.Name @ "original return code:" @ OriginalCode,, 'IRITEST');
+
 	Index = MappingDecorations.Find('OriginalCode', OriginalCode);
 
 	if (Index != INDEX_NONE)
 	{
+		`LOG("Found remapped code:" @ Index @ ":" @ MappingDecorations[Index].ReplacementCode,, 'IRITEST');
 		return MappingDecorations[Index].ReplacementCode;
 	}
+
+	`LOG("Did not find remapped code, defaulting to:" @ UnmappedCode == '' ? OriginalCode : UnmappedCode,, 'IRITEST');
+
 	return UnmappedCode == '' ? OriginalCode : UnmappedCode;
 }
 
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_MappingDecorator.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_MappingDecorator.uc
@@ -52,17 +52,17 @@ private function name RemapCondition(const name OriginalCode)
 
 event name CallMeetsCondition(XComGameState_BaseObject kTarget) 
 {
-	return RemapCondition(InnerCondition.CallMeetsCondition(kTarget));
+	return RemapCondition(InnerCondition.MeetsCondition(kTarget));
 }
 
 event name CallMeetsConditionWithSource(XComGameState_BaseObject kTarget, XComGameState_BaseObject kSource) 
 { 
-	return RemapCondition(InnerCondition.CallMeetsConditionWithSource(kTarget, kSource));
+	return RemapCondition(InnerCondition.MeetsConditionWithSource(kTarget, kSource));
 }
 
 event name CallAbilityMeetsCondition(XComGameState_Ability kAbility, XComGameState_BaseObject kTarget) 
 {
-	return RemapCondition(InnerCondition.CallAbilityMeetsCondition(kAbility, kTarget));
+	return RemapCondition(InnerCondition.AbilityMeetsCondition(kAbility, kTarget));
 }
 
 function bool CanEverBeValid(XComGameState_Unit SourceUnit, bool bStrategyCheck)
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_MappingDecorator.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_MappingDecorator.uc
@@ -1,0 +1,82 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2Condition_MappingDecorator.uc
+//  AUTHOR:  X2CommunityCore
+//  PURPOSE: Remap return codes of a provided X2Condition.
+//---------------------------------------------------------------------------------------
+class X2Condition_MappingDecorator extends X2Condition;
+
+struct MappingDecorationStruct
+{
+	var name OriginalCode;
+	var name ReplacementCode;
+};
+
+var X2Condition InnerCondition;
+
+var bool bUseCanEverBeValid;
+var bool bReverseCanEverBeValid;
+
+// If specified, will be used as the default return code for return codes that do not have mapping decoration config.
+var name UnmappedCode;
+
+var array<MappingDecorationStruct> MappingDecorations;
+
+function AddMappingDecoration(const name OriginalCode, const name ReplacementCode)
+{
+	local MappingDecorationStruct NewMappingDecoration;
+
+	NewMappingDecoration.OriginalCode = OriginalCode;
+	NewMappingDecoration.ReplacementCode = ReplacementCode;
+
+	MappingDecorations.AddItem(NewMappingDecoration);
+}
+
+private function name RemapCondition(const name OriginalCode)
+{
+	local int Index;
+
+	Index = MappingDecorations.Find('OriginalCode', OriginalCode);
+
+	if (Index != INDEX_NONE)
+	{
+		return MappingDecorations[Index].ReplacementCode;
+	}
+	return UnmappedCode == '' ? OriginalCode : UnmappedCode;
+}
+
+event name CallMeetsCondition(XComGameState_BaseObject kTarget) 
+{
+	return RemapCondition(InnerCondition.CallMeetsCondition(kTarget));
+}
+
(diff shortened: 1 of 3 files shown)


Iridar:
This PR should have been a draft, it has no docs yet, and I don't think it even works. We also should probably have a discussion on names of properties and methods of the `X2Condition_MappingDecorator`, as well as the name of the class itself, because personally `mapping decorator` rings completely hollow for me, and associates with map design at best.

Iridar:
Yeah, this doesn't work, although I don't understand why. The InnerCondition seems to always return `AA_Success` regardless of how I set it up.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_MappingDecorator.uc
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	`LOG("Did not find remapped code, defaulting to:" @ UnmappedCode == '' ? OriginalCode : UnmappedCode,, 'IRITEST');
+
+	return UnmappedCode == '' ? OriginalCode : UnmappedCode;
+}
+
+event name CallMeetsCondition(XComGameState_BaseObject kTarget) 
+{
+	return RemapCondition(InnerCondition.CallMeetsCondition(kTarget));
```
robojumper, line 55:
```suggestion
	return RemapCondition(InnerCondition.MeetsCondition(kTarget));
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_MappingDecorator.uc
```diff
@@ -0,0 +1,88 @@
...
+}
+
+event name CallMeetsConditionWithSource(XComGameState_BaseObject kTarget, XComGameState_BaseObject kSource) 
+{ 
+	return RemapCondition(InnerCondition.CallMeetsConditionWithSource(kTarget, kSource));
```
robojumper, line 60:
```suggestion
	return RemapCondition(InnerCondition.MeetsConditionWithSource(kTarget, kSource));
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_MappingDecorator.uc
```diff
@@ -0,0 +1,88 @@
...
+}
+
+event name CallAbilityMeetsCondition(XComGameState_Ability kAbility, XComGameState_BaseObject kTarget) 
+{
+	return RemapCondition(InnerCondition.CallAbilityMeetsCondition(kAbility, kTarget));
```
robojumper, line 65:
```suggestion
	return RemapCondition(InnerCondition.AbilityMeetsCondition(kAbility, kTarget));
```

robojumper (changes requested):
I believe the problem here is that you're unconditionally using the UnrealScript implementation of `InnerCondition` even if it's a native condition. For reference, this is in `X2Condition`:

```unrealscript
native function name MeetsCondition(XComGameState_BaseObject kTarget);
native function name MeetsConditionWithSource(XComGameState_BaseObject kTarget, XComGameState_BaseObject kSource);
native function name AbilityMeetsCondition(XComGameState_Ability kAbility, XComGameState_BaseObject kTarget);

/*
 * These should ONLY be called from the native functions above.
 */
event name CallMeetsCondition(XComGameState_BaseObject kTarget) { return 'AA_Success'; }
event name CallMeetsConditionWithSource(XComGameState_BaseObject kTarget, XComGameState_BaseObject kSource) { return 'AA_Success'; }
event name CallAbilityMeetsCondition(XComGameState_Ability kAbility, XComGameState_BaseObject kTarget) { return 'AA_Success'; }
```

The general idea is that the native (or [denativized](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/918)) `XCGS_Ability:GatherAbilityTargets` calls the native `MeetsCondition`/`MeetsConditionWithSource`/`AbilityMeetsCondition`. If the `X2Condition` subclass in question is implemented with native code, this allows evaluating the condition without ever leaving C++ and calling into UnrealScript.

Only if the `X2Condition` subclass doesn't override a function, the base implementation in `X2Condition` calls the respective `Call*` variant of the function, which allows UnrealScript code to do things.

Result: You should make these three changes :)

Iridar:
Oh snap. Looks like you already guarded me against this error for [this wiki article](https://www.reddit.com/r/xcom2mods/wiki/index/conditions#wiki_using_conditions_inside_conditions). Thanks!

Iridar:
Looks like this still doesn't work for a typical expected use case. 

I made a copy of standard shot as a test ability, removed all other conditions from it, and then did this:
```
	ReverseCondition = new class'X2Condition_ReverseCondition';

	UnitProperty = new class'X2Condition_UnitProperty';
	UnitProperty.ExcludeLargeUnits = true;
	UnitProperty.ExcludeDead=false;
	UnitProperty.ExcludeFriendlyToSource=false;
	UnitProperty.ExcludeCosmetic=false;
	UnitProperty.ExcludeInStasis=false;

	ReverseCondition.InnerCondition = UnitProperty;
	ReverseCondition.bUseCanEverBeValid = false;

	Template.AbilityTargetConditions.AddItem(ReverseCondition);
```
Essentially, I'm adding a unit property condition that - if used normally - would allow you to target only regular-sized units, e.g. can target soldiers and troopers, but not large units, like Sectopods. This does work as expected when I add the UnitProperty condition directly, without reversing it.

So when reversed you would expect to be able to target *only* Sectopods. However, in testing the ability cannot target *any* units. The logs from it are:
```
[0043.26] IRITEST: Remapping Condition: X2Condition_UnitProperty original return code: AA_UnitIsTooLarge
[0043.26] IRITEST: Did not find remapped code, defaulting to: AA_Success
[0043.26] IRITEST: Remapping Condition: X2Condition_UnitProperty original return code: AA_Success
[0043.26] IRITEST: Found remapped code: 0 : AA_AbilityUnavailable
[0043.26] IRITEST: Remapping Condition: X2Condition_UnitProperty original return code: AA_Success
[0043.26] IRITEST: Found remapped code: 0 : AA_AbilityUnavailable
[0043.26] IRITEST: Remapping Condition: X2Condition_UnitProperty original return code: AA_Success
[0043.26] IRITEST: Found remapped code: 0 : AA_AbilityUnavailable
[0043.26] IRITEST: Remapping Condition: X2Condition_UnitProperty original return code: AA_Success
[0043.26] IRITEST: Found remapped code: 0 : AA_AbilityUnavailable
[0043.26] IRITEST: Remapping Condition: X2Condition_UnitProperty original return code: AA_Success
[0043.26] IRITEST: Found remapped code: 0 : AA_AbilityUnavailable
[0043.26] IRITEST: Remapping Condition: X2Condition_UnitProperty original return code: AA_Success
[0043.26] IRITEST: Found remapped code: 0 : AA_AbilityUnavailable
```
As we can see, the remapped condition does return `AA_Success` once, presumably when the condition is checked against the Sectopod I have dropped in addition to a Trooper, an allied Muton and an allied SPARK with BIT and a Templar.

My only guess here is what's happening is that the property condition returns `AA_UnitIsTooLarge` in `UnitProperty.MeetsCondition()`, but returns `AA_Success` in `UnitProperty.MeetsConditionWithSource()`, which then gets remapped to `AA_AbilityUnavailable`.

So it seems we have a design issue on our hands when dealing with conditions that implement several checks at the same time.

We either have to remap them separately or opt out of remapping them. Both would complicate things for the user. 

This issue was originally supposed to make reversing conditions easier, but if this thing is gonna require a doctorate in programming to use properly, it might not be worth the effort. Further discussion on the topic is welcome, because I'm not sure what to do at this point.

robojumper:
@Iridar Oof, that's a pretty unfortunate oversight indeed. On a more abstract level, an ability is available for a unit and target if and only if all operators of all conditions return `AA_Success`. This is conceptually a logical conjunction of the condition results, each of which is a conjunction of the return values of `MeetsCondition`, `MeetsConditionWithSource`, `AbilityMeetsCondition`. We'd like to invert one condition, so that's basically:

`!(Condition.MeetsCondition(...) && Condition.MeetsConditionWithSource(...) && Condition.AbilityMeetsCondition(...))`

Since the overall condition checking happens in native code, we have no way to invert the result of the conjunction. We could break the term down further:

`(!Condition.MeetsCondition(...) || !Condition.MeetsConditionWithSource(...) || !Condition.AbilityMeetsCondition(...))`

But that runs afoul of the exact same issue -- we can't simply use a disjunction because it's native code. The only way is -- as you said -- "remap [the operators] separately".

> So it seems we have a design issue on our hands when dealing with conditions that implement several checks at the same time.

I'm not sure what you mean. Every condition always implements the three operators and returns `AA_Success` unless it has a reason not to. A "is not large" check turns `Condition.MeetsCondition(...) && Condition.MeetsConditionWithSource(...) && Condition.AbilityMeetsCondition(...)` into `Unit.NotLarge() && true && true` because the operators yielding an unconditional `true` are fundamentally a "don't care" that our operators then turn into a `false`.

The "don't care" thing is a new insight for me. I guess the only way to implement a system that supports these decorators is to introduce three-valued logic with a "AA_DontCare" or something.
The "remapping separately" comes close. By remapping operators selectively, we essentially turn the unconditional `AA_Success` into a "don't care", but as you said, this requires insight into the internal workings of the decorated condition (most of which are native). At this point I think this is pretty hard to get right and custom conditions should be preferred. Recommending closing the PR and issue as infeasible.

Iridar:
Closing per robojumper's recommendation.

PR #948, opened 2021-01-12
Add PostInventoryLoadoutApplied event

### LINKED ISSUE #800: Add an event to XCGS_Unit::ApplyInventoryLoadout
Add an 'OnInventoryLoadoutApplied' event to XCGS_Unit::ApplyInventoryLoadout so that mods can change Unit's loadout after it has been applied. I want to use this for a mod that changes a rookie's starting primary weapon depending on their nationality. Could be used by any other mods that want to dynamically change a soldier's starting loadout.

Iridar:
Implements #800 (closed)

Draft until I perform some testing.


Iridar:
This PR adds events to the end of `XCGS_Unit::ApplyInventoryLoadout` as #800 suggested, and also to `XCGS_Unit::ApplySquaddieLoadout()` for good measure.

All very standard, but the joke is that these events seem to be completely useless for the use case outlined in #800 - making changes to **rookies'** equipment, as it appears they never fire for rookies. Not for the ones you get on Gatecrasher, not for ones you get when you return from it, not for ones you recruit from Avenger.

This PR should be good to go in terms of implementation/docs (although docs might require correction), but I don't see a reason to merge it until somebody has an actual use case for it.

robojumper (approved):

PR #949, opened 2021-01-13
Add PostRollForTrainingCenterAbilities event

### LINKED ISSUE #815: Conditionally block cross-class perks for specific soldiers
Would be nice if there was a way to disallow certain cross-class perks from appearing on certain soldiers.

For example, I have made a number of cool melee abilities that could potentially benefit a class like a vanilla Ranger, but I can't make them universally cross class on the likely chance that they would get rolled by a class that doesn't use melee at all. 

I'm willing to do something about this myself, but a bit stressed for time currently, so opening the issue mostly so I don't forget.

Iridar:
Closes #815

I have tested this to be working fine for a typical use case, although for some reason the event is triggered twice for each soldier. If I make changes to the `AbilityTree` the first time, the changes are reflected when the event is triggered the second time. This should probably be reflected in the docs.

Here's log output from a quick dumb test that replaces `RapidFire` with `Reaper` in the `AbilityTree` row with the Training Center abilities:
```
[0089.94] IRITEST: PostRollForTrainingCenterAbilities triggered for: Lucy Wood
[0089.94] IRITEST: Lucy Wood 0 RapidFire
[0089.94] IRITEST: Replacing Rapid Fire with Reaper
[0089.94] IRITEST: Lucy Wood 1 BulletShred
[0089.94] IRITEST: PostRollForTrainingCenterAbilities triggered for: Lucy Wood
[0089.94] IRITEST: Lucy Wood 0 Reaper
[0089.94] IRITEST: Lucy Wood 1 BulletShred
```
Also, somewhat unconventionally, the event trigger passes `none` as `EventData`; the event has nothing to put there at this time. 
I was thinking we could keep it `none` for now, and potentially put something into it later if we have a use for it.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -3702,7 +3702,7 @@ function RollForTrainingCenterAbilities()
 		///```event,notemplate
 		/// EventID: PostRollForTrainingCenterAbilities,
 		/// EventData: none,
-		/// EventSource: inout XComGameState_Unit (UnitState),
+		/// EventSource: XComGameState_Unit (UnitState),
 		/// NewGameState: none
 		///```
 		/// Example code:
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -3689,43 +3689,43 @@ function RollForTrainingCenterAbilities()
 				EligibleAbilities.Remove(AbilityIdx, 1); // Remove the ability which was chosen so it won't get picked again
 			}
 		}
-	}
 
-	/// HL-Docs: feature:PostRollForTrainingCenterAbilities; issue:815; tags:strategy
-	/// The `PostRollForTrainingCenterAbilities` event allows mods to make arbitrary changes
-	/// to a Soldier after their Training Center abilities have been rolled.
-	///
-	/// Typicaly this should be used to add or remove Training Center abilities
-	/// from the Soldier. To do so you have to access `UnitState.AbilityTree` directly. 
-	/// If the Soldier has rolled any Training Center abilities, 
-	/// they will be in the very last Rank Index.
-	///
-	///```event
-	/// EventID: PostRollForTrainingCenterAbilities,
-	/// EventData: none,
-	/// EventSource: XComGameState_Unit (UnitState),
-	/// NewGameState: none
-	///```
-	/// Example code:
-	///```unrealscript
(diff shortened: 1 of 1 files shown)


robojumper (approved):

robojumper (changes requested):
Actually, I think you should move the `TriggerEvent` call to inside of the `if` block. `RollForTrainingCenterAbilities` is idempotent and can be called any number of times, and only does work the first time, so the event should only be triggered once.

Iridar:
> Actually, I think you should move the `TriggerEvent` call to inside of the `if` block. `RollForTrainingCenterAbilities` is idempotent and can be called any number of times, and only does work the first time, so the event should only be triggered once.

Yeah, that makes a lot of sense. Doneso. Also added `notemplate` to the event docs, since my docs already have an example. Is my usage correct?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -3692,6 +3692,42 @@ function RollForTrainingCenterAbilities()
...
+		///
+		/// Typicaly this should be used to add or remove Training Center abilities
+		/// from the Soldier. To do so you have to access `UnitState.AbilityTree` directly. 
+		/// If the Soldier has rolled any Training Center abilities, 
+		/// they will be in the very last Rank Index.
+		///
+		///```event,notemplate
+		/// EventID: PostRollForTrainingCenterAbilities,
+		/// EventData: none,
+		/// EventSource: inout XComGameState_Unit (UnitState),
```
robojumper, line 3708:
```suggestion
		/// EventSource: XComGameState_Unit (UnitState),
```

Iridar:
Right, sorry, I forgot that we do `inout` stuff only for the Tuple now.

Iridar:
On discord, robojumper has raised a rightful concern that for event listeners it would be hard to determine which of the perks in the `AbilityTree` are Training Center, and which have been there before. We discussed a couple of options of accomplishing that, such as going back to history and examining the prior Unit State, but I just realized two tings:

1) Since training center abilities are generated right then and there, just before the event trigger, we could keep a list of the training center perks that were added, and pass it to the event trigger. Although it's an array of `SoldierClassAbilityType` structs, so passing them to the event trigger will require some sort of hack, like creating a copy of the Unit State.

2) At that point, wouldn't it be better to trigger the event after the list of training center perks has been generated, but before they are inserted into the ability tree? So that mods can alter the list of the eligible training center perks instead of the Ability Tree. This also removes [this concern](https://www.reddit.com/r/xcom2mods/wiki/index/events#wiki_important_note), and if mods want to specifically alter the AbilityTree *after* the perks have already been added to the AbilityTree, they can listen to the same event with `ELD_OnStateSubmitted` instead.

Iridar:
Closing, implemented differently in another PR.

PR #951, opened 2021-01-13
Allow mods to stop civilians running from units

### LINKED ISSUE #666: Civilians run from patrolling aliens
See [lwotc#705](https://github.com/long-war-2/lwotc/issues/705) for an example of this. It's not really a problem for the base game, but in LWOTC and RedDobe's Yellow Alert Gameplay mod, this will put the enemy units in sound range into yellow alert.

pledbrook:
I've added a 'ShouldCivilianRun' event that allows mods to determine whether a given civilian unit should run from another unit that moved close to it. The default behaviour is for the civilian to run.

The event takes the form:
```
  ID: ShouldCivilianRun,
  Data: [in XCGS_Unit MovedUnit, in bool AIAttacksCivilians, out bool ShouldRun],
  Source: Unit (Civilian)
```

The `ShouldRun` parameters defaults to `true`.

Resolves #666.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
@@ -1654,6 +1654,15 @@ function EventListenerReturn HandleNeutralReactionsOnMovement(Object EventData, 
 			continue;
 		}
 
+		// Start Issue #666
+		//
+		// Let mods have their say on whether this civilian should run from MovedUnit.
+		if (!TriggerShouldCivilianRun(CivilianState, MovedUnit, bAIAttacksCivilians))
+		{
+			continue;
+		}
+		// End Issue #666
+
 		// Non-rescue behavior is kicked off here.
 		if( class'Helpers'.static.IsTileInRange(CivilianState.TileLocation, MovedToTile, MaxTileDistSq) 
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
```diff
@@ -1674,6 +1683,41 @@ function EventListenerReturn HandleNeutralReactionsOnMovement(Object EventData,
...
 
+// Start Issue #666
+/// HL-Docs: feature:ShouldCivilianRun; issue:666; tags:tactical
+/// Triggers a 'ShouldCivilianRun' event that allows listeners to determine whether
+/// neutrals should run in reaction to another unit moving by them. Returns `true`
+/// if the given neutral unit should run, `false` otherwise.
+///
+/// ```event
+/// EventID: ShouldCivilianRun,
+/// EventData: [in XComGameState_Unit MovedUnit, in bool AIAttacksCivilians, out bool ShouldRun],
```
robojumper, line 1694:
```suggestion
/// EventData: [in XComGameState_Unit MovedUnit, in bool AIAttacksCivilians, inout bool ShouldRun],
```

A side effect of the way CHL docs generate listener templates is that `out` causes an assignment with a default-initialized value (false), so the unmodified listener template would cause a change in behavior. Either mark this `inout` or use a logic where the default parameter is false ("ShouldNotRun"?)

robojumper (changes requested):

robojumper (approved):

PR #954, opened 2021-01-13
Add ModifyMaxSquadSize DLC hook

robojumper:
I personally don't plan to publish any mod depending on this, but I think this could come in helpful for a mod that gives more control over squad size overall (base size, squad size upgrades, special missions).

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
@@ -664,6 +664,51 @@ static function UnitPawnPostInitAnimTree(XComGameState_Unit UnitState, XComUnitP
 }
 /// End Issue #455
 
+// Start Issue #953
+/// HL-Docs: feature:ModifyMaxSquadSize; issue:953; tags:
+/// The DLC hook `ModifyMaxSquadSize` is called from `X2StrategyGameRulesetDataStructures:GetMaxSoldiersAllowedOnMission`. It gives
+/// mods a chance to modify the components that make up the maximum squad size:
+///
+/// * `MissionDefLimit`: `MaxBaseSize` was imposed by the mission definition
+/// * `MaxBaseSize`: Base max squad size
+/// * `SquadSizeUpgradeMod`: Modifier from GTS squad size purchases. `0` if `MissionDefLimit == true`
+/// * `SituationMod`: Modifiers from SITREPs and the intel purchase (network tower mission)
+/// * `SitRepMax`: Maximum squad size imposed by SITREPs. Will be `MaxInt` if no SITREP limits squad size (none of the vanilla SITREPs do)
+///
+/// Example usage:
+///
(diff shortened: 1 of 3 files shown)


Xymanek:
Why a DLCInfo hook and not an event?

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -664,6 +664,51 @@ static function UnitPawnPostInitAnimTree(XComGameState_Unit UnitState, XComUnitP
...
+/// 	{
+/// 		// Leviathan gets four more soldiers because reasons
+/// 		SituationMod += 4;
+/// 	}
+/// }
+/// ```
+///
+/// The final max squad size will be calculated with the expression `Min(Max(1, MaxBaseSize + SquadSizeUpgradeMod + SituationMod), SitRepMax)`.
+///
+/// This hook is best combined with `GetValidFloorSpawnLocations` or `SPAWN_EXTRA_TILE` (TODO: Docs for these) in order to support
```
Iridar, line 704: I'm not sure we'd want to ship a "TODO", but I assume this is here just because it's a Draft PR? It is marked as "ready for review", so I'm getting some mixed signals here.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc
```diff
@@ -664,6 +664,51 @@ static function UnitPawnPostInitAnimTree(XComGameState_Unit UnitState, XComUnitP
...
+/// 	XComHQ = `XCOMHQ;
+/// 	if (`XCOMHISTORY.GetCurrentHistoryIndex() > -1 && XComHQ != none) // cargo-culted from original code
+///     {
+///     	if (XComHQ.SoldierUnlockTemplates.Find('SquadSizeIIIUnlock') != INDEX_NONE) { SquadSizeUpgradeMod += 1; }
+///     	if (XComHQ.SoldierUnlockTemplates.Find('SquadSizeIVUnlock') != INDEX_NONE) { SquadSizeUpgradeMod += 1; }
+///     }
+/// 
+/// 	if (MissionSite != none && MissionSite.GeneratedMission.Mission.MissionName == 'AssaultFortressLeadup')
+/// 	{
+/// 		// Leviathan gets four more soldiers because reasons
```
Iridar, line 696: Personally I had no trouble following these docs and I find the provided explanations perfectly sufficient, but perhaps they should be a bit lighter on the joke-ish wordings like "cargo cult" and "because reasons".

Iridar (approved):
I don't have strong objections to anything in this PR, and a hook like that is no doubt gonna come in handy on many occasions, although same as Xymanek, I wonder if perhaps making this a delegate method (like the one Xym and I made for #889) would have been better.

Although on second thought, delegates have more cumbersome setup, while mission squad size is not particularly hot code, so it's probably okay to have it as a DLC hook.

Issue #524, opened 2019-05-21
Allow mods to define required and incompatible mod dependencies


Musashi1584:
I suggest adding two new functions to the DLCInfo

```
function array<string> GetIncompatibleDLCIdentifiers();
function array<string> GetRequiredDLCIdentifiers();
```

These can be overriden by mods DLCInfo classes to define dependencies and incompatiblities.
The would be then checked in a shell screen listener that displays the results to the player-

Xymanek:
Why is this still open?

PR #955, opened 2021-01-13
Mods can specify required Highlander version in config

### LINKED ISSUE #909: Expand CHModDependency functionality to check the required Highlander version for each mod individually
... and display a warning message in-game if a mod requires highlander version higher than currently installed.

Iridar:
Closes #909

Localized descriptions are placeholder and we should probably have a discussion about them. Once we finalize them, I can get a person to translate them (and other localized Highlander text) to Russian. 

I didn't commentate the code because it's relatively straightforward and heavily based on Musashi's implementation of #524.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_ModDependencies.uc
+++ b/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_ModDependencies.uc
@@ -17,6 +17,14 @@ var localized string ModRequiredPopupTitle;
 var localized string ModIncompatiblePopupTitle;
 var localized string DisablePopup;
 
+// Begin Issue #909
+var localized string ModRequiresNewerHighlanderVersionTitle;
+var localized string ModRequiresNewerHighlanderVersionText;
+var localized string CurrentHighlanderVersionTitle;
+var localized string RequiredHighlanderVersionTitle;
+var localized string ModRequiresNewerHighlanderVersionExtraText;
+// End Issue #909
+
 static function array<ModDependency> GetRequiredMods()
 {
 	local array<X2DownloadableContentInfo> DLCInfos;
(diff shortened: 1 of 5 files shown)


X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_ModDependencies.uc
```diff
@@ -174,4 +182,41 @@ function static string Join(array<string> StringArray, optional string Delimiter
...
+	CurrentCHLVersion.MajorVersion = class'CHXComGameVersionTemplate'.default.MajorVersion;
+	CurrentCHLVersion.MinorVersion = class'CHXComGameVersionTemplate'.default.MinorVersion;
+	CurrentCHLVersion.PatchVersion = class'CHXComGameVersionTemplate'.default.PatchVersion;
+}
+
+static private function bool IsCurrentCHLVersionOlderThanRequired(const out CHLVersionStruct CurrentCHLVersion, const out CHLVersionStruct RequiredCHLVersion)
+{
+    return CurrentCHLVersion.MajorVersion < RequiredCHLVersion.MajorVersion ||
+           CurrentCHLVersion.MinorVersion < RequiredCHLVersion.MinorVersion ||
+           CurrentCHLVersion.PatchVersion < RequiredCHLVersion.PatchVersion;
```
robojumper, line 220: I'm afraid this is not correct. If current were `1.22.0` and required is at least `1.21.1`, then this would say the HL is outdated.

Iridar: Good eye, I'll get that fixed.

Iridar:
I've added a separate commit that makes `DisplayName` in `CHModDependency` a `config` var. It appears this was an oversight (bug) in the initial implementation. The comment mentions this string being "equivalent to empty if not specified in localization", but the commented out code example assigns `DisplayName` in config, right along other config variables. This is also where I've been assigning "DisplayName" for all of my mods that used this feature while being unaware of this bug. So I've adjusted the comment as well.

robojumper:
Status: This will almost certainly be on the receiving end of a merge conflict due to #966, which I want to merge first (after Musashi gives it a quick look). Assigning waiting-on-author because of that.

pledbrook:
It seems unlikely that Musashi will be reviewing that DLC order PR any time soon, so should this PR really wait on that review? Perhaps we need to review how we're going to deal with the mod dependency and DLC run order PRs.

Iridar:
Replaced by #1007.

PR #958, opened 2021-01-14
Make TypicalAbility_BuildVisualization get AbilityState and SourceWeapon from VisualizeGameState when possible

### LINKED ISSUE #879: Vanilla Xcom2 is looking at weapon's most recent Game State, rather than the Game State where the ability was activated
**PROBLEM :**

the issue is that when visualizing weapon projectile shots, and additional effects applied from an X2AmmoTemplate, the game is looking at weapon's most recent Game State, rather than the Game State where the ability was activated.

This means that if special ammo was removed from the weapon after the ability has gone through, but before its visualization starts, the additional target effects from the Ammo Template, and additional projectile effects from the Ammo Template will not be visualized.

It's a visual bug only.

when the game is visualizing stuff, it should be looking at the state objects as they were in the game state that is being visualized right now.

Iridar:
Closes #879

I've added `none` checks and made the code fallback on getting these state objects from history if the code fails to retrieve them from the `VisualizeGameState` - presumably because the `BuildGameStateFn()` did not add them to the Game State for whatever reason. 

So best case scenario, this fixes the bug, and worst case scenario, fallbacks on vanilla behavior anyway.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability.uc
@@ -765,7 +765,9 @@ static function TypicalAbility_BuildVisualization(XComGameState VisualizeGameSta
 	VisualizationMgr = `XCOMVISUALIZATIONMGR;
 	Context = XComGameStateContext_Ability(VisualizeGameState.GetContext());
 	AbilityContext = Context.InputContext;
-	AbilityState = XComGameState_Ability(History.GetGameStateForObjectID(AbilityContext.AbilityRef.ObjectID));
+	/// HL-Docs: ref:Bugfixes; issue:879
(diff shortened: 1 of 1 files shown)


Iridar:
Note that this is a very obscure, low priority bug that manifests only when you do something hacky, like make hot changes to AbilityState / ItemState after the ability has been activated, but before it was visualized. Not sure that anybody will have a use case for this fix besides myself.

robojumper (changes requested):
I don't think we should check the game state itself at all -- instead, we should ask the History for the object at a certain point in time, e.g.:

    History.GetGameStateForObjectID(AbilityContext.AbilityRef.ObjectId, , VisualizeGameState.HistoryIndex);

Iridar:
Makes sense. Accomplishes the same thing with less boilerplate. Done and tested, this at least doesn't break the base game abilities. Squashed.

robojumper (approved):

Xymanek (approved):

Issue #885, opened 2020-06-17
Add PRES Equip for Utility Slot


Iridar:
That would allow grenades equipped in the Utility Slot to show up, and make mods such as Utility Slot Secondaries Extended much simpler. Unless there's a good reason why Utility Slots cannot have PRES Equip?

Iridar:
Implemented internally, waiting on green light to submit a PR.

PR #959, opened 2021-01-14
Add OverrideWeaponScale event

### LINKED ISSUE #921: Weapon scale is overridden for female soldiers
While it's possible to rescale a weapon's mesh in WeaponInitialized (the logical place to do so if you wanted to rescale a weapon), the mesh scale will be overridden by game's code that rescales weapons for female soldiers to 85% of their original size.

A simple change to the game's code so that mesh scale is **multiplied** by 0.85 rather than **set** to 0.85 will make it possible for mods to arbitrarily rescale weapons for soldiers and other units regardless of their gender, without any changes to vanilla functionality.

Iridar:
Closes #921 

It suddenly dawned on me that the proper way to resolve #921 is through a CHL event. Tested and working as expected. 

I'd love to use this in Weapon Skin Replacer and - eventually - in XSkin with some in-game UI.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
@@ -2013,6 +2013,8 @@ simulated function EquipWeapon( XComWeapon kWeapon, bool bImmediate, bool bIsRea
 	local XComGameStateHistory History;
 	local X2WeaponTemplate WeaponTemplate;
 	local int Idx;
+	// Single variable for Issue #921
+	local float fOverrideWeaponScale;
 
 	// Jerad@Psyonix, do some stuff that InventoryManager.CreateInventory does...
 	if (kWeapon != None)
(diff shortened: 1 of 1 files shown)


robojumper:
Closing and re-opening to see if anything changes wrt. missing event docs comment.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
```diff
@@ -2080,6 +2089,39 @@ simulated function EquipWeapon( XComWeapon kWeapon, bool bImmediate, bool bIsRea
 
 	MarkAuxParametersAsDirty(m_bAuxParamNeedsPrimary, m_bAuxParamNeedsSecondary, m_bAuxParamUse3POutline);
 }
+// Start Issue #921
+/// HL-Docs: feature:OverrideWeaponScale; issue:921; tags:pawns
+/// The `OverrideWeaponScale` event allows mods to rescale weapons for unit pawns. 
+/// This event is triggered from `XComUnitPawn::EquipWeapon()` and `XComUnitPawn::AttachItem()`.
+/// The `ItemState` component of the Tuple will be `none` in the second case.
```
robojumper, line 2096: I'm not familiar with this legacy code. Is it even possible to write a hook that makes use of `ItemState` if one of the two call sites doesn't pass it?

Iridar: The `ItemState` is technically not absolutely necessary, as the Scale is a property in SkeletalMesh. 

The only reference to`XComUnitPawn::AttachItem()` in source code is `XGInventoryNativeBase::PresAddItem()`, which has no references in source code by itself, but it's an `event` function, which - from what I understand - means it can be called by native code. 

I did some testing just now, and it does in fact get called. I set up an Event Listener to resize the weapon only if the ItemState is passed, and it looks like the `AttachItem()` is called for grenades and other weapons in the utility slot. Primary and secondary weapons, as well as ones in the grenade pocket, do pass an Item State. 

What's interesting is that if I use the functionality of #885 to display the weapons in utility slots on the soldier, then they use the `XComUnitPawn::EquipWeapon()` pipeline, and do get resized properly.

So this inconsistency not a problem for the intended use case, which is to conditionally resize specific weapons for specific soldiers.

I'm gonna expand the docs with the findings of my tests.

robojumper (approved):

PR #960, opened 2021-01-14
Detach all equippable items before destroying them

### LINKED ISSUE #337: UIArmory_Loadout doesn't detach non-Primary-Secondary weapons when replacing them
https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/50d24a1fe39bf32492ff28f82296dfb91b4adaaa/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIArmory_Loadout.uc#L1153-L1163

Nothing bad seems to happen, but apparently NULL components are bad. Not sure why, but probably something to fix down the line (or just remove the code if it's redundant).

Iridar:
Closes #337 

According to my tests this "fix" doesn't seem to actually fix anything, or at least it doesn't have an effect on the number of `Attempting to detach NULL component` log warnings. Check the comment section of #337 for details.

As such, I'm not sure there's much reason to merge this PR. I'll leave this decision to seniors.


robojumper:
Thanks for doing the investigation and providing numbers. I'm going to close this and leave #337 open because there's more to this story than this PR covers, and since we don't actually know what this PR does or what effect it has, I don't see a reason to merge this. You may want to keep the branch for others to pick it up or reproduce the numbers.

PR #965, opened 2021-01-19
Rework mod dependency checking

robojumper:
The original code was somewhat loose with its terminology, in particular it didn't actually make clear whether a mod dependency was an edge or a reverse edge. This should be equivalent in functionality, but a bit more legible (both in terms of terminology and spelling) and efficient.

On top of that, this code doesn't rely on DLCIdentifiers anymore and instead reads Dependency Info for every DLCIdentifier and every DLCName (while ignoring duplicates).

Unresolved questions:

- [x] Does this work? Never used that feature.
- [x] Are the renamed functions and structs covered by our stability guarantee? Note that there's no way to make some of these functions private or protected, since `X2WOTCCommunityHighlander` needs to access them in `XComGame`.
- [x] Is `IgnoreIncompatibleMods` even a workable design? (moved to different issue)
- [x] Actually document this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,19 +7,6 @@ No new changes should be added here. Documented features should be removed from 
 
 ### Ini settings
 
-#### Mod compatibility
-
-In `XComGame.ini` mods can specify an array of incompatible and/or required mods. This will be used to show an warning popup if they are present. (#524)
-
-```
-[ModSafeName CHModDependency]
-+IncompatibleMods=OtherModSafeName
-+IgnoreIncompatibleMods=OtherModSafeName
-+RequiredMods=OtherModSafeName
-+IgnoreRequiredMods=OtherModSafeName
-DisplayName="Fancy Mod"
-```
-
 #### DLC Run Order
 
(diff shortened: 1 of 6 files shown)


robojumper:
Pinging @Musashi1584 as the original author of this system.
@Xymanek expressed interest in extending this functionality to be not based on DLCInfo classes, which I support.

robojumper:
These changes now allow mods to declare dependency info even if they don't have any DLCIdentifier. The dep checker does this by creating dep info for every DLCIdentifier and for every enabled DLCName. This approach also allows mods to declare dep info for other mods -- that info will be ignored unless that other mod is actually enabled.

Caveat emptor: All of this needs testing.

robojumper:
I'm just going to go ahead and merge the now rebased PR. This is unlikely to receive attention from anyone familiar with the code, the changes are a clear improvement in terms of both functionality, internal workings and documentation, and if any issues crop up in the beta, I'm just going to fix them. The beta is probably also the only way we're going to catch any serious issues because this interacts with users' particular mod setups more directly than other directly testable changes. This is also a prerequisite for a bunch of other PRs.

PR #966, opened 2021-01-22
Rework CHL Run Order

robojumper:
A commit with no real user-facing changes for now. Original DFS-based algorithm is kept intact, just enhanced with cycle checking and some less complex code (and better docs/var names).

Mod project with test config can be found at https://github.com/robojumper/CHLRunOrderTest

Checklist:

- [x] Finish mod dependency checking PR
- [x] Add user-facing error display?
- [x] Document feature


robojumper:
Pinging -- again, sorry -- @Musashi1584 as the original author of this system.

robojumper:
This is now based on #965 because I updated both PRs to ensure this doesn't crash the game if XComGame replacements aren't loaded. See 4fc7f16...6ffc0a5a for the actual changes.

MrNiceUK:
Is the "run order" which results when cycles exist changed, or is the "enhancement" just notifying the user of the problem?

robojumper:
This code should yield the same order that the original code created.

robojumper:
Also going to merge this with the justification in <https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/965#issuecomment-850903313>.

PR #981, opened 2021-02-27
Fix backwards compatibility breaks added by PR #910

### LINKED ISSUE #975: None checks - ensure BC behaviour
#910 has done a great effort in cleaning up the logspam left by FXS. Unfortunately, it is also technically a BC break in some cases due to how unreal VM handles missing/default values.

### Example 1

image

In original case with `none`, `ParamTag.StrValue2` would be set to `""`, while now it isn't set at all

### Example 2

image

Originally, `AS_SetFactionIcon` would be still called with a `StackedUIIconData` struct set to all default values

------

I do not see a reason to document every single case due to highly deterministic/predictable behaviour of the VM although I will be happy to answer any questions on discord

Assigning to @Iridar - the author of the PR in question

**This needs to be addressed before releasing 1.22.0** to prevent conflicting BC policy circumstances

Iridar:
Closes #975

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIImageChip.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIImageChip.uc
@@ -76,7 +76,9 @@ simulated function SetWidth(float newWidth)
 {
 	width = newWidth;
 
-	//Issue #295 - Add a 'none' check before accessing BG.
+	//Issue #295
+	//Highlight.SetX( BG.X - (PADDING_HIGHLIGHT * 0.5) );
+	//Call Highlight.SetX() with different 0 instead of BG.X if BG is 'none'.
 	if (BG != none)
 	{
 		BG.SetWidth(width);
@@ -84,7 +86,7 @@ simulated function SetWidth(float newWidth)
 	}
 	else
 	{
-		Highlight.SetX( 0 /* BG.X when BG is none */ - (PADDING_HIGHLIGHT * 0.5) );
+		Highlight.SetX( 0 - (PADDING_HIGHLIGHT * 0.5) );
 	}
 	Highlight.SetWidth( PADDING_HIGHLIGHT + width );
 	Image.SetWidth(width);
@@ -95,6 +97,8 @@ simulated function SetHeight(float newHeight)
 	height = newHeight;
 
 	//Issue #295 - Add a 'none' check before accessing BG.
+	//Highlight.SetY( BG.Y - (PADDING_HIGHLIGHT * 0.5) );
+	//Call Highlight.SetY() with different 0 instead of BG.Y if BG is 'none'.
 	if (BG != none)
 	{
 		BG.SetHeight(height);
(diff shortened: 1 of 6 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnit.uc
```diff
@@ -3690,6 +3690,11 @@ simulated static function CreateVisualizer(XComGameState FullState, XComGameStat
 		{
 			UnitVisualizer.GetPawn().AppendAbilityPerks( Ability.GetMyTemplate().GetPerkAssociationName() );
 		}
+		else
+		{
+			// Still call AppendAbilityPerks() if Ability is 'none' to preserve backwards compatibility.
+			UnitVisualizer.GetPawn().AppendAbilityPerks('');
+		}
```
Xymanek, line 3697: The original was `if( Ability.GetMyTemplate() != none )`. Assuming `Ability == none`, we get

* `if( (none).GetMyTemplate() != none )`
  * `(none).GetMyTemplate()` gives `none`
* `if( none != none )`
* `if( false )`

As such, the `UnitVisualizer.GetPawn().AppendAbilityPerks( Ability.GetMyTemplate().GetPerkAssociationName() );` line was never invoked with the `Ability == none`, hence this else branch is incorrect

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIImageChip.uc
```diff
@@ -80,9 +80,13 @@ simulated function SetWidth(float newWidth)
...
 	{
 		BG.SetWidth(width);
+		Highlight.SetX( BG.X - (PADDING_HIGHLIGHT * 0.5) );
+	}
+	else
+	{
+		Highlight.SetX( 0 /* BG.X when BG is none */ - (PADDING_HIGHLIGHT * 0.5) );
 	}
 	Highlight.SetWidth( PADDING_HIGHLIGHT + width );
-	Highlight.SetX( BG.X - (PADDING_HIGHLIGHT * 0.5) );
```
Xymanek, line 85: Instead of outright removing the line, please comment it out with a small explanation of what happened here.

It's also a good idea (from BC perspective) to avoid reordering the calls, but in this particular case I can't see any potential issues from doing so, so I'm fine with the current approach

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIImageChip.uc
```diff
@@ -94,9 +98,13 @@ simulated function SetHeight(float newHeight)
...
 	{
 		BG.SetHeight(height);
+		Highlight.SetY( BG.Y - (PADDING_HIGHLIGHT * 0.5) );
+	}
+	else
+	{
+		Highlight.SetY(0 /* BG.Y when BG is none */ - (PADDING_HIGHLIGHT * 0.5) );
 	}
 	Highlight.SetHeight( PADDING_HIGHLIGHT + height );
-	Highlight.SetY( BG.Y - (PADDING_HIGHLIGHT * 0.5) );
```
Xymanek, line 99: Same comment as above

Xymanek (changes requested):
You still need to adjust/fix

* `UITLE_SkirmishModeMenu.uc`
* `X2Action.uc`

Iridar:
@Xymanek thanks for the review, should be fixed now.

Xymanek (approved):

PR #987, opened 2021-03-18
UIStrategyPolicy misc events

### LINKED ISSUE #986: Add events to support making UIStrategyPolicy columns scrollable
`UIStrategyPolicy` columns are hardcoded to support max 3 cards per column with no scrolling capability. This feature adds a bunch of events that will allow mods to implement the scrolling without the need to MCO the screen. A sample usage can be found here: https://github.com/WOTCStrategyOverhaul/CovertInfiltration/pull/640/commits/35d26a5236e57d1091882e16c8e046ddc0890de1

Why not integrate the above implementation directly into the CHL? I considered that, and when I was first experimenting I actually did all the changes directly in CHL. In the end I came to a conclusion that it would be better to not do it directly in CHL because
1) there are still visual imperfections with my implementation. It's ok enough for CI but not something I'm confident to make an "official" thing for everyone out there
2) the 3 slots max is such a hardcoded assumption (in both uc and as) that I would honestly consider it part of the BC. As can be seen, I needed to adjust behaviour in quite a few places which is "problematic" as it can lead to a chance to break the logic (and I would prefer it to be CI vs other mod and not CHL vs other mod) because
   * all of the UISP's API is public, so mods could be calling functions/modifying values in an unexpected manner
   * MCOs could be not calling the parent implementation everywhere

Xymanek:
Closes #986

Since the events themselves have nothing to do with implementing the scrolling functionality, I named this feature "misc events" - maybe someone will have a different use for them

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIStrategyPolicy.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIStrategyPolicy.uc
@@ -10,6 +10,52 @@
 class UIStrategyPolicy extends UIScreen
 	dependson(UIDialogueBox, UIStrategyPolicy_Card);
 
+/// HL-Docs: feature:UIStrategyPolicy_MiscEvents; issue:986; tags:strategy,ui
+/// Various miscellaneous events that you can use however you want.
+/// Check the tracking issue for details on how these came to be.
+///
+/// ```event
+/// EventID: UIStrategyPolicy_PreRefreshAllDecks,
+/// EventData: none,
+/// EventSource: UIStrategyPolicy (StrategyPolicy),
+/// NewGameState: none
+/// ```
+///
+/// ```event
+/// EventID: UIStrategyPolicy_PostRealizeColumn,
+/// EventData: UIList (Column),
+/// EventSource: UIStrategyPolicy (StrategyPolicy),
+/// NewGameState: none
+/// ```
+///
+/// ```event
+/// EventID: UIStrategyPolicy_PreSelect,
+/// EventData: UIStrategyPolicy_Card (TargetCard),
+/// EventSource: UIStrategyPolicy (StrategyPolicy),
+/// NewGameState: none
+/// ```
+///
+/// ```event
+/// EventID: UIStrategyPolicy_PreClearSelection,
+/// EventData: none,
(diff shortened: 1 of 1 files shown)


pledbrook:
I think it would be good to have at least brief descriptions for when each event triggers in terms of what a player sees. At the moment, I feel forced to look at the highlander code to see whether those events might be relevant or not to me.

Xymanek:
I could add short descriptions to each event, but they would just largely duplicate the info conveyed in the event name already, I think the reality is that it's not really possible to effectively use these events without looking at the `UIStrategyPolicy`'s source anyway.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIStrategyPolicy.uc
```diff
@@ -10,6 +10,52 @@
 class UIStrategyPolicy extends UIScreen
 	dependson(UIDialogueBox, UIStrategyPolicy_Card);
 
+/// HL-Docs: feature:UIStrategyPolicy_MiscEvents; issue:986; tags:strategy,ui
+/// Various miscellaneous events that you can use however you want.
+/// Check the tracking issue for details on how these came to be.
```
pledbrook, line 15: Just a suggestion to provide a modicum of context:
```suggestion
/// `UIStrategyPolicy` is the screen that allows players to allocate
/// resistance orders at end of "month". The Community Highlander adds
/// the following events to that screen so mods have more options for
/// modifying it without overriding it entirely.
///
/// See the tracking issue for background on why these events were
/// added, and hence a potential use for them.
///
/// **Note** You'll need to look at the code for `UIStrategyPolicy` in
/// order to understand when these events are fired, and hence which
/// ones you'll need for your own customisations.
```
Not required, but I think it will help folks that are just browsing the docs looking for what stuff they can do or those specifically looking for control over the resistance orders screen.

pledbrook (approved):
Suggested a docs change. If you want to leave as it, then that's fine and I'll pre-approve.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIStrategyPolicy.uc
Xymanek: Good suggestion - applied in the force push

pledbrook (approved):

PR #991, opened 2021-05-10
Switch to X2BuildCommon

robojumper:
Opened this mostly because there are some changes I don't want to get lost, like the conversion of the `_Compiletest` flag to a Pre-Make hook, and fixes for the `update_versions.ps1` script.

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/a0a8244eb20e1f484a1ed00a5bcbf5bbd15cc666/.scripts/build.ps1#L73-L78

Some things cribbed off ec0424bf0299d86d505ffd02282c176e30fd5d9b. Supersedes #933.


Xymanek (commented):
Unless I'm missing something, the `"isDefault": true` under `Build (cooked)` prevents me from choosing the other build options when pressing ctrl+shift+b. I suggest removing the `isDefault` so that the build task selection is presented instead (of defaulting to that one)

robojumper:
Before I forget it -- this really ought to perform `git submodule init` and `git submodule update` so that the use of a submodule is *mostly* transparent to users using command line git (though preferably in a way that doesn't break local development with the submodule).

Xymanek:
With the new build system I see 3 important points:

1) The build entrypoint/configration (i.e. `build.ps1`) should be as simple as possible
2) The whole system should work the same if the project can't/won't use git 
3) ... or if someone zipped the project and sent the zip

Adding git operations to the entrypoint, along with necessary checks to support 2 and 3 + local development, is extremely likely to conflict with 1. Of course the argument could be made the CHL isn't meant to be built without git + the entrypoint is semi-complicated already but I would like all usages of X2BuildCommon to stay roughly the same (sans the exact configuration) - isn't that why we made it a common/shared lib in first place (instead of duping and customizing the old build script)?

As an alternative approach, how about using git's `post-checkout` and `post-merge` [hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)? `build_common.ps1` (where adding complexity is not a concern) would then install said hooks on the first run (or something like that)

Xymanek:
Now that we've decided to use `git subtree` instead of the a submodule, can this PR be updated?

Xymanek (approved):

PR #997, opened 2021-05-19
Add event to override X2ReactionFireSequencer::IsReactionFire()

### LINKED ISSUE #996: Add an event to override ReactionFireSequencer
ReactionFireSequencer is hardcoded to add cinematic slo-mo to abilities that have `bReactionFire = true` set on their ToHitCalc. This is a big inflexible, as there are legitimate cases where one would want to use cinematic slo-mo with non-reaction abilities, and don't use cinematic slo-mo with non-reaction abilities, and in general it's weird to have game state logic be tied with visualization logic like that.

As such, I propose adding an event into `X2ReactionFireSequencer::IsReactionFire()` that will allow mods to override which abilities count as valid for cinematic slo-mo.

Iridar:
Closes #996

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2ReactionFireSequencer.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2ReactionFireSequencer.uc
@@ -1,0 +1,357 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2ReactionFireSequencer.uc
+//  AUTHOR:  Ryan McFall --  07/08/2015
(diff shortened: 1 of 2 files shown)


Iridar:
Hmm not sure why, but this commit doesn't modify .x2proj file, and I couldn't add it via terminal either.

Iridar:
Tested to be working for the purposes of removing reaction fire slo mo under specific circumstances. 

As mentioned in the issue #996, I also want to add an event to modify reaction fire Hit Modifier, is this appropriate to do it in this issue, or should I file a separate one?

Xymanek (approved):

PR #999, opened 2021-05-28
Bring back the DLC2 component (X2ModBuildCommon)

### LINKED ISSUE #803: Revisit DLC HL Components

Xymanek:
Supersedes #858 - this time using X2ModBuildCommon 

Closes #803 (+ context)


Xymanek:
Current (WIP) popups:

XCom2_U4i8BpvEhY
rV2S9hqtxq
XCom2_LQSrKxJ3BE

Xymanek:
"Final" versions of the popups:

XCom2_xYtPm4bvG6
XCom2_1tT00hAyDP

For naming I decided:

* Use `X2WOTCCommunityHighlander` as it is now on the workshop
* Use `Alien Hunters Community Highlander` for the DLC2HL

Xymanek:
Rebased to resolve merge conflict

Xymanek:
Future TODOs:

* Expand the troubleshooting wiki page (mention DLC2HL)
* Expand the workshop description to link the DLC2HL

For reviewers: there are several (4 afaik) changes pending to XCGS_ARM (some were already reviewed and approved). I will re-submit them in a separate PR, but I added the file here to make reading the actual changes PR easier

Components/DLC2CommunityHighlander/DLC2CommunityHighlander/ReadMe.txt
```diff
@@ -0,0 +1 @@
+﻿You created an XCOM 2 Mod Project!
```
Iridar, line 1: This is nitpicking, but perhaps put player-facing basics from the #803 description here? It's good that ReadMe.md has modmaker-facing description in there, though.

Components/DLC2CommunityHighlander/DLC2CommunityHighlander/Src/DLC_2/Classes/CHDLC2UIListener_Shell.uc
```diff
@@ -0,0 +1,41 @@
...
+// Not part of the public API - mods should not rely on its existence
+
+class CHDLC2UIListener_Shell extends UIScreenListener;
+
+var localized string strMissingCHL_Title;
+var localized string strMissingCHL_Text;
+
+event OnInit (UIScreen Screen)
+{
+	if (!Screen.IsA(class'UIShell'.Name)) return;
```
Iridar, line 11: Pure idle curiosity, but how is this superior to `Screen.IsA('UIShell')`?

X2WOTCCommunityHighlander/Localization/X2WOTCCommunityHighlander.int
```diff
@@ -19,4 +19,8 @@ CycleErrorTitle="Cycle in Run Order Detected"
...
 GroupErrorTitle="Conflicting Run Order data"
 GroupErrorText="Mods specified conflicting run order information:\n\n%FACTS\nThis behavior is specified, but may indicate a configuration error."
-BlameText="The following DLCIdentifiers provided config that lead to this problem: %BLAME\n\nPlease inform the respective mod authors about this problem."
\ No newline at end of file
+BlameText="The following DLCIdentifiers provided config that lead to this problem: %BLAME\n\nPlease inform the respective mod authors about this problem."
+
+[X2WOTCCH_UIScreenListener_Shell_DLCIntro]
+strDLC2HLIntroduction_Title="Alien Hunters Community Highlander"
+strDLC2HLIntroduction_Body="Starting with X2WOTCCommunityHighlander v1.22.0 we are introducing a highlander for the Alien Hunters DLC.<br><br>Please install it to<br><Bullet/> Ensure correct behaviour of mods that depend on X2WOTCCommunityHighlander<br><Bullet/> Benefit from the bugfixes<br><br>You can find it on workshop under the \"Alien Hunters Community Highlander\" name (or just follow a link in X2WOTCCommunityHighlander's description)"
```
Iridar, line 26: This text looks good to me.

Iridar (approved):
Looks great to me, for what it's worth.

Components/DLC2CommunityHighlander/DLC2CommunityHighlander/Src/DLC_2/Classes/CHDLC2UIListener_Shell.uc
Xymanek: `class''.Name` will catch typos, e.g. `class'UISheII'.Name` (capital `i` instead of `l`) won't compile, while just `''` will compile and cause problems at runtime

Iridar: I see, thanks for the clarification. I'm impressed with your level of care, even rich plutonium is probably not handled that delicately :D

Components/DLC2CommunityHighlander/DLC2CommunityHighlander/ReadMe.txt
Xymanek: I don't see why not, but then it IMO should be a fully-fledged description rather a 1/2 liner. I suggest not holding this PR for it and instead adding it in a separate PR

Iridar: by all means

Xymanek:
I'll merge this in a couple of days so that I can proceed with the pending changes to XCGS_ARM in case there are no more reviews/comments

PR #1002, opened 2021-05-30
Add localization for required/incompatible mods popup

Iridar:
Adds localization to the popup added in Issue #524.

Per robojumper's instructions, I kept localized files in UTF-8 format, although I'm wary how now all of them contain question marks. This should be double-checked, IMO.


Iridar:
>although I'm wary how now all of them contain question marks. This should be double-checked, IMO.

This was checked and fixed.

Xymanek (commented):
Not sure what/how to review this... The "infrastructure" aspect looks correct

PR #1007, opened 2021-06-08
Mods can specify required Highlander version in config (reupload)

### LINKED ISSUE #909: Expand CHModDependency functionality to check the required Highlander version for each mod individually

Iridar:
Closes #909

Replaces #955

Here's how the popup looks: 

https://i.imgur.com/Mrll6e9.jpg

A slight issue with this approach is that we don't display Highlander version on the workshop, so most people who see this popup will have trouble figuring out what are they supposed to do to update the Highlander.

### DIFF
--- a/X2WOTCCommunityHighlander/Localization/X2WOTCCommunityHighlander.int
+++ b/X2WOTCCommunityHighlander/Localization/X2WOTCCommunityHighlander.int
@@ -14,6 +14,13 @@ ModIncompatiblePopupTitle="detected INCOMPATIBLE mods"
 ModIncompatible="Please DEACTIVATE the following mods and restart or %s will not work properly:"
 DisablePopup="DO NOT SHOW ME THIS AGAIN"
 
+;    Issue #909
+ModRequiresNewerHighlanderVersionTitle = "MOD REQUIRES NEWER HIGHLANDER VERSION"
+ModRequiresNewerHighlanderVersionText = "Mod requires newer Highlander version:"
+CurrentHighlanderVersionTitle = "Installed Version:"
+RequiredHighlanderVersionTitle = "Required Version: "
+ModRequiresNewerHighlanderVersionExtraText = "Please update X2WOTCCommunityHighlander to the required version. Otherwise you may experience crashes and other malfunctions."
+
 [X2WOTCCH_UIScreenListener_ShellPopup]
 ErrorTitle="Mod Configuration Error"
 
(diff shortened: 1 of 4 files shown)


robojumper:
> A slight issue with this approach is that we don't display Highlander version on the workshop

Good call. We should do that.

robojumper (commented):
Approach LGTM. This PR has some formatting issues -- much code uses spaces instead of tabs (and there is some trailing whitespace in `ModDependencies.uc(198)` and a double space in the signature in `ModDependencies.uc(207)`).

Iridar:
> Approach LGTM. This PR has some formatting issues -- much code uses spaces instead of tabs (and there is some trailing whitespace in `ModDependencies.uc(198)` and a double space in the signature in `ModDependencies.uc(207)`).

Ah, probably because I copied it off Source Tree interface from the old PR rather than code itself. My bad, will get that fixed.

Iridar:
Hopefully should be all fixed now, thanks for your vigilance and details.

Xymanek:
I'm not sure about the overall value/use case of this system. The popup states:

> Mod requires newer Highlander version
> Please update X2WOTCCommunityHighlander

However, 99% of CHL users are using it by subscribing to it on workshop, which handles the updates on its own. In fact, they simply have no ability to affect the process, so the call to action is likely to just create confusion.

Perhaps add some sort of reference to the beta version?

Iridar:
> I'm not sure about the overall value/use case of this system. The popup states:
> 
> > Mod requires newer Highlander version
> > Please update X2WOTCCommunityHighlander
> 
> However, 99% of CHL users are using it by subscribing to it on workshop, which handles the updates on its own. In fact, they simply have no ability to affect the process, so the call to action is likely to just create confusion.
> 
> Perhaps add some sort of reference to the beta version?

Good point. The general use for this popup is to let the user know they need to switch to beta, yeah. It could also be potentially used by projects that need custom highlander (which is a practice I personally disagree with), but it looks like both of them will be released after 1.22 hits the workshop anyway, so the point is moot.

Perhaps, the popup should direct the user to the problematic mod's description, which would refer the required highlander version?

Also, not everyone uses steam workshop, people might be modding non-steam version of the game, or load highlander locally for whatever reason, so the note about needing to update the chl seems valid to me.

Iridar:
I rebased the PR to the main branch and confirmed it still works fine.

PR #1010, opened 2021-06-09
Add `SetPriorityTargetStack` behavior tree node (#1009)

### LINKED ISSUE #1009: Add behaviour tree node that only selects priority targets
In LW2 and LWOTC, there is a very specific problem involving Najas (sniper Vipers) and Berserkers on Intel Raid missions (see [this LWOTC issue](https://github.com/long-war-2/lwotc/issues/879)). All the enemies on the map start in yellow alert, which allows them to shoot at the priority target (a relay device). This is normally fine, but Najas can shoot at squadsight range at anything they have squadsight vision of. If they can't shoot the priority target, they instead will shoot at any XCOM soldiers or rebels that they have line of sight on. This is about as BS as you can get in XCOM.

Berserkers are different: if they can't attack the priority target, they will just attack one of the other members of their pod.

The source of this problem is that the standard yellow alert action tree selector, `TrySelectYellowAlertAction`, has the node `TryAttackPriorityTarget`, which in turn ultimately uses `SetTargetStack-StandardShot` to pick a target to shoot. However, the `SetTargetStack` node will also pick any other non-priority targets. So if the priority target can't be attacked (and it can only be attacked once per turn from yellow alert), then any non-priority targets become fair game.

I would like to modify the AI node that uses `SetTargetStack-StandardShot` for this scenario so that it is limited to the priority target only. This would mean creating a new BT node in code. It's technically possible to add this to a subclass of `XGAIBehavior` and configure all character templates to use the new subclass for behaviours. However, that could easily run into mod conflicts with different mods providing different `XGAIBehavior` classes.

So, I would like to add something like an `XGAIBehavior.SetPriorityTargetStack()` function that I can use from a modified AI config for yellow alert actions in order to prevent the above scenario, which is incredibly frustrating for players and rightfully considered a bug.

pledbrook:
This adds a new behavior tree node that is based on `SetTargetStack` except it *only* selects priority targets. It has no fallback to other targets like `SetTargetStack` does.

It is primarily needed for AI that should attack priority targets, but no others, for example units on yellow alert. Note that the AI config needs to be modified to take advantage of this node; it doesn't do anything by itself.

Resolves #1009.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIBehavior.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIBehavior.uc
@@ -3229,6 +3229,77 @@ function bool BT_SetTargetStack( Name strAbility )
 	return false;
 }
 
+// Start Issue #1009
+/// HL-Docs: feature:SetPriorityTargetStack; issue:1009; tags:tactical
+/// This new behavior tree node will add only the priority target to
+/// the AI's target stack, if there is a priority target available.
+/// No other potential targets will be added. If there is no priority
+/// target, the target stack will be empty.
+///
+/// Example usage in XComAI.ini:
+/// ```ini
+/// +Behaviors=(BehaviorName=SelectPriorityTargetForStandardShot, NodeType=Sequence, \\
+///             Child[0]=SetPriorityTargetStack-StandardShot, Child[1]=SelectPriorityTarget, \\
+///             Child[2]=HasValidTarget-StandardShot)
+/// ```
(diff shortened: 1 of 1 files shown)


Iridar (approved):
Approach LGTM and worse case scenario its LWOTC users who will suffer 👍

pledbrook:
I need to test a suggestion by Xymanek that may mean this pull request is not necessary.

pledbrook:
The custom node mechanism used by Alien Hunters and Shen's Last Gift works, albeit rather clunkily. So this can be done in LWOTC itself, or any other mod.

PR #1011, opened 2021-06-09
Fix pods not scampering (#1008)

### LINKED ISSUE #1008: Immobilising pod leader on scamper prevents the rest of the pod from scampering
Here's the scenario:

1. XCOM is in concealment
2. One of the soldiers targets a pod leader with an immobilising ability (Maim, Crippling Strike from LWOTC, or whatever)
3. The ability breaks concealment, triggering the "scamper"
4. The pod leader can't move, so doesn't scamper
5. None of the rest of the pod scampers, typically leaving them all flanked for an easy clean up

The problem is in the `XGAIBehavior.SkipTurn()` function, which has this bit of code:
```
// If unrevealed, the entire group skips its turn.  Fixes assert with group movement, after group leader skips its move.
if( StartedTurnUnrevealed() )
{
	SkipGroupTurn();
}
```
When a pod needs to scamper, this condition is always true. That means the rest of the pod will skip their turns too.

In fact, I suspect this isn't just a problem for immobilising abilities that force the pod leader to skip their turn. It may also happen if the pod leader is by some cover and just stays where they are. Not sure if this is also the cause of pod members failing to scamper if the pod leader dies on overwatch. Less likely I think.

Proposed fix is to add an additional condition to the above `if` statement:
```
if( StartedTurnUnrevealed( && !m_kPlayer.IsScampering(UnitState.ObjectID) )
{
	SkipGroupTurn();
}
```
I tested it briefly and it fixed the immediate issue. I'm fairly sure the comment about fixing an assert is most likely connected with green or yellow alert actions rather than scampering, but it's hard to be 100% sure.

pledbrook:
The `SkipTurn()` implementation in `XGAIBehavior` skips the entire pod's turn if the pod starts the turn "unrevealed", which basically means they haven't activated by the start of the turn. This happens when the pod
leader skips moving on activation, for example if they're immobilized or panicked.

The fix I've implemented here is to check whether the pod is scampering and only skip the pod's turn if they *aren't*. This ensures that the pod members that can scamper actually do so.

Fixes #1008.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIBehavior.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIBehavior.uc
@@ -9988,8 +9988,11 @@ simulated function SkipTurn( optional string DebugLogText="" )
 	RefreshUnitCache();
 	if (UnitState.NumAllActionPoints() != 0)
 	{
+		/// HL-Docs: ref:Bugfixes; issue:1008
+		/// Entire pods no longer fail to scamper if the pod leader skips its turn,
+		/// for example if it's immobilized.
 		// If unrevealed, the entire group skips its turn.  Fixes assert with group movement, after group leader skips its move.
-		if( StartedTurnUnrevealed() )
+		if( StartedTurnUnrevealed() && /* Issue #1008 */ !m_kPlayer.IsScampering(UnitState.ObjectID) )
 		{
 			SkipGroupTurn();
 		}


Iridar (approved):
Solution make sense, needs (public) testing to make sure it catches what it is supposed to catch.

pledbrook:
Tested with the config var. Behaves as expected.

Iridar:
> Tested with the config var. Behaves as expected.

Ah good. Question; why is this gated behind a config var? Seems like a straightforward bugfix to me.

pledbrook:
Config var was part of a potential compromise to get the change into 1.22. That didn't happen.

Iridar:
> Config var was part of a potential compromise to get the change into 1.22. That didn't happen.

I see. Personally I think it's fine to include this into 1.22.2 along with #1032, without the config var, enabled as baseline. @Xymanek, what's your take on this?

Xymanek:
I feel like we keep going in circles on this matter. My position was that 1.22.0 has a lot of stuff already and if something went wrong (which did happen) the debugging process would painful due to the huge diff. I suggested that we wait until 1.22.0 (well, 1.22.2 at this point) is released to stable, then ~2 week(end)s and then if we don't get any reports of issues, we can start preparing 1.23, including this. Nothing has changed, I still stand by this opinion.

P.S. I do agree that this should be done without the config var

Iridar:
> I feel like we keep going in circles on this matter. My position was that 1.22.0 has a lot of stuff already and if something went wrong (which did happen) the debugging process would painful due to the huge diff. I suggested that we wait until 1.22.0 (well, 1.22.2 at this point) is released to stable, then ~2 week(end)s and then if we don't get any reports of issues, we can start preparing 1.23, including this. Nothing has changed, I still stand by this opinion.
> 
> P.S. I do agree that this should be done without the config var

Fine be me. My memory is trash and the last week was super busy, so I was a bit fuzzy on the matter, sorry.

Iridar:
In that case @pledbrook please update this PR to go without a config var and we'll merge it into 1.23 when the time comes.

pledbrook:
Removed config variable and rebased against `master`. No reports from LWOTC of issues related to this change and it appears to fix the issues with pods not scampering.

Iridar (approved):

PR #1013, opened 2021-06-10
Manually clean up Particle System Components spawned into Emitter Pool by X2UnifiedProjectile

### LINKED ISSUE #720: Suppression drains performance
**Problem:**
Firing cosmetic shots during Suppression (perhaps, firing any kind of projectile volley) reduces game's performance (frames per second).

**Reproduction:**
Set up approx. 6 XCOM soldiers vs 6 enemies and have them all suppressing each other. Witness the FPS gradually reducing over time.

**Iridar's Test Report**
> I've done my own test with the Suppression. I cleared config and launched the game only with Highlander, and set up 6 cannon grenadiers to suppress 6 advent troopers. RAM consumption started at 2 GB, and FPS at 90. Over the course of the next 10 minutes, RAM consumption grew up to 3.6 GB, and FPS dropped to 15. Note: I still had about 8 GB of free RAM to go. Then I entered "SkipAI" and ended turn. When the next turn began, RAM consumption dropped to 3.0 GB, but FPS stayed at 15.
> 
> RAM consumption was growing only when there was some shooting happening on screen. It was stable before and after the suppression turn.
> 
> FPS didn't recover even after killing all enemies and skipping a few turns and waiting a few minutes.
> 
> After a few more minutes RAM consumption dropped to 2.3 GB and kept falling, but FPS still didn't recover.

**Temporary Fix**
Loading a saved game seems to instantly recover any lost FPS. Restarting the game is not necessary.

**Ramifications**
This seems like a bug that is endemic to XCOM 2 WOTC (perhaps vanilla XCOM 2 as well). 

When playing WOTC without increasing enemy count and squad size, this bug is not much of an issue. Missions are short, shots are few and there is barely any suppression going in.
The opposite example of that is Beagle's Operators vs Aliens mod collection, which relies on increasing squad size, enemy count, and almost every unit gains Suppression, and missions quite often become trench warfare with multiple turns of multiple units using Suppression non stop, which drains performance quickly. Missions themselves are longer, further exacerbating the problem.

For the good of all people who play modded XCOM, this issue must be resolved, and Highlander seems the perfect place for it, assuming the problem can be resolved without overly invasive measures and nasty side effects.

**Potential Causes**
It has been speculated that there is a some sort of memory leak that is at fault. Indeed, during most of my tests, the XCOM's memory consumption would gradually grow over time, following the reduction in performance.

However, if *memory loss* was the cause of the performance loss, it would be a sharp loss of performance that would happen only as the system would run out of RAM space. But in my tests the FPS loss was gradual, and would reach unplayable levels long before my system would run out of RAM.

Additionally, in some of my tests, the consumed memory would eventually get freed by the game, but the FPS would not recover. 

Therefore I'm compelled to conclude that memory leak is a side effect of the problem rather than the direct cause of performance loss. 

I speculate that each time the unit fires a projectile volley, the game creates some sort of *actor* or *object*, and then "forgets" to remove it. If there are any functions that iterate over all instances of *actors* or *objects* of that type, they would take more and more time to do their job as the number of these actors or objects would grow, which would explain the very gradual loss in performance.

The fact that FPS seems to recover the moment you load a save game (which loads all the state objects, but destroys any currently active visualization objects), I'm guessing the problematic *objects* or *actors* are a part of the game's visualization system rather than the gamestate system.

**Thoughts of other prominent modders on the topic:**

**robojumper:** 
> You should still try to find what precisely is leaking.
> Tip: Write some sort of function that loops over Actor types and prints out interesting actors, do the same with impact components (need to find what they're attached to though)

**Xymanek:** 
> isn't there some sort of console command to get classes by instance count?

**robojumper:**
> Oh yeah OBJ LIST or something

**Musashi:**
> there is ListDynamicActors - Outputs a list of all dynamic actors (actors which were created after the start of the level) to the log.
https://docs.unrealengine.com/udk/Three/ExecFunctions.html

**E3245:** 
> https://docs.unrealengine.com/udk/Three/ConsoleCommands.html#Debugging%20Commands
> Also this. The memory commands are also useful.

**Next Step**
As robojumper has indicated, the next step seems to be confirming the problem is indeed caused by the game creating and forgetting some sort of *actor* or *object*, narrowing down said actor or object, and figuring out if we modders can do something about it, within the confines of Highlander or otherwise.

Iridar:
Closes #720 

Let me know of comments are insufficient or we need docs. Or a ref:bugfix tag.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
@@ -402,6 +402,39 @@ function EndConstantProjectileEffects( )
 			Projectiles[ Index ].ParticleEffectComponent.OnSystemFinished = none;
 			Projectiles[ Index ].ParticleEffectComponent.DeactivateSystem( );
 			Projectiles[ Index ].ParticleEffectComponent.KillParticlesForced( );
+			// Start Issue #720
+			/// HL-Docs: feature:ProjectilePerformanceDrain; issue:720; tags:tactical
+			/// The game uses a global pool `WorldInfo.MyEmitterPool` to store Particle System Components (PSCs),
+			/// which helps optimize performance. When a PSC is no longer needed, it is returned to the pool 
+			/// so that it can be reused again later.
+			///
+			/// To return a PSC back to the pool, the `EmitterPool::OnParticleSystemFinished()` function must be called, 
+			/// which usually happens seamlessly - the pool sets the `ParticleSystemComponent::OnSystemFinished` delegate 
+			/// to the aforementioned function when the PSC is initially borrowed from the pool 
+			/// (via `EmitterPool::SpawnEmitter`).
+			///
+			/// However, `X2UnifiedProjectile` overwrites the `OnSystemFinished` delegate with its own one 
+			/// after borrowing the PSC from the pool, which means the `EmitterPool::OnParticleSystemFinished()`
+			/// is never called for these PSCs. 
+			///
+			/// As such, the pool is never made aware that those PSCs are ready for reuse and when a new PCS is requested, 
+			/// the pool has no choice but to simply create a new one, resulting in the pool getting bloated 
+			/// with all the PSCs ever spawned by `X2UnifiedProjectile`s.
+			///
+			/// This bloat leads to two observable effects: endlessly increasing RAM usage and a significant 
+			/// performance degradation. The more projectiles are fired over the course of a mission, the worse 
+			/// these effects become. Frequent use of the Suppression ability or Idle Suppression mod 
+			/// kick the problem into overdrive.
+			///
+			/// To address this bug, we manually call `WorldInfo.MyEmitterPool.OnParticleSystemFinished()` on PSCs 
+			/// in `X2UnifiedProjectile` when the projectile is done with them.
+			///
+			/// Mods that create Particle System Components using the Emitter Pool must carefully handle them the same way:
+			/// if the PSC's `OnSystemFinished` delegate is replaced, then `EmitterPool::OnParticleSystemFinished()` must
+			/// be called for this PSC manually when the PSC is no longer needed, otherwise the same "memory leak" 
+			/// will occur. This applies to all PSCs using the Emitter Pool, not just those in `X2UnifiedProjectile`.
+			WorldInfo.MyEmitterPool.OnParticleSystemFinished(Projectiles[ Index ].ParticleEffectComponent);
+			// End Issue #720
 			Projectiles[ Index ].ParticleEffectComponent = none;
 			Projectiles[ Index ].SourceAttachActor.SetPhysics( PHYS_None );
 			Projectiles[ Index ].TargetAttachActor.SetPhysics( PHYS_None );
@@ -1072,6 +1105,12 @@ function OnParticleSystemFinished(ParticleSystemComponent PSystem)
 	{
 		if (Element.ParticleEffectComponent == PSystem)
 		{
+			// Start Issue #720
+			/// HL-Docs: ref:ProjectilePerformanceDrain
+			// Allow the pool to reuse this Particle System's spot in the pool.
+			WorldInfo.MyEmitterPool.OnParticleSystemFinished(Element.ParticleEffectComponent);
+			// End Issue #720
+
 			Element.ParticleEffectComponent = none;
 			return;
 		}
@@ -2121,6 +2160,11 @@ state Executing
 				{
 					Projectiles[ Index ].ParticleEffectComponent.OnSystemFinished = none;
 					Projectiles[ Index ].ParticleEffectComponent.DeactivateSystem( );
+					// Start Issue #720
+					/// HL-Docs: ref:ProjectilePerformanceDrain
+					// Allow the pool to reuse this Particle System's spot in the pool.
+					WorldInfo.MyEmitterPool.OnParticleSystemFinished(Projectiles[ Index ].ParticleEffectComponent);
+					// End Issue #720
 					Projectiles[ Index ].ParticleEffectComponent = none;
 				}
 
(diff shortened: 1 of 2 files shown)


Xymanek (commented):
Why `class'WorldInfo'.static.GetWorldInfo()` instead of just `WorldInfo`?

Iridar:
> Why `class'WorldInfo'.static.GetWorldInfo()` instead of just `WorldInfo`?

Didn't know it was an option, I'm not familiar with WorldInfo so I dunno where it exists. I'll adjust.

Xymanek:
Regarding the docs: IMO this deserves a separate page because there is tons of explanation to present the full chain of events and there are more than 1 places in code which were changed

Iridar:
Alright, I'll add some docs then.

Iridar:
Added docs.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
```diff
@@ -402,6 +402,29 @@ function EndConstantProjectileEffects( )
...
+			/// into the pool, which means the `EmitterPool::OnParticleSystemFinished()` is never called
+			/// for these PSCs, causing a "memory leak" bug, where the Pool becomes bloated with all the
+			/// PSCs ever spawned by `X2UnifiedProjectile`s.
+			///
+			/// This causes a significate performance degradation if there are a lot of shots fired during a mission,
+			/// particularly if the squad and enemy pods have increased sizes and/or the mission is long.
+			/// Frequent use of the Suppression ability or Idle Suppression mod kick the problem into overdrive.
+			/// 
+			/// To address this bug, we manually call `WorldInfo.MyEmitterPool.OnParticleSystemFinished()`
+			/// on PSCs in `X2UnifiedProjectile` when it's time for them to be removed.
```
Xymanek, line 425: That's not a correct explanation/understand of how an object pool works. Here's a better one:

> When a PSC is "done", it is returned to the pool so that it can be reused again later.
> 
> To return a PSC back to the pool, the `EmitterPool::OnParticleSystemFinished()` function must be called, which usually happens seamlessly - the pool sets the `ParticleSystemComponent::OnSystemFinished` delegate to the aforementioned function when the PSC is initially borrowed from the pool (via `EmitterPool::SpawnEmitter`).
> 
> However, `X2UnifiedProjectile` overwrites the `OnSystemFinished` delegate with its own one (after borrowing the PSC), which means the `EmitterPool::OnParticleSystemFinished()` is never called for these PSCs. As such, the pool is never made aware that those PSCs are ready for reuse and when a new PCS is requested, the pool has no choice but to simply create a new one, resulting in the pool getting bloated with all the PSCs ever spawned by `X2UnifiedProjectile`s.
> 
> This bloat leads to 2 observable effects: endlessly increasing RAM usage and a significant performance degradation, especially if there are a lot of shots fired during a mission, particularly if the squad and enemy pods have increased sizes and/or the mission is long. Frequent use of the Suppression ability or Idle Suppression mod kick the problem into overdrive.
> 
> To address this bug, we manually call `WorldInfo.MyEmitterPool.OnParticleSystemFinished()` on PSCs in `X2UnifiedProjectile` when the projectile is done with them.

That said, I'm not a huge fan of phrasing (especially the `[...] especially [...], particularly [...]`) so I would prefer some editorial work on it first

Xymanek (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
Iridar: Okay, I'll adjust.

Iridar:
Adjusted docs, how's that?

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
```diff
@@ -402,6 +402,34 @@ function EndConstantProjectileEffects( )
...
+			/// after borrowing the PSC from the pool, which means the `EmitterPool::OnParticleSystemFinished()`
+			/// is never called for these PSCs. 
+			///
+			/// As such, the pool is never made aware that those PSCs are ready for reuse and when a new PCS is requested, 
+			/// the pool has no choice but to simply create a new one, resulting in the pool getting bloated 
+			/// with all the PSCs ever spawned by `X2UnifiedProjectile`s.
+			///
+			/// This bloat leads to two observable effects: endlessly increasing RAM usage and a significant 
+			/// performance degradation. The more projectiles are fired over the course of a mission, the worse 
+			/// these effects become.  Frequent use of the Suppression ability or Idle Suppression mod
```
Xymanek, line 426:
```suggestion
			/// these effects become. Frequent use of the Suppression ability or Idle Suppression mod 
```

Xymanek (commented):
I think it's also a good idea to add a paragraph about X2UP subclasses - they might need to do the same if they manually create/destroy PSCs or overwrite the fixed functions

Iridar:
> I think it's also a good idea to add a paragraph about X2UP subclasses - they might need to do the same if they manually create/destroy PSCs or overwrite the fixed functions

Will do, although we don't have anything to worry too much about anyone doing that, support for subclasses of X2UnifiedProjectile has only recently became available, and to best of my knowledge only Musashi and I took advantage of it.

Iridar:
Though point actually stands for *all* PSCs created into the Pool.

Xymanek:
The reason I mentioned X2UP subclasses is that overriding a fixed function without re-applying the fix will bring the issue back. That said, perhabs a better place to mention this is the `OverrideProjectileInstance` documentation (i.e. just add a mention/link to `ProjectilePerformanceDrain`).

Iridar:
> The reason I mentioned X2UP subclasses is that overriding a fixed function without re-applying the fix will bring the issue back. That said, perhabs a better place to mention this is the `OverrideProjectileInstance` documentation (i.e. just add a mention/link to `ProjectilePerformanceDrain`).

Do you want me to do this within this PR?

Xymanek:
Yes

Iridar:
> Yes

Doneso.

Xymanek (approved):

PR #1020, opened 2021-06-16
Fix AllTargetStylesNative detection

### LINKED ISSUE #1019: Script targeting style detection is too overzealous
https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/84ae0e3b34458d45b968f0893d0c3d3ae238907f/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTemplate.uc#L257-L283

I've added a bit of logging:

```uc
`log("Instantiating XComGameState_Ability_CH for" @ DataName,, 'CH_ScriptTargetStyle');
Ability = XComGameState_Ability(NewGameState.CreateNewStateObject(class'XComGameState_Ability_CH', self));
```

Result is a bit more than one would expect: https://gist.github.com/Xymanek/7d568340cb4811e8e64e5ccd50cd72d5

cc @robojumper as the author of the latest implementation (#918)

robojumper:
Fixes #1019.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTemplate.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTemplate.uc
@@ -277,8 +277,8 @@ function XComGameState_Ability CreateInstanceFromTemplate(XComGameState NewGameS
 simulated private function bool AllTargetStylesNative()
 {
 	// The AbilityPassiveAOEStyle is never required by native code, so we don't have to check it here.
-	return (AbilityTargetStyle != none && CH_ClassIsNative(AbilityTargetStyle.Class))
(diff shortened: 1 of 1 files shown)


robojumper:
I'm 99% sure this is the correct fix but I haven't tested it.

Xymanek (commented):
Can confirm that the replacement is no longer used when it's not needed.

However, I do not have a test case for when the replacement is actually needed.

Iridar:
I have a test case for this, I'll see about getting this tested.

Iridar (approved):
I've tested and can confirm it's working as intended for abilities with custom multi target styles.

Iridar:
Perhaps I can use this as an opportunity to merge something? How do I proceed, just merge the PR with the default merge message?

PR #1021, opened 2021-06-16
Include DLC2 Highlander in documentation

### LINKED ISSUE #1017: DLC2HL is not processed by the documentation infrastructure
See #1015 - there was no auto-generated comment about listener diff (`AllowRulerOnMission` event was added) and the new features do appear on the documentation site

robojumper:
Fixes #1017. Folders list got a bit unwieldy, so we allow a separate file to list documentation input directories.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -37,7 +37,7 @@ jobs:
         # Build base docs (master)
         git checkout ${{ github.base_ref }}
         mkdir target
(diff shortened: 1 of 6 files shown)


robojumper:
`check-event-templates` fails because this adds a new command line option and tries to call the current `master` script with the new command line option, which fails. Shouldn't be an issue really.

Xymanek (approved):
Looks OK at first glance, time to test in prod

PR #1023, opened 2021-06-16
Add gameplay visibility condition to second shots (#1022)

### LINKED ISSUE #1022: RapidFire2 and ChainShot2 are missing gameplay visibility conditions
Perhaps the author thought the condition was unnecessary as the first shot has the condition. However, when the game cache for the unit is initialised, these abilities are included, and they pick up every enemy on the map as valid targets, causing lots of aim checks. When you have 40+ enemies on the map, that can slow things down.

pledbrook:
The Rapid Fire and Chain Shot second shot abilities did not have the gameplay visibility condition, so their valid targets included every enemy unit on the map, which was causing performance issues in LWOTC. There appears to be no reason for these abilities not to have this condition, so adding them in.

Fixes #1022.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_GrenadierAbilitySet.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_GrenadierAbilitySet.uc
@@ -898,6 +898,7 @@ static function X2AbilityTemplate ChainShot2()
 	Template.AddShooterEffectExclusions();
 
 	Template.AbilityTargetConditions.AddItem(default.LivingHostileTargetProperty);
+	Template.AbilityTargetConditions.AddItem(default.GameplayVisibilityCondition);
 
 	Template.AddTargetEffect(HoloTargetEffect());
 	Template.AssociatedPassives.AddItem('HoloTargeting');
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_RangerAbilitySet.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_RangerAbilitySet.uc
@@ -1,0 +1,843 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2Ability_RangerAbilitySet.uc
+//  AUTHOR:  Timothy Talley  --  03/06/2014
+//  PURPOSE: Defines all Ranger Based Class Abilities
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+class X2Ability_RangerAbilitySet extends X2Ability
+	dependson (XComGameStateContext_Ability) config(GameData_SoldierSkills);
+
+var config int BLADEMASTER_DMG, BLADEMASTER_AIM;
+var config int STEALTH_CHARGES;
+var config int SHADOWSTRIKE_AIM, SHADOWSTRIKE_CRIT;
+var config int MAX_UNTOUCHABLE;
+var config int REAPER_COOLDOWN;
+var config int INSTINCT_DMG, INSTINCT_CRIT;
+var config int RAPIDFIRE_AIM;
+var config int RUNANDGUN_COOLDOWN;
+
+var name SwordSliceName;
+
+static function array<X2DataTemplate> CreateTemplates()
+{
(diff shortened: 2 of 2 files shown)


Xymanek:
I gather this will improve performance for everyone with more-than-normal units in play (i.e. larger squad/pod sizes) and not only lwtoc users?

pledbrook:
Absolutely. There's nothing LWOTC specific about the bug or its impact. Just the numbers of enemies on the map and perhaps how expensive aim calculations are.

robojumper:
> There appears to be no reason for these abilities not to have this condition, so adding them in.

I could imagine one. Consider a target that the shooter only has LOS on through stepping out of cover. The first Rapid Fire shot destroys the cover, now the unit doesn't have LOS on the target for the second shot. Without the gameplay visibility condition, the second shot will succeed, but with these conditions, the second shot ability is unavailable and the event listener cannot trigger the ability.

Moreover, I could imagine that these conditions were taken out in response to such a bug report.

pledbrook:
It could be argued that if the second shot doesn't have gameplay visibility, then it *shouldn't* activate. But I can see that being very confusing to the player. These abilities are a massive burden on the game, though, if they don't have some condition that prevents the entire map of enemies being valid targets.

Perhaps a "within visual range" condition would be a reasonable compromise. I don't think any of the abilities work from squad sight. Could make the condition honour squadsight setting maybe.

robojumper:
I don't see much wrong with changing it to trade for performance. I see four options, in order of my preference:

1. Merge this gated behind a config flag so that users or mods that know what they're doing can opt into this perf improvement without having the CHL change it by default. Document the issue and the tradeoffs.
2. Merge this with an opt-out config flag so that users who don't want this behavior change can opt out. Document the issue and the tradeoffs.
3. Don't merge this change; instead make it the responsibility of mod authors that increase enemy counts and skill availability to add these conditions in OPTC.
4. Merge this without a config flag, give everyone slightly and lwotc massively better performance. Document the issue and the tradeoffs.

Xymanek:
Removing this from v1.22.0 milestone due to concerns mentioned above and the subsequent discussion on discord (wrt possible alternatives)

Iridar:
Personally I'd be fine with any of the solutions suggested by robojumper, I'm just happy this performance issue was found.

I didn't realize that a visibility condition check could increase performance. I always thought vision checks were expensive, so normally I don't add it to my abilities when it's not necessary. I'll have to brush through my abilities, see which of them might need it.

Iridar (approved):

pledbrook:
I'm planning to test this fix and the alternative of using unit values in the next dev build of LWOTC. Will hopefully get some feedback from players as to which is generally preferable.

pledbrook:
As an addendum, the fix that involves testing for the existence of a unit value won't work for Dual Strike because the bondmate shot needs to be available before the ability is used, otherwise the Dual Strike ability icon simply doesn't appear. So I'm strongly leaning towards making all these follow-up-shot abilities require gameplay visibility.

Iridar:
Personally I'm fine with adding a gameplay visibility condition. Yes it can result in altering the base game behavior, and it's controversial, because while, say, bondmate dual strike is visualized in sequence, the fantasy - I assume - is that they happen at the same time. So it would be weird while the primary shot hit, but the follow-up didn't because the target became obscured due to environmental destruction.

On the other hand, it fixes a performance drain experienced daily by literally everybody, granted that it's not instantly noticeable when playing with regular-sized pods. 

Yes, this implementation of a bugfix is on the edge of modifying base game behavior, but IMO it's (A) worth it, (B) arguably makes sense anyway, because neither Rapid Fire nor Dual Strike *actually* visualize at the same time.

Xymanek:
👎 from me - you are causing the player to lose an ability charge/cooldown due to circumstances entirely outside of player control

Iridar:
> As an addendum, the fix that involves testing for the existence of a unit value won't work for Dual Strike because the bondmate shot needs to be available before the ability is used, otherwise the Dual Strike ability icon simply doesn't appear. So I'm strongly leaning towards making all these follow-up-shot abilities require gameplay visibility.

Perhaps, modifying that behavior is not out of the question?

Iridar:
Just as a point of information for potential lurkers, a standalone bugfix mod for this issue has been out for nearly a year: https://steamcommunity.com/sharedfiles/filedetails/?id=2847640303

Iridar:
Archiving this due to lack of attention, but the referenced issue is still currently relevant.

PR #1024, opened 2021-06-16
Add event for overriding psi training times (#1016)

### LINKED ISSUE #1016: Allow mods to influence Psi training times
LW2 achieved this via a "PsiTrainingBegin" event. This could also be used by any mod that's interested in when a psi training project starts, irrespective of whether they want to modify the training times or not.

pledbrook:
This adds a `OverridePsiTrainingProjectPoints` event, modelled on the existing `OverrideRespecSoldierProjectPoints` event, that allows mods to override the initial training time for psi operatives.

Resolves #1016.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersProjectPsiTraining.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersProjectPsiTraining.uc
@@ -37,11 +37,15 @@ function SetProjectFocus(StateObjectReference FocusRef, optional XComGameState N
 		// Randomly choose a branch and ability from the starting two tiers of the Psi Op tree
 		iAbilityRank = `SYNC_RAND(2);
 		iAbilityBranch = `SYNC_RAND(2);
-		ProjectPointsRemaining = CalculatePointsToTrain(true);
+		// Issue #1016
+		/// HL-Docs: ref:OverridePsiTrainingProjectPoints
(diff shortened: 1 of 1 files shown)


Iridar (approved):
Nitpicking, but I'd prefer if the docs mentioned something about converting points into days, e.g. if I want for training to take X days what should I set to `ProjectPoints`. 

Also, `SetProjectFocus()` has a Game State as an optional argument, so in theory the override trigger could pass it with the event, not sure there's any use for it, though.

pledbrook:
I oscillated over whether to pass `NewGameState`, but decided in the end to stay consistent with the [event for respeccing soldiers](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/625).

Good point about explaining project points. Providing the default points per hour or per day should be sufficient. Worth bearing in mind that staffed scientists increase the points per hour.

Iridar:
> I oscillated over whether to pass `NewGameState`, but decided in the end to stay consistent with the [event for respeccing soldiers](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/625).
> 
> Good point about explaining project points. Providing the default points per hour or per day should be sufficient. Worth bearing in mind that staffed scientists increase the points per hour.

Yeah something along the lines would be worth mentioning in the docs.

pledbrook:
Updated the docs to explain how project points relate to time.

This should be squashed when merged.

Iridar (approved):

pledbrook:
This needs more work. The current implementation has no effect on the displayed number of training days in various dialogs, in particular in `UIChoosePsiAbility`, which simply replicates the code in the Psi Training project class 😢

Iridar:
Since no progress has been made for ~two years, I'm archiving this.

PR #1028, opened 2021-06-20
Reorder evac of carrying and carried units (#1027)

### LINKED ISSUE #1027: Carried VIPs evac'ed last don't register as evac'ed
As explained in [this issue](https://github.com/long-war-2/lwotc/issues/1342#issuecomment-863226330), if the soldier carrying a VIP is evacuated last from a mission, the VIP is not correctly registered as being recovered. The problem is that the Kismet for "team out of playable units" is triggered as soon as the soldier's evac is processed, which happens before the VIP is processed. So the mission ends and the VIP is treated as having been left behind on the mission.

Proposed solution is to modify `XComGameState_Unit.EvacuateUnit()` so that it fires the "UnitEvacuated" event for the carried unit before it fires the event for the carrier.

pledbrook:
In order to ensure that Kismet properly handles units that are carried when evacuated, this change processes the evacuation of the carried unit first. Note that this only changes the `SeqEvent_EvacActivated` Kismet node.

Fixes #1027.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqEvent_EvacActivated.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqEvent_EvacActivated.uc
@@ -1,0 +1,19 @@
+class SeqEvent_EvacActivated extends SeqEvent_GameEventTriggered;
+
+function EventListenerReturn EventTriggered(Object EventData, Object EventSource, XComGameState GameState, Name InEventID, Object CallbackData)
+{
+	// fire an evac event for all units in the associated state. This will normally be just the evac'ing unit,
+	// but it's possible he is carrying someone as well
+	foreach GameState.IterateByClassType(class'XComGameState_Unit', RelevantUnit)
+	{
+		CheckActivate(RelevantUnit.GetVisualizer(), none);
+	}
+
+	return ELR_NoInterrupt;
+}
+
+DefaultProperties
+{
+	ObjName="Evac Activated"
+	EventID="EvacActivated"
+}
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqEvent_EvacActivated.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqEvent_EvacActivated.uc
@@ -2,12 +2,34 @@ class SeqEvent_EvacActivated extends SeqEvent_GameEventTriggered;
 
 function EventListenerReturn EventTriggered(Object EventData, Object EventSource, XComGameState GameState, Name InEventID, Object CallbackData)
 {
+	// Variables for issue #1027
+	local array<XComGameState_Unit> EvacUnits;
+	local int i;
+
 	// fire an evac event for all units in the associated state. This will normally be just the evac'ing unit,
 	// but it's possible he is carrying someone as well
 	foreach GameState.IterateByClassType(class'XComGameState_Unit', RelevantUnit)
 	{
+		// Issue #1027: store in a local array so we can process the units
+		// in a different order.
+		EvacUnits.AddItem(RelevantUnit);
+	}
+
+	// Start Issue #1027
+	/// HL-Docs: ref:Bugfixes; issue:1027
+	/// VIPs that are carried by the last soldier to evacuate will now register
+	/// as having evacuated, which is particularly important for the Neutralize
+	/// VIP mission as it will grant the bonus reward.
+	//
+	// Process in reverse order so that the unit activating the ability
+	// is handled last. This ensures that carried units - such as VIPs -
+	// are processed before the carrier in the Kismet.
+	for (i = EvacUnits.Length - 1; i >= 0; i--)
+	{
+		RelevantUnit = EvacUnits[i];
 		CheckActivate(RelevantUnit.GetVisualizer(), none);
 	}
+	// End Issue #1027
 
 	return ELR_NoInterrupt;
 }


Iridar (approved):
Looks good to me, but are we absolutely sure the IterateByClassType will give us state objects in the order they were added to the game state? And if so, perhaps we could change the place in code where evacing units are being added to the game state instead?

pledbrook:
Unfortunately (or fortunately?) the soldier whose evac ability gets activated is guaranteed to be the first one in the new game state. I tried changing the order that units were added to the game state in `XCGS_Unit.EvacuateUnit()`, but it had no effect because of that.

pledbrook:
> Looks good to me, but are we absolutely sure the IterateByClassType will give us state objects in the order they were added to the game state? And if so, perhaps we could change the place in code where evacing units are being added to the game state instead?

I looked into being defensive and ordering the evac based on whether a unit is being carried or not, but the "carried" effect is removed by the time we get to this code. We could infer that unconscious units are being carried, so order based on that, but I don't know if that's any more reliable than assuming game state objects are iterated over in the order that they are added to the game state.

@tracktwo Any thoughts?

tracktwo:
Maybe check the unit template for hostile VIP (or, any non-soldier) and order those first?

This really is at it's root a kismet bug in the neutralize VIP mission, but that's harder to fix at the root in the highlander. Specifically patching for this particular bug on this mission type seems reasonable to me, rather than trying to define a more general solution. Anyone else writing custom missions that rely on order of extraction should write kismet that doesn't have bugs like this. I'd only suggest going further with the generalization if there are existing modded missions out there that copied the same bug from Neutralize VIP and are no longer being updated.

pledbrook:
> This really is at its root a kismet bug in the neutralize VIP mission

Is that true? I'm just wondering what happens in cases where a VIP on a rescue or extract mission is carried out. Would they be awarded at the end of the mission with the current code? And since modders will often just copy base-game stuff, one can expect the same sorts of issues to appear in modded missions.

My gut feeling is that ensuring EvacActivated is fired for carried units first will simplify the Kismet for missions. hence why I'd prefer a more general solution rather than one narrowly focused on the Neutralize VIP mission. That said, I still don't understand how the reward code works, or whether "Team out of playable units" fires in between the EvacActivated events for carried and carry units.

Xymanek (changes requested):
I'm 👎 for relying on iteration order of states inside a frame. There are no documented guarantees, it's not used anywhere like that and reordering some preceding logic (or having a just "right" listener from a mod) can easily cause this "fix" to fall apart

tracktwo:
> Is that true? I'm just wondering what happens in cases where a VIP on a rescue or extract mission is carried out.

I can't definitively say such a bug doesn't exist, but the evidence I have against it is I've seen the neutralize VIP bug reported regularly on evac all after the bug was introduced (with DLC3) but never any other similar bug for other mission types. The first bug report I got on evac all for this was 2016-06-30, which is the same day SLG released, and it blew up the mod page for a bit. Similarly I "fixed" the bug with a patched kismet sequence in the original evac all, and after that the bug stopped being reported, except for people using mods that also replaced that mission map (e.g. to extend/remove timers) and reintroduced the bug.

My initial analysis of this bug was back in that time frame and in vanilla, so it's also possible that I'm misremembering details and/or if WoTC has changed things, but IIRC the reward for the hostile VIP (and others) is based only off whether the corresponding objective is marked as complete in kismet and the unit state is not 'dead' (`XCGS_MissionSite.GetRewardVIPStatus`). It's specifically relevant for neutralize VIP missions because the _only_ way to get a hostile VIP out is to carry them, extract VIPs are player-controlled on the XCOM team and so don't need to have kismet checks to make sure there are other units still around able to extract the unit. The path in kismet for setting this strategy objective as complete is short-circuited and ends without marking the objective complete if the Mission.VIPKilled variable is true, and this variable is set when the last soldier evacs. Killing the VIP when the last XCOM unit extracts was added to avoid mission failure in SLG: if you extract all your humans but leave the VIP on the ground while still having SPARKs you've hard failed the mission because sparks can't pick up the VIP, but they also can't shoot the unconscious VIP to complete the 'kill' objective.

I've lived with this bug for over 5 years now, and the current workaround in Evac All is 'good enough' for me in that it solves the vast majority of cases, but it surely would be nice to have this fixed completely in the highlander and it seems to me to fit the highlander mandate of fixing vanilla bugs.

pledbrook:
I'm currently considering whether to check for "unconscious" or "hostile VIP". I'm pretty sure "unconscious" is the only state that allows living units to be carried. It covers the Neutralize VIP mission while providing cover for other mission types with similar problems that are perhaps just more rarely encountered. Any downsides to that approach that folks can see?

Iridar:
> I'm currently considering whether to check for "unconscious" or "hostile VIP". I'm pretty sure "unconscious" is the only state that allows living units to be carried. It covers the Neutralize VIP mission while providing cover for other mission types with similar problems that are perhaps just more rarely encountered. Any downsides to that approach that folks can see?

Unless I'm missing something, a simple good solution would be something like:

```
foreach GameState.IterateByClassType(class'XComGameState_Unit', RelevantUnit)
{
	if (RelevantUnit.IsBeingCarried())
	{
		CarriedUnits.AddItem(RelevantUnit);
	}
	else
	{
		OtherUnits.AddItem(RelevantUnit);
	}
}

foreach CarriedUnits(RelevantUnit)
{
	CheckActivate(RelevantUnit.GetVisualizer(), none);
}

foreach OtherUnits(RelevantUnit)
{
	CheckActivate(RelevantUnit.GetVisualizer(), none);
}
```

Iridar:
Personally I'd prefer some kind of definitive check of "Was unit being carried when Evac was activated".

Xymanek:
Just a reminder that the history system allows looking all the way back to the start of the battle

pledbrook:
If anyone can provide the code for getting an appropriate history index, for example the start of the turn on which the carried unit was evac'ed out on, then I'll happily add it in. Can't work it out myself at the moment.

Iridar:
Since the author has bSince no progress has been made for ~two years, I'm archiving this.een AWOL for over two years, I'm archiving this.

PR #1030, opened 2021-06-22
Use writable state for units spawned from Avenger (#1029)

### LINKED ISSUE #1029: Units spawned from Avenger sometimes don't have abilities initialised
`SeqAct_SpawnUnitFromAvenger.AddStrategyUnitToBoard()` has an issue where it submits a game state just before initialising a unit's abilities. This sets the unit to read-only mode. Robojumper says this is normally fine (hence why most players don't encounter an issue with units that spawn from the Avenger), but if a mod or some bit of base-game code calls `ModifyGameState()` on the unit before the abilities are initialised, then those changes to the old copy are discarded when the game state is submitted. At least that's my understanding.

The fix is to ensure that `AddStrategyUnitToBoard()` calls `ModifyStateObject()` on the new game state for the unit that's being added to the board.

This fixes a [recent LWOTC issue](https://github.com/long-war-2/lwotc/issues/1352) that was reported.

pledbrook:
This specifically fixes the issue where a non-writeable state is currently used for the initialisation of unit abilities. While it mostly
works as it is, there are apparently circumstances in which the changes to the old, read-only unit state are discarded, meaning the abilities don't get initialised properly.

Fixes #1029.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqAct_SpawnUnitFromAvenger.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/SeqAct_SpawnUnitFromAvenger.uc
@@ -197,6 +197,16 @@ private static function XComGameState_Unit AddStrategyUnitToBoard(XComGameState_
 	// Must happen after unit is submitted, or it gets confused about when the unit is in play or not 
 	NewGameState = class'XComGameStateContext_ChangeContainer'.static.CreateChangeState("Adding Reserve Unit Abilities");
 
+	// Start Issue #1029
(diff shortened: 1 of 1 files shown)


Iridar:
Since this is a simple bugfix that 99.99% cannot cause any negative side effects, I'm inclined to approve and merge it right away, but I'd like to confirm if this was already tested to actually fix the issue it is supposed to fix?

pledbrook:
It was tested against the save provided in [the originating LWOTC issue](https://github.com/long-war-2/lwotc/issues/1352).

Xymanek (approved):

PR #852, opened 2020-05-01
Add an event to override XCGS_Unit::HasHeightAdvantageOver

Iridar:
Fixes #851 

Already includes HL Docs. Tested in-game to work correctly.

Issue #851, opened 2020-05-01
Add an event to override XCGS_Unit::HasHeightAdvantageOver


Iridar:
I'd like to put an event in the mentioned function so that my Jet Shot can benefit from various abilities that trigger if the shooter has a height advantage, such as Death From Above. After all, the whole point of that ability is to provide a height advantage for the shooter, which is currently done through a hit modifier.

Does `OverrideHasHeightAdvantageOver` work as the event name?

Xymanek:
Can't you change the z position of the unit? Like the Blazing Pinions work?

Iridar:
> Can't you change the z position of the unit? Like the Blazing Pinions work?

Maybe I could, but I would have to redo everything about that ability from scratch, including animations. It would also make it more complicated, because I would have to split flying up and landing into separate abilities, because some perks that check height advantage fo it in post ability cost paid. It would be a massive mess. Doing this through an event has the advantage of also allowing to make a perk that denies height advantage to the shooter.

PR #1032, opened 2021-07-07
Fix moving melee hangs

### LINKED ISSUE #1031: Game lock when using Rage strike
Game locks when trying to use the rage armour. I use the BETA highlander and the DLC2 plugin. It locks after actually using the attack. You can still see the yellow squares where you can choose which to attack from. To get out of it I have to control alt delete sign out and then cancel it or resign in. I have attached the aml mod list and save file. It is currently equipped on an akimbo class
[AML.txt](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/files/6758673/AML.txt)
[save_AUTOSAVE- Campaign 40, Mission 66, Turn 7.zip](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/files/6758689/save_AUTOSAVE-.Campaign.40.Mission.66.Turn.7.zip)

Didn't happen before the CHL BETA update

Iridar:
Closes #1031 by reverting #852 and adding a warning to `X2AbilityToHitCalc_StandardAim` about the dangers of triggering events in ToHitCalc code. 

Implements #851 through the delegate system instead.

Fixed a bug in #855 (first implementation of the delegate system) and updated its docs while I was a it.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
@@ -247,13 +247,27 @@ struct ShouldDisplayMultiSlotItemInTacticalStruct
 		Priority = 50
 	}
 };
-
 var protectedwrite array<ShouldDisplayMultiSlotItemInStrategyStruct> ShouldDisplayMultiSlotItemInStrategyCallbacks;
 var protectedwrite array<ShouldDisplayMultiSlotItemInTacticalStruct> ShouldDisplayMultiSlotItemInTacticalCallbacks;
-
-delegate EHLDelegateReturn ShouldDisplayMultiSlotItemInStrategyDelegate(XComGameState_Unit UnitState, XComGameState_Item ItemState, out int bDisplayItem, XComUnitPawn UnitPawn, optional XComGameState CheckGameState);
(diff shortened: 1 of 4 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
```diff
@@ -852,6 +842,84 @@ static function CHHelpers GetCDO()
...
+		{
+			OverrideHasHeightAdvantageCallbacks.Remove(i, 1);
+			return true;
+		}
+	}
+	return false;
+}
+
+// Called by XComGameState_Unit::HasHeightAdvantageOver()
+simulated function TriggerOverrideHasHeightAdvantage(XComGameState_Unit Attacker, XComGameState_Unit TargetUnit, out int bHasHeightAdvantage)
```
Xymanek, line 906: Please add a comment that this is an internal CHL API and is not covered by BC policy

Xymanek (commented):
I have a few nitpicks around docs, but they are minor so let's resolve that later.

Sans the BC comment, looks good overall. Assuming tested (cuz reviewing this in depth is a pain 😄)

X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
Iridar: Цшдд вщю

Iridar:
Should be good to go now, but I'd appreciate one final look.

Iridar:
It's not absolutely necessary though, I really did well test it this time.

PR #733, opened 2020-01-20
Unrestrict Psi PCS to PsiOperative class (#602)

boundir:
Hope I did everything right this time 😅

Issue #88, opened 2017-10-21
Remove hard-coding referring to PsiOperative class in XComGameState_Unit and Allow other classes to be treated as Psionic


zarquonhigardi:
There are several mods on the workshop that recreate the Psi-Operative as a new class in order to allow them to rank up like normal units. However, the Chosen (and likely other things) react to specifically the PsiOperative class. One mod (which seems to have been removed by the author) fixed this by overriding several functions in XCOMGameState_Unit

Several functions in XComGameState_Unit function refer to PsiOperative class specifically in order to do things such as the Hair-Coloring and Random Perk Allocation (RankUpSoldier), and Preventing the soldier from ever gaining AP or ranking up normally (HasAvailablePerksToAssign).

Another function, IsPsionic(), checks for this class as well. As does IsPsiOperative()

Ideally, a fix would allow a unit to be treated as a PsiOperative in all respects **and** allowing the PsiOperative to gain AP and rank up normally.

BTernaryTau:
A hook for overriding the output of HasAvailablePerksToAssign() can be taken from the vanilla highlander: https://github.com/X2CommunityCore/X2CommunityHighlander/pull/202/files

robojumper:
> Ideally, a fix would allow a unit to be treated as a PsiOperative in all respects and allowing the PsiOperative to gain AP and rank up normally.

This is worded very ambiguosly. If it gains XP and ranks up normally, it *by definition* isn't treated like a Psi Operative in all respects.

zarquonhigardi:
Sorry, I meant in all other respects the PsiOperative is treated as... a PsiOperative while having the ability to gain XP, Ranks, and assign perks normally. 
IE: Separating being Psionic/PsiOperative from ranking up normally and gaining AP normally.

VampireMonkey:
Many places around the code it simply checks if the soldier's classname is Psi Operative, changing those to ex: use the bHasPsiGift property that's already define in the codebase could be a nice way to remove this weird check.

bHasPsiGift is not used anyways.

Iridar:
To catch up on this issue.

There were a lot of attempts to address it.

#604 suggested that any unit with `eStat_PsiOffense > 0` should count as psionic, the PR was even merged, but then reverted, because it breaks with LWOTC.

#673 suggested to add a configurable list of soldier class templates that count as psionic classes, it was merged, but then reverted, as it had issues keeping its scope coherent. 

[...]

PR #1033, opened 2021-07-09
Revert PR #733 (Issue #602)

Iridar:
As discussed in #88, this PR should be reverted while it can still be done without breaking backwards compatibility, and a more universal solution will be put in its place. Author of the original PR has given their blessing on this.

Specific commit being reverted is: https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/commit/4c381f453d3ac0a9aa55383e73affdd3b70a69aa


Iridar:
This looks clean to me line for line, so I'll be merging this right away.

PR #1041, opened 2021-07-18
Add straightforward introduction to run order popups

### LINKED ISSUE #1040: Make run order popups easier to understand for users
Context: https://discord.com/channels/165245941664710656/273238884433788928/866298353049796608

LW2SW + RPGO conflict popup

We've been getting questions from confused users and want to make adjustments to this class of popups.

* This particular warning should be yellow instead of red to indicate that it's not a critical error (unlike the cycle error).
* Run order errors should give clearer instructions to users, including that their only course of action is to report the issue to mod authors (and click "don't show again" if possible)
* In particular, indicate that the issue cannot be fixed by messing with what they believe to be load order.

Code pointers:

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/fb6622d26fdae062153f9bad20b692a69705b1b1/X2WOTCCommunityHighlander/Src/X2WOTCCommunityHighlander/Classes/X2WOTCCH_UIScreenListener_ShellPopup.uc#L148-L236

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/fb6622d26fdae062153f9bad20b692a69705b1b1/X2WOTCCommunityHighlander/Localization/X2WOTCCommunityHighlander.int#L20-L22 (Run order errors only have INT localization)

robojumper:
Fixes #1040.

This is the RPGO + LW2SW case:
groups popup

And this is the cycle case, which we haven't observed in the wild yet:
cycle popup

### DIFF
--- a/X2WOTCCommunityHighlander/Localization/X2WOTCCommunityHighlander.int
+++ b/X2WOTCCommunityHighlander/Localization/X2WOTCCommunityHighlander.int
@@ -15,18 +15,17 @@ ModIncompatible="Please DEACTIVATE the following mods and restart or %s will not
 DisablePopup="DO NOT SHOW ME THIS AGAIN"
 
 [X2WOTCCH_UIScreenListener_ShellPopup]
-CycleErrorTitle="Cycle in Run Order detected"
-GroupErrorTitle="Conflicting Run Order data"
+ErrorTitle="Mod Configuration Error"
 
-HardErrorText="Things will behave unpredictably. Proceed at your own risk."
-PotentialErrorText="Things will behave predictably, but potentially not in the way the author intended."
+AdviceText="The following mods have incorrect configuration: %BLAME. Show the error report to the respective authors to help them fix it."
 
-AdviceText="If you are a mod user, please forward this message to the authors of the following mods: %BLAME. Only they can fix the cause of the problem."
+HardErrorText="You can continue playing but mods will behave unpredictably. Proceed at your own risk."
+PotentialErrorText="You can continue playing and mods will behave predictably, but potentially not in the way their authors intended."
 
 CycleErrorText="Mods specified a cyclic run order:\n\n%FACTS\n...completing the cycle. Until this is corrected, run order will be undefined."
 GroupErrorText="Mods specified conflicting run order information:\n\n%FACTS\nThis behavior is specified, but may indicate a configuration error."
 
-BlameText="The following DLCIdentifiers provided config that lead to this problem: %BLAME"
+BlameText="The following DLCIdentifiers provided config in XComGame.ini that lead to this problem: %BLAME."
 
 [X2WOTCCH_UIScreenListener_Shell_DLCIntro]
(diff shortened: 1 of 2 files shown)


Xymanek:
Could we use the `DisplayName` instead of the DLCInfo class name? Should be much easier to understand for players

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/fb6622d26fdae062153f9bad20b692a69705b1b1/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHModDependency.uc#L35

robojumper:
Maybe? This is a different system though (run order vs dep checker)

Xymanek:
Right, I keep forgetting that they are separate systems, since whenever there are changes they are usually done to both at the same time 😐

Iridar (approved):

Iridar:
I think the text can still use some adjustment to make it more friendly to regular mod users. 

"Conflicting Run Order data" wouldn't mean anything to them; "Things will behave predictably, but potentially not in the way the author intended" - the question comes to mind: what author? Author of what? This hasn't been established in the text. 

I propose replacing the `"Conflicting Run Order Data"` title with `"Mod Configuration Error"`, so that the mod user that sees this can tell the problem is with mods right away. The `"Conflicting Run Order Data"` error title can be moved somewhere deeper into the message's body or dropped entirely.

I also suggest removing the `"Things will behave predictably, but potentially not in the way the author intended"` sentence entirely, as it's likely to be more confusing than helping. Or at least moving it later into the text.

Also, "Only they can fix the cause of the problem" is technically not true, since anybody can go and change DLC run order config, in theory.

Xymanek:
> "Things will behave predictably, but potentially not in the way the author intended" - the question comes to mind: what author? Author of what?

Just change it to `mod author` to be super specific?

> The "Conflicting Run Order Data" [...] can be [...] or dropped entirely

I disagree with dropping it entirely - it quickly presents the type of the issue, which is important when dealing with problems. It can probably just be prefixed with `[Mod config]` for a balance between conveying the type and the source of the problem, while keeping the header manageable

> I also suggest removing the "Things will behave predictably, but potentially not in the way the author intended" sentence entirely, as it's likely to be more confusing than helping

While the wording could be improved, I disagree with removing it - this is very important info. Just like with `none accessed` warnings in logs - the behaviour will stay the same as long as nothing else changes, but it's a good indication that something isn't working the way it's supposed to and under different circumstances might break

> Also, "Only they can fix the cause of the problem" is technically not true, since anybody can go and change DLC run order config, in theory.

While you are technically correct, practically it is impossible as it would require understanding about code (likely of 2+ mods) that an average player does not possess

robojumper:
Updated based on Discord discussion.

cycle error popup

group error popup

Note that the lower popup will actually use the names "XCOM2RPGOverhaul" and "WOTC_LW2SecondaryWeapons" in practice, so it should be descriptive enough for people to figure out which mods are meant.

Iridar:
Thanks, looks good to me now.

Xymanek (approved):

PR #1043, opened 2021-07-20
Fix some identified performance issues

### LINKED ISSUE #1042: Lag after performing actions
[`8fb9869` (#259)](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/259/commits/8fb98698e3503b24f9b24dd17ab18807284a77ad) (one commit part of #259).

grafik

grafik

In order to update focus reliably, we subscribe to OnVisualizationBlockFinished. This causes a lot of UI updates. 2s is excessive due to another mod, but we're still causing 24ms of flash UI calls.

robojumper:
Fixes #1042.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_SoldierInfo.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_SoldierInfo.uc
@@ -7,7 +7,7 @@
 //  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
 //--------------------------------------------------------------------------------------- 
 
-class UITacticalHUD_SoldierInfo extends UIPanel implements(X2VisualizationMgrObserverInterface); // Issue #257 -- correct update calls
+class UITacticalHUD_SoldierInfo extends UIPanel;
 
 var string HackingToolTipTargetPath;
 var localized string FocusLevelLabel;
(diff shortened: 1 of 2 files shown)


Xymanek (commented):
Confirmed reduces (removes) the flashing (updating) of the unit name (bottom left) while just running around

X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_SoldierInfo.uc
```diff
@@ -112,7 +110,7 @@ simulated function UpdateStats()
 	else
 	{
 		//UITacticalHUD(Screen).m_kInventory.m_kBackpack.Update( kActiveUnit );
-		if( LastVisibleActiveUnitID != kActiveUnit.ObjectID )
+		if( /*LastVisibleActiveUnitID != kActiveUnit.ObjectID*/ true ) // Issue #257, this is also called when refreshing after action
```
Iridar, line 113: Could we win some extra CPU cycles here by commenting out the redundant `if (true)` evaluation, or would the compiler strip it out automatically?

robojumper: The compiler compiles this literally without any dead code elimination or constant propagation. The performance impact is negligible and basically cannot be measured outside of a microbenchmark, so I'd keep it in to keep the diff small (this wasn't the cause of any measured performance issues).

Xymanek (approved):

Issue #1044, opened 2021-07-21
Add self-profiling system to `UpdateAbilityAvailability`


robojumper:
Mods are running into bugs where some abilities, their conditions and targets take way too long to evaluate. A rudimentary self-profiling system would go a long way to identify issues where the call stack debugger reaches its limits (since it cannot tell abilities apart).

PR #1046, opened 2021-07-21
Simple ability self profiler

robojumper:
Helps towards #1044.

@Xymanek suggested:
> add an array to CHHelpers for ability templates to force usage of script GAT (XCGS_A_CH), so that we can see what exact part of logic is spending so much time. The DoesBlockUnitPathing calls makes me suspect that we are actually spending the time in X2AbilityTarget_MovingMelee/pathfinding


robojumper:
Flash of inspiration: We don't strictly need `XCGS_A_CH` for profiling -- we can just add a denativized `GatherAbilityTargets` function with a different name that's called for specific abilities instead. We can even add code to record the number of initial targets, filtered targets etc.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Ability.uc
```diff
@@ -384,6 +393,105 @@ simulated function int SortAvailableTargets(AvailableTarget TargetA, AvailableTa
 }
 */
 
+simulated function name CHGatherAbilityTargetsScript(out array<AvailableTarget> Targets, optional XComGameState_Unit OverrideOwnerState)
```
robojumper, line 396: Note to self: This should be `final`.

PR #1055, opened 2021-08-07
Fix the if() check in XComGameStateContext_TacticalGameRule::ContextBuildVisualization()

### LINKED ISSUE #1054: Bugged logic in XComGameStateContext_TacticalGameRule::ContextBuildVisualization()
[Relevant code](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/de17827ec72dd9662170c96b8f74c41d534bf212/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameStateContext_TacticalGameRule.uc#L493)

The `if()` condition can never succeed, so the `class'X2Action_BeginTurnSignals'.static.AddBeginTurnSignalsToBlock(self);` never runs.

Iridar:
Closes #1054


Iridar:
Note that without this fix, the condition could *never* succeed, so frankly we got no idea what this fix will entail. We might be unleashing a kraken.

Iridar:
Closing until some proper testing / code tracing is done, no reason to fly blind.

PR #1072, opened 2021-09-04
Delay PSC return to pool by 1 min in several cases

### LINKED ISSUE #1061: Projectile Particle System Components are cleaned up too soon
While fixing #720 by #1013, we created another problem: all spawned particle effects are cleaned up the moment projectile hits.

For comparison, here are the videos of firing the heavy weapon rocket launcher in CHL 1.21 and CHL 1.22:

* 1.21: https://youtu.be/gyUuhWBpsvA
* 1.22: https://youtu.be/Hm4xz083wjI

Notice how in the second video the rocket's smoke trail abruptly disappears the moment the rocket hits. 

This is only a visual bug, but a pretty egregious one. By fixing the "memory leak" we essentially broke *all* projectile trails.

Xymanek:
Fixes #1061

TODO:

* [x] Expose the delay to configuration
* [x] Better docs
* [x] Test

cc @Iridar

### DIFF
--- a/X2WOTCCommunityHighlander/Config/XComGame.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGame.ini
@@ -246,3 +246,7 @@ bEnableVersionDisplay=true
 ;	they will display an Overwatch "eye" icon under their unit flag.
 +ValidReserveAPForUnitFlag = "overwatch"
 +ValidReserveAPForUnitFlag = "pistoloverwatch"
+
+;;; HL-Docs: ref:ProjectilePerformanceDrain
+; The following line is an example/template. See feature documentation for more info.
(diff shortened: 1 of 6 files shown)


Iridar:
I tested a bunch of different weapons and everything seems to be in order.

Xymanek:
PR ready & tested on my side, although I was testing only the "framework" itself - @Iridar I suggest you test different weapons again just in case, since technically there some changes to the code.

When merging, this should be squashed.

Iridar (approved):
Reviewed and tested, everything looks good. The only nitpick I could come up with is the commented out logs, but I don't mind them being there, they might come in handy if somebody decides to test it locally later without polluting the log output for those that don't want to test this specific feature.

Xymanek:
The log spam is very real, which is why I left it commented out by default

PR #1075, opened 2021-10-03
Adding new inventory slot for Viper spit items

SV1002:
Adds one new entry allowing the use of spits for vipers without using a utility slot (And consequently, allowing Non-Vipers equipping the item) and allowing it to be switched out for a different item that grants a different type of spit for them.


Iridar:
This PR is okay, though I'm still not convinced you actually need an inventory slot for this. You can attach different types of spits to different abilities with different perk contents.

Iridar (approved):

Xymanek:
If it's possible to achieve the desired behaviour without adding a slot, then it's better to avoid adding a slot. The number of slots we can add is plenty, but not infinite

Xymanek:
What's the status of this PR? Is it still required or a different approach was used?

SV1002:
Still waiting for the item slots to be added, yes. I have plans to make new items for said slot in future once it's made available.

Xymanek (approved):

PR #1087, opened 2021-11-05
Issue #1086 - Added 'none' check to X2Item_DefaultUpgrades::AreUpgradesEmpowered()

### LINKED ISSUE #1086: X2Item_DefaultUpgrades::AreUpgradesEmpowered() needs 'none' check
```
static function bool AreUpgradesEmpowered()
{
	local XComGameState_HeadquartersXCom XComHQ;

	XComHQ = XComGameState_HeadquartersXCom(`XCOMHISTORY.GetSingleGameStateObjectForClass(class'XComGameState_HeadquartersXCom'));
	return XComHQ.bEmpoweredUpgrades;
}
```

This function throws loud redscreens when XComHQ doesn't exist, which will happen if this check happens to be performed in the Shell. This is normally not a problem, but if a mod allows equipping items in Character Pool (like my Appearance Manager), there will be horrible log spam.

Iridar:
Closes #1086 

Technically this prevents the game reporting about what would be a critical error in an actual campaign, so concerns about removing that are understandable. 

No HL-Docs because none seem appropriate, but let me know if I need to add some.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Item_DefaultUpgrades.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Item_DefaultUpgrades.uc
@@ -1129,6 +1129,11 @@ static function bool AreUpgradesEmpowered()
 {
 	local XComGameState_HeadquartersXCom XComHQ;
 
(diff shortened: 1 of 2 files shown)


Xymanek (approved):
I'm not a fan of suppressing real errors, but rj made a good point - if XComHQ is missing in an actual campaign, the amount of things that will outright break is massive (the game will likely crash)

PR #1100, opened 2021-12-02
Add a different trunctuation method to damage modifier hooks

### LINKED ISSUE #1099: Add An option to Truncate The increase/decrease instead of the result in Highlander Damage Modifiers
Non-Hook %damage modifiers were effectively trunctuated, making them weak for lower values but strong for higher
Current Hook trunctuates the result, which makes it very inconsistent to how %DR modifiers were behaving previously. 

In my personal opinion that makes %DR way too hard to balance because while previously it was weaker than armor at lower damage values, currently even 1% DR is more effective than 1 armor which is quite silly.

I propose a config option that changes the behavior so backwards compatibility can be preserved.

Grobobobo:
Fixes #1099

### DIFF
--- a/X2WOTCCommunityHighlander/Config/XComGameCore.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGameCore.ini
@@ -28,3 +28,7 @@
 ; Set to false/commented out if you want damage to hit shields/ablative, then armor, then health (vanilla behaviour)
 ; Set to true/uncomment it if you want damage to hit armor, then shield/ablative, then health
 ;ARMOR_BEFORE_SHIELD=true
+
+; Issue 1099
+; Set to true if you want pre and post default damage modifiers to trunctuate the difference rather than the result
+;TRUNCTUATE_DIFFERENCE_INSTEAD_OF_RESULT = true
--- a/X2WOTCCommunityHighlander/Config/XComGameCore.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGameCore.ini
@@ -30,5 +30,5 @@
 ;ARMOR_BEFORE_SHIELD=true
 
 ; Issue 1099
-; Set to true if you want pre and post default damage modifiers to trunctuate the difference rather than the result
-;TRUNCTUATE_DIFFERENCE_INSTEAD_OF_RESULT = true
+; Set to true if you want pre and post default damage modifiers to truncate the difference rather than the result
+;TRUNCATE_DIFFERENCE_INSTEAD_OF_RESULT = true
--- a/X2WOTCCommunityHighlander/Config/XComGameCore.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGameCore.ini
@@ -28,7 +28,3 @@
 ; Set to false/commented out if you want damage to hit shields/ablative, then armor, then health (vanilla behaviour)
 ; Set to true/uncomment it if you want damage to hit armor, then shield/ablative, then health
 ;ARMOR_BEFORE_SHIELD=true
-
-; Issue 1099
-; Set to true if you want pre and post default damage modifiers to truncate the final damage modification rather than the final damage itself
-;TRUNCATE_FINAL_DAMAGE_MODIFIER = true
--- a/X2WOTCCommunityHighlander/Config/XComGameCore.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGameCore.ini
@@ -30,5 +30,5 @@
 ;ARMOR_BEFORE_SHIELD=true
 
 ; Issue 1099
-; Set to true if you want pre and post default damage modifiers to truncate the difference rather than the result
-;TRUNCATE_DIFFERENCE_INSTEAD_OF_RESULT = true
+; Set to true if you want pre and post default damage modifiers to truncate the final damage modification rather than the final damage itself
+;TRUNCATE_FINAL_DAMAGE_MODIFIER = true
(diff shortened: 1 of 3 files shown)


X2WOTCCommunityHighlander/Config/XComGameCore.ini
```diff
@@ -28,3 +28,7 @@
 ; Set to false/commented out if you want damage to hit shields/ablative, then armor, then health (vanilla behaviour)
 ; Set to true/uncomment it if you want damage to hit armor, then shield/ablative, then health
 ;ARMOR_BEFORE_SHIELD=true
+
+; Issue 1099
+; Set to true if you want pre and post default damage modifiers to truncate the difference rather than the result
+;TRUNCATE_DIFFERENCE_INSTEAD_OF_RESULT = true
```
pledbrook, line 34: I would prefer a name that provides a bit more context. Perhaps `TRUNCATE_FINAL_DAMAGE_MODIFIER`? I would also expand on the comment to explain that the setting "truncates the final damage modification rather than the final damage itself".

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
```diff
@@ -1245,6 +1248,14 @@ simulated protected function int ApplyPreDefaultDamageModifierEffects(
 	// handles float -> int conversion in most cases. Note that this may
 	// result in the sum of the shot modifiers not adding up to the overall
 	// damage modifier, but damage is generally shown as a range anyway.
+			
+	//Handle the option with the possibility of trunctuating the difference rather than result
+	if(class'X2Effect_ApplyWeaponDamage'.default.TRUNCATE_DIFFERENCE_INSTEAD_OF_RESULT)
```
pledbrook, line 1253:
```suggestion
	// Truncate the change in damage rather than the final damage itself if
	// this setting is enabled.
	if (class'X2Effect_ApplyWeaponDamage'.default.TRUNCATE_DIFFERENCE_INSTEAD_OF_RESULT)
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
```diff
@@ -1329,6 +1341,14 @@ simulated protected function int ApplyPostDefaultDamageModifierEffects(
 	// handles float -> int conversion in most cases. Note that this may
 	// result in the sum of the shot modifiers not adding up to the overall
 	// damage modifier, but damage is generally shown as a range anyway.
+
+	//Handle the option with the possibility of trunctuating the difference rather than result
+	if(class'X2Effect_ApplyWeaponDamage'.default.TRUNCATE_DIFFERENCE_INSTEAD_OF_RESULT)
```
pledbrook, line 1346:
```suggestion
	// Truncate the change in damage rather than the final damage itself if
	// this setting is enabled.
	if (class'X2Effect_ApplyWeaponDamage'.default.TRUNCATE_DIFFERENCE_INSTEAD_OF_RESULT)
```
```suggestion
	//Handle the option with the possibility of trunctuating the difference rather than result
	if(class'X2Effect_ApplyWeaponDamage'.default.TRUNCATE_DIFFERENCE_INSTEAD_OF_RESULT)
```

pledbrook (changes requested):
I'm seriously wondering whether to make this the default behaviour. The original behaviour was a mistake and having an option here will make it difficult for mods that provide abilities to handle both "modes". Yes, the change would break backwards compatibility in terms of behaviour and I'm worried about the implications of that. But I also don't think anyone other than LWOTC/-R are using it yet (that's not to say they aren't).

Even if they are using it, I can't imagine them not benefiting from the change, unless they explicitly balanced around the existing behaviour. It's a tricky decision, but I'm definitely leaning towards treating this as a bug in the highlander (and I accept full responsibility for that).

Grobobobo:
I agree that current state seems more like a mistake rather than the valid mode
It also seems like I'm the only one that changed values accordingly with the new damage calculation, so it's probable that there won't be any repercussions for breaking the backwards compatibility. Still, it's just a guess, so just to be sure I've added it as a seperate option.

Iridar:
Quote from discord discussion for context:

> The main problem was that the calculation behavior is different than the % damage modifiers from the normal attacking modifiers - the old one truncated the difference, while the new one truncates the final result
> say you have a 10% damage reduction. against 3 damage, the old one will truncate the difference (from -0,3 to 0) and it will still deal 3 damage, while the new one will truncate the result (from 2,7 to 2), causing the attack to deal 2 damage. This causes %DR to be significantly stronger overall, especially against lower damage values, and switching to it has caused some major balance issues a while back 
> The purpose of the hook was to make sure the damage result didn't depend on the order of effects - to always make flat damage modifiers go first, and % damage modifiers go second, the behavior change is not a desired effect

(c) Grobobobo

Truncating the difference between the original damage and modified damage allows to create a specific behavior:
- If the modified damage is lower than the original, it is effectively rounded up.
- If higher, it is rounded down. 

This fixes the balancing issue where the percentage-based damage reduction is unintentionally strong. 

Even though it will be technically a backwards compatibility break, **I recommend making this the default behavior, without a config bool toggle**, since arguably it's a necessary fix to otherwise broken functionality, and also because it's unlikely there are many (any) mods that currently use this feature.

Grobobobo:
Changed it so the difference truncation is now be the default behavior.

Iridar:
I'll probably do some tiny adjustments to docs when this will be merged, but other than that it's good to go, I think.

PR #1102, opened 2021-12-06
Resolves #1101

### LINKED ISSUE #1101: Weapon Upgrade with negative stat modifier shows as "--x"
Better explained here: https://imgur.com/a/9BzzAh7

This is caused by a minor bug in `XComGameState_Item::AddStatModifier`. This has never come up as an issue as 99% of weapon upgrades from base game and modded ones do not have negative stat modifier (or maybe the negative stat don't show up on this UI).

TeslaRage:
Resolves #1101
Minor change to only show "+" when `Value >= 0`
Will never include "-" as that is already in `Value`


Xymanek (changes requested):
Please remake this PR with a dedicated source branch (and fix the commits mess)

PR #1105, opened 2021-12-07
Add eInvSlot_ entries for modular sparks.

h4ilst0rm:
A massive amount of inventory slots (3) to allow a massive stack of equippable items for a SPARK class overhaul.
Will add a comment in #137 once the mod gets to the workshop.
image

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRulesetDataStructures.uc
@@ -112,6 +112,9 @@ enum EInventorySlot
 	eInvSlot_MZAux,
 	eInvSlot_TemplateMaster,
 	eInvSlot_ViperMAW,
+	eInvSlot_SparkComputer,
+	eInvSlot_SparkActuator,
+	eInvSlot_SparkModule,
 
 	// Marker slot, don't use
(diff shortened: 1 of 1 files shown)


Iridar:
Are you sure you can't accomplish your task by using utility slots or one multi-item slot?

h4ilst0rm:
This is at least conceptually the proper way to do it.

I have 3 separate sets of items and the 3 slots you can equip them in, which act as a replacement for the class skill tree and should stay separate from the regular utility items.  Two slots are large and take one item of their respective categories. One is a small multi-slot with varying quantity and does not take duplicates. I think using regular utility slots would likely cause problems if someone wants to grant them to sparks for the purpose of using utility items, and they would make the code unnecessarily noodlier with the unique restriction, varying amount, exclusivity for sparks and loadout validation spread out. (Or at least I have not seen any examples of such implementation so I don't know how to approach it.)

Using just one multi-slot instead would be a lot clunkier for the player to use, requiring knowledge which items are from which category so they can replace the correct one as the order shifts when items are changed in the slot. As far as I know there is no way to know which slot of a multi-slot is being interacted within the delegates, so I can't use the indices for organization to keep the categories in their proper places and localize accordingly. If this were the case I could make do with 2 multi-slots (large + small) or compromise with just one (all small), but I much prefer the combination of large and small slots visually and conceptually. (2 "core" parts with n "additional modules")

Iridar:
Alright, make sense.

Iridar (approved):

PR #1117, opened 2021-12-31
Add 'OverrideMetaHitEffect' event

### LINKED ISSUE #1116: Add 'OverrideMetaHitEffect' event
Issue #825 adds an event to override Hit Effects, but no such event was added to override Meta Hit Effect. 

Regular Hit Effects include stuff like psionic flashes when projectiles collide with a Templar Parry, while Meta Hit Effect is responsible for playing (or not playing) the effect of blood gushing out of the unit that is being attacked.

Being able to override both regular Hit Effects and Meta Hit Effect is necessary for the use case of making an ability that blocks only a certain amount of damage, but mimics the original Parry otherwise.

Iridar:
Closes #1116

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
@@ -577,7 +577,7 @@ simulated function PlayMetaHitEffect(vector HitLocation, name DamageTypeName, ve
 	local XComPerkContentShared kPerkContent;
 
 	// Start Issue #1116
-	if (TriggerOnOverrideHitEffects(HitLocation, DamageTypeName, Momentum, bIsUnitRuptured, HitResult))
+	if (TriggerOnOverrideMetaHitEffect(HitLocation, DamageTypeName, Momentum, bIsUnitRuptured, HitResult))
 	{
 		return;
 	}
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
@@ -511,20 +511,25 @@ simulated function PlayHitEffects(float Damage, Actor InstigatedBy, vector HitLo
 }
 
 /// HL-Docs: feature:OverrideMetaHitEffect; issue:1116; tags:tactical
-/// Allows listeners to override the default behavior of XComUnitPawn.PlayMetaHitEffect.
+/// Allows listeners to change the behavior of `XComUnitPawn::PlayMetaHitEffect()`.
 /// Meta Hit Effects are intended to communicate the overall effect of the attack,
 /// and include things like blood gushing out of the unit.
-/// If `OverrideMetaHitEffect` is set to `true`, the default behavior is omitted,
-/// and no meta hit effect is played.
+///
+/// Listeners can set `OverrideMetaHitEffect` to `true`, 
+/// and then the default behavior will be omitted entirely,
+/// and no meta hit effect will be played.
+///
+/// Alternatively, listeners can modify the parameters passed with the Tuple 
+/// to modify the default behavior.
 ///
 /// ```event
 /// EventID: OverrideMetaHitEffect,
 /// EventData: [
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
```diff
@@ -510,6 +510,62 @@ simulated function PlayHitEffects(float Damage, Actor InstigatedBy, vector HitLo
...
+/// Allows listeners to override the default behavior of XComUnitPawn.PlayMetaHitEffect.
+/// Meta Hit Effects are intended to communicate the overall effect of the attack,
+/// and include things like blood gushing out of the unit.
+/// If `OverrideMetaHitEffect` is set to `true`, the default behavior is omitted,
+/// and no meta hit effect is played.
+///
+/// ```event
+/// EventID: OverrideMetaHitEffect,
+/// EventData: [
+///     out bool OverrideMetaHitEffect,
```
Xymanek, line 523: Technically this is `inout`

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
```diff
@@ -510,6 +510,62 @@ simulated function PlayHitEffects(float Damage, Actor InstigatedBy, vector HitLo
...
+///     inout vector HitLocation,
+///     inout name DamageTypeName,
+///     inout vector Momentum,
+///     in bool bIsUnitRuptured,
+///     inout enum[EAbilityHitResult] HitResult,
```
Xymanek, line 528: Why are these `inout` with the exception of `bIsUnitRuptured`? Are listeners meant to be able to override these values while keeping the vanilla logic?

Iridar: > Are listeners meant to be able to override these values while keeping the vanilla logic?

Yeah, why not? This would be consistent with changes I did for the other similar event nearby.

> with the exception of bIsUnitRuptured?

It's a bool and I'm lazy about converting it to int and back and explaining it in docs, and I'm not sure it even makes a difference anyway. If there's a use case for it later, it can be added into the Tuple.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
Iridar: Okay, I will adjust.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
Xymanek: > I'm not sure it even makes a difference anyway

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/5bac3f6fa3a159915a138ea0e3027de92f522214/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc#L607-L618

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
```diff
@@ -510,6 +510,62 @@ simulated function PlayHitEffects(float Damage, Actor InstigatedBy, vector HitLo
 	}
 }
 
+/// HL-Docs: feature:OverrideMetaHitEffect; issue:1116; tags:tactical
+/// Allows listeners to override the default behavior of XComUnitPawn.PlayMetaHitEffect.
+/// Meta Hit Effects are intended to communicate the overall effect of the attack,
+/// and include things like blood gushing out of the unit.
+/// If `OverrideMetaHitEffect` is set to `true`, the default behavior is omitted,
+/// and no meta hit effect is played.
```
Xymanek, line 518: Add something like 

> ... Keeping it set to `false` will allow the original logic to run, but you can also modify the parameters passed to it

Iridar:
@Xymanek implemented changes you requested. I'll do the same for OverrideHitEffects.

Xymanek (approved):

PR #1127, opened 2022-01-13
Allow Mods to disable Popups of New Resistance Order Cards being available

### LINKED ISSUE #1126: Allow Mods to disable Popups of New Resistance Order Cards being available
We Already have an option to disable opening the monthly resistance order screen, but when you for example finish a covert action that increases faction influence, you still get a pop-up regarding new resistance cards being available. This is annyoying because I'd rather not get reminded of a garbage mechanic like that existing.
Together with #368 It should be possible to make resistance orders completely disappear from the game.

Grobobobo:
Fixes #1126

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc
@@ -1343,22 +1343,53 @@ function DisplayNewStrategyCardPopups()
 	local StateObjectReference CardRef;
 	local int idx;
 
-	if (NewPlayableCards.Length > 0)
+	//Issue #1126
+	if(ShouldDisplayNewCardPopUps())
 	{
-		// Queue popups for each new card received
-		foreach NewPlayableCards(CardRef, idx)
+		if (NewPlayableCards.Length > 0)
 		{
-			// Only play the "new card" VO on the last card added to the stack, which will be the first card displayed to the player
-			`HQPRES.UIStrategyCardReceived(CardRef, (idx == NewPlayableCards.Length - 1));
-		}
+			// Queue popups for each new card received
+			foreach NewPlayableCards(CardRef, idx)
+			{
+				// Only play the "new card" VO on the last card added to the stack, which will be the first card displayed to the player
+				`HQPRES.UIStrategyCardReceived(CardRef, (idx == NewPlayableCards.Length - 1));
+			}
 
-		// Then clear the new playable card array since the player has now seen the info about them
-		NewGameState = class'XComGameStateContext_ChangeContainer'.static.CreateChangeState("Clear New Faction Cards Array");
-		FactionState = XComGameState_ResistanceFaction(NewGameState.ModifyStateObject(class'XComGameState_ResistanceFaction', ObjectID));
-		FactionState.NewPlayableCards.Length = 0;
-		`XCOMGAME.GameRuleset.SubmitGameState(NewGameState);
+			// Then clear the new playable card array since the player has now seen the info about them
+			NewGameState = class'XComGameStateContext_ChangeContainer'.static.CreateChangeState("Clear New Faction Cards Array");
+			FactionState = XComGameState_ResistanceFaction(NewGameState.ModifyStateObject(class'XComGameState_ResistanceFaction', ObjectID));
+			FactionState.NewPlayableCards.Length = 0;
+			`XCOMGAME.GameRuleset.SubmitGameState(NewGameState);
+		}
 	}
 }
+
+// Start issue #1126
+/// HL-Docs: feature:UIPopups_DisplayNewCardPopups; issue:1126; tags:strategy,ui
+/// Allows overriding whether the new resistance order cards pop ups will be displayed.
+/// Default: Popups for new cards will be shown.
+///
+/// ```event
+/// EventID: UIPopups_DisplayNewCardPopups,
+/// EventData: [inout bool ShouldShow],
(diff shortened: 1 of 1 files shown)


Iridar (changes requested):
If this is meant as a global on/off switch for all new cards, whatever they may be, then an event is overkill. IMO this should be done as a config bool flag in CHHelpers.

Grobobobo:
It's meant as an off switch in my particular case, Someone else may want more conditional changes. Of course I'm more than fine with making it a config variable myself.

Grobobobo:
Looks like it works now

Iridar (approved):
LGTM

Grobobobo:
Ah, wait, now I remember why I made this an event - We have a second wave that enables resistance orders, so the check was supposed to be based on whether that second wave is enabled. That's not possible with a config variable, so I think I should revert to the hook after all. Sorry about that.

Iridar:
> Ah, wait, now I remember why I made this an event - We have a second wave that enables resistance orders, so the check was supposed to be based on whether that second wave is enabled. That's not possible with a config variable, so I think I should revert to the hook after all. Sorry about that.

If you build against Highlander, you can simply set `class'CHHelpers'.default.YourVarName = true;` in OPTC conditionally.

Grobobobo:
Huh, I didn't know you could change config variables in runtime. Yeah, then current state is fine.

Grobobobo:
Although won't you have to keep setting it every time you run the game? I guess it can be set every time on loaded save, and on campaign start.

Iridar:
> Although won't you have to keep setting it every time you run the game? I guess it can be set every time on loaded save, and on campaign start.

Yeah that's what it would take.

Iridar:
Per discussion with Xymanek on discord, it was decided that the event is the better approach. Sorry for the wasted work, it be like that sometimes.

Grobobobo:
I've reverted the changes to the event.

Iridar:
I think this is good to go, but I'll wait for Xym to confirm.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc
```diff
@@ -1343,23 +1343,54 @@ function DisplayNewStrategyCardPopups()
 	local StateObjectReference CardRef;
 	local int idx;
 
-	if (NewPlayableCards.Length > 0)
+	//Issue #1126
+	if(ShouldDisplayNewCardPopUps())
 	{
-		// Queue popups for each new card received
-		foreach NewPlayableCards(CardRef, idx)
+		if (NewPlayableCards.Length > 0)
```
Xymanek, line 1349:
```
if (NewPlayableCards.Length > 0 && /* Issue #1126 */ ShouldDisplayNewCardPopUps())
```

Please do not waste CPU time when there is no need for it

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc
```diff
@@ -1343,23 +1343,54 @@ function DisplayNewStrategyCardPopups()
...
+/// Allows overriding whether the new resistance order cards pop ups will be displayed.
+/// Default: Popups for new cards will be shown.
+///
+/// ```event
+/// EventID: UIPopups_DisplayNewCardPopups,
+/// EventData: [inout bool ShouldShow],
+/// EventSource: XComGameState_ResistanceFaction,
+/// NewGameState: none
+/// ```
+simulated protected function bool ShouldDisplayNewCardPopUps()
```
Xymanek, line 1378:
```suggestion
simulated private function bool ShouldDisplayNewCardPopUps()
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc
```diff
@@ -1343,23 +1343,54 @@ function DisplayNewStrategyCardPopups()
...
+
+   Tuple = new class'XComLWTuple';
+   Tuple.Id = 'UIPopups_DisplayNewCardPopups';
+   Tuple.Data.Add(1);
+   Tuple.Data[0].kind = XComLWTVBool;
+   Tuple.Data[0].b = true;
+
+   `XEVENTMGR.TriggerEvent('UIPopups_DisplayNewCardPopups', Tuple, self, none);
+
+	return Tuple.Data[0].b;
```
Xymanek, line 1390: Please fix indentation

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_ResistanceFaction.uc
```diff
@@ -1343,23 +1343,54 @@ function DisplayNewStrategyCardPopups()
...
+{
+   local XComLWTuple Tuple;
+
+   Tuple = new class'XComLWTuple';
+   Tuple.Id = 'UIPopups_DisplayNewCardPopups';
+   Tuple.Data.Add(1);
+   Tuple.Data[0].kind = XComLWTVBool;
+   Tuple.Data[0].b = true;
+
+   `XEVENTMGR.TriggerEvent('UIPopups_DisplayNewCardPopups', Tuple, self, none);
```
Xymanek, line 1388: While the current event name isn't problematic, something like `ShouldDisplayNewStrategyCardPopups` would make more sense IMO

Xymanek (changes requested):

Grobobobo:
I'm Unlikely to use this after all.

PR #1132, opened 2022-01-23
Fix Homing Mine always applying bonus environmental damage from Shrapnel

### LINKED ISSUE #1131: Claymore / Homing Mine environmental damage is inconsistent
As I was doing some research for #989, I've noticed something weird.

Regular Claymore deals 10 env. damage, as specified in its destructible archetype, and it takes two to detonate a car. This is consistent with Frag Grenades, which also deal 10 env. damage and it also takes two frags to detonate a car.

Shrapnel Claymore deals 20 env. damage, and it takes just one to detonate a car. 

[Some code](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/69d40c887e4b69504c1783c71de62c1f80cbf70b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Ability.uc#L1289-L1292) for getting environmental damage for targeting method preview uses a dev hack - instead of getting environmental damage from the destructible archetype, the code just grabs environmental damage for the Homing Mine, since it's supposedly the same as for Claymores anyway.

However, the configured env. damage for both shrapnel and regular Homing Mines is 10. 

So okay, I thought, apparently env. damage preview for the Shrapnel Claymore is incorrect, even though in-game I can't tell the difference between env. damage previews between regular and shrapnel Claymores. 

However, the weirdest part is that both regular and shrapnel Homing Mines detonate cars in just one go!

I have a hunch that Homing Mines apply their environmental damage twice. 

Pending further investigation.

Iridar:
Closes #1131 

Homing Mine detonation ability has two effects: `X2Effect_HomingMineDamage` to apply unit damage and regular environmental damage, and `X2Effect_ApplyWeaponDamage`, which is supposed to apply only additional environmental damage, and only if the soldier has the `Shrapnel` ability. 

To accomplish this, Firaxis set up the Weapon Damage Effect with a `X2Condition_AbilityProperty` target condition that is supposed to succeed only if the owner unit has the `Shrapnel` ability. 

In practice this solution does not work, because environmental damage is not triggered against any specific target, so there's nothing to check the target condition against. In result, the Weapon Damage Effect is always applied, so Homing Mine detonation ability always applies bonus environmental damage, regardless if the unit has Shrapnel or not.

My proposed solution is neutering the Weapon Damage Effect, and making the Homing Mine Damage effect decide whether it should apply bonus environmental damage or not. 

The Weapon Damage Effect is neutered rather than removed to make sure the ability has the same number of effects with and without Highlander, because mods may be relying on specific effect indexing if they want to patch the Homing Mine Detonation. It would be bad practice, though.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_ReaperAbilitySet.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_ReaperAbilitySet.uc
@@ -1,0 +1,1345 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2Ability_ReaperAbilitySet.uc
+//  AUTHOR:  Joshua Bouscher  --  7/6/2016
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+class X2Ability_ReaperAbilitySet extends X2Ability 
+	native(Core)							//	some native code needs to access config values for Remote Start
+	config(GameData_SoldierSkills);
+
+var config int		InfiltrationHackBonus;
+var config float    ShadowMobilityMod;
+var config int      ShadowCooldown;
+var config int		ShadowCharges;
+var config int		ShadowRisingCharges;
+var config int      ClaymoreCharges;
+var config int		ClaymoreRange;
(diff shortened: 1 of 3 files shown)


Iridar:
Alternative solution would be to not touch the `X2Effect_HomingMineDamage`, and instead replace the Weapon Damage Effect with a subclass that has the condition built-in.

PR #1135, opened 2022-01-27
Make `X2SoldierClassTemplate::IsWeaponAllowedByClass()` more flexible

### LINKED ISSUE #1057: `X2SoldierClassTemplate::IsWeaponAllowedByClass()` is inflexible
`UIArmory_Loadout::GetDisabledReason()` runs `X2SoldierClassTemplate::IsWeaponAllowedByClass()` to check if the weapon is allowed by unit's soldier class, which is a very simple check:
```
function bool IsWeaponAllowedByClass(X2WeaponTemplate WeaponTemplate)
{
	local int i;

	switch(WeaponTemplate.InventorySlot)
	{
	case eInvSlot_PrimaryWeapon: break;
	case eInvSlot_SecondaryWeapon: break;
	default:
		return true;
	}

	for (i = 0; i < AllowedWeapons.Length; ++i)
	{
		if (WeaponTemplate.InventorySlot == AllowedWeapons[i].SlotType &&
			WeaponTemplate.WeaponCat == AllowedWeapons[i].WeaponType)
			return true;
	}
	return false;
}
```
As you can see, for primary and secondary weapons, the function checks the category and slot specified in the weapon template. However, many mods like True Primary Secondaries or Dedicated Pistol Slot allow equipping secondary weapons into slots other than specified in their weapon templates. 

However, `GetDisabledReason()` will forbid equipping the item through armory loadout UI for certain soldier classes under the premise that their soldier class template does not allow equipping the weapon in question into primary/secondary inventory slot, even though the inventory slot might have absolutely nothing to do with primary or secondary weapon slot.

This creates the requirement of using `CanAddItemToInventory_CH_Improved`, potentially causing the duplicate items bug described in #1056 if handled improperly, not to mention just having to use a function to bypass unnecessarily inflexible piece of code.

I think the optimal solution to this problem is to  create a wrapper for `IsWeaponAllowedByClass()` that would also pass the desired inventory slot as an argument, allowing the logic to be more accurate, and apply primary/secondary weapon restrictions only when the weapon is actually being equipped into these slots. Other places in code calling `IsWeaponAllowedByClass()` would have to be adjusted to use the wrapper when relevant, which is around ~7 places.

Iridar:
Closes #1057 

This replaces all instances of `IsWeaponAllowedByClass()` with a new wrapper in existing CHL and base game code.

I've tested the PR and it works as intended, e.g. if a `Pistol` weapon category is added to soldier class' AllowedWeapons, then the event listener for `OverrideShowItemInLockerList` is all that's required to equip it in-game.

```
static function EventListenerReturn OnOverrideShowItemInLockerList(Object EventData, Object EventSource, XComGameState GameState, Name EventID, Object CallbackObject)
{
    local XComGameState_Item ItemState;
    local XComLWTuple Tuple;
    local bool bSlotShowItemInLockerList;
    local EInventorySlot Slot;
    local XComGameState_Unit UnitState;

    ItemState = XComGameState_Item(EventSource);
    Tuple = XComLWTuple(EventData);

    bSlotShowItemInLockerList = Tuple.Data[0].b;
    Slot = EInventorySlot(Tuple.Data[1].i);
    UnitState = XComGameState_Unit(Tuple.Data[2].o);

    // Your code here
    if (Slot == eInvSlot_PrimaryWeapon && ItemState.GetWeaponCategory() == 'Pistol'
      //&& UnitState.GetSoldierClassTemplate().IsWeaponAllowedByClass_CH(X2WeaponTemplate(ItemState.GetMyTemplate()), eInvSlot_PrimaryWeapon)
        )
	{
		bSlotShowItemInLockerList = true;
	}

    Tuple.Data[0].b = bSlotShowItemInLockerList;

    return ELR_NoInterrupt;
}
```
The `IsWeaponAllowedByClass_CH()` in this listener itself is optional; if omitted then pistols will show up for all soldier classes, but the in-game UI will forbid equipping them, mentioning that they are "sharpshooter only", or whichever soldier class is found by `GetAllowedClassForWeapon_CH()` first.

Note: the `HeavyWeaponTemplate.InventorySlot == eInvSlot_HeavyWeapon` check in  `XCGS_Unit::GetBestHeavyWeaponTemplates()` is not useless, even though the original `X2SoldierClassTemplate::IsWeaponAllowedByClass()` allows any weapons that are not primary or secondary weapons, nothing is stopping mods from creating their own soldier classes that would restrict the use of heavy weapons.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIArmory_Loadout.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIArmory_Loadout.uc
@@ -753,9 +753,10 @@ simulated function string GetDisabledReason(XComGameState_Item Item, EInventoryS
 	if(WeaponTemplate != none)
 	{
 		SoldierClassTemplate = UpdatedUnit.GetSoldierClassTemplate();
-		if(SoldierClassTemplate != none && !SoldierClassTemplate.IsWeaponAllowedByClass(WeaponTemplate))
+		if(SoldierClassTemplate != none && !SoldierClassTemplate.IsWeaponAllowedByClass_CH(WeaponTemplate, SelectedSlot)) // Issue #1057 - call IsWeaponAllowedByClass_CH() instead of the original.
 		{
-			AllowedSoldierClassTemplate = class'UIUtilities_Strategy'.static.GetAllowedClassForWeapon(WeaponTemplate);
+			// Issue #1057 - call GetAllowedClassForWeapon_CH() instead of the original.
+			AllowedSoldierClassTemplate = class'UIUtilities_Strategy'.static.GetAllowedClassForWeapon_CH(WeaponTemplate, SelectedSlot);
 			if(AllowedSoldierClassTemplate == none)
 			{
 				DisabledReason = m_strMissingAllowedClass;
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalQuickLaunch_UnitSlot.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalQuickLaunch_UnitSlot.uc
@@ -1,0 +1,1009 @@
+//---------------------------------------------------------------------------------------
+//  *********   FIRAXIS SOURCE CODE   ******************
+//  FILE:    UITacticalQuickLaunch_UnitSlot
+//  AUTHOR:  Sam Batista --  02/26/14
+//  PURPOSE: Contains all the options and logic to outfit a unit (DEBUG ONLY)
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//--------------------------------------------------------------------------------------- 
+
+// TODO: Add Utility, Mission, then Backpack dropdowns / selectors
+
+class UITacticalQuickLaunch_UnitSlot extends UIPanel;
+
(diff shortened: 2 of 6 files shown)


Xymanek (changes requested):
Please add `HL-Docs: ref:` to the changed calls

Xymanek:
Better compare: https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/compare/05184185e4b3689f34d6dbbc3812178d662b477f...5d6bfcef6cbf08219bb2a37b59fd8f16190d4cad

Xymanek:
Something went quite wrong with the generated docs: 

image

@robojumper any ideas?

robojumper:
> Something went quite wrong with the generated docs:

what would you expect given the documentation comments? the body of `ref` items is appended to the body of the main `feature` item

Xymanek:
> the body of `ref` items is appended to the body of the main `feature` item

I see (wasn't aware of that behaviour). @Iridar mind adjusting so that the generated docs page is sanely readable?

Iridar:
> > the body of `ref` items is appended to the body of the main `feature` item
> 
> I see (wasn't aware of that behaviour). @Iridar mind adjusting so that the generated docs page is sanely readable?

Sure, I'll try.

Iridar:
Do you mind giving me a link where I can preview the docs?

Iridar:
Adjusted based on feedback and discord conversation, rebased to have just two commits.

Issue #1134, opened 2022-01-27
Request:  Add event for overriding Resistance Faction Icon in XComGameState_Unit


FearTheBunnies:
I would like a new method added to `XComGameState_Unit` that would trigger an event for overriding the Resistance Faction Icon so that custom hero classes can show custom icons in the promotion screen, soldier list, and on the tactical map.

```
function StackedUIIconData GetResistanceFactionIcon()
{
    local XComLWTuple Tuple;
    local XComGameState_ResistanceFaction FactionState;
    local StackedUIIconData CustomIcon;
    local int idx;

    FactionState = GetResistanceFaction();
    if(FactionState != none)
    {
        CustomIcon = FactionState.GetFactionIcon();
    }

    Tuple = new class'XComLWTuple';
    Tuple.Id = 'OverrideResistanceFactionIcon';
    Tuple.Data.Add(2);

    Tuple.Data[0].kind = XComLWTVArrayStrings;
    Tuple.Data[0].as = CustomIcon.Images;
    Tuple.Data[1].kind = XComLWTVBool;
    Tuple.Data[1].b = CustomIcon.bInvert;

    `XEVENTMGR.TriggerEvent('OverrideResistanceFactionIcon', Tuple, self, none);

    CustomIcon.Images = Tuple[0].as;
    CustomIcon.bInvert = Tuple[1].b;

    for(idx = 0; idx < CustomIcon.Images.Length; i++)
    {
        CustomIcon.Images[idx] = Repl(CustomIcon.Images[idx], "img:///", "");
    }

    return CustomIcon;
}
```

I added a guard to strip prefixed `img:///` tags from image links for end-user niceness.  Not required to get the achieved effect.

PR #1136, opened 2022-01-29
Add event for overriding resistance faction icon #1134

FearTheBunnies:
This is for issue #1134 which will allow mods to override the resistance faction icon when using custom hero classes.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAfterAction_ListItem.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAfterAction_ListItem.uc
@@ -63,13 +63,14 @@ simulated function UpdateData(optional StateObjectReference UnitRef)
 	local XComGameState_Unit Unit, Bondmate;
 	local SoldierBond BondData;
 	local StateObjectReference BondmateRef;
-	local XComGameState_ResistanceFaction FactionState;
+	//local XComGameState_ResistanceFaction FactionState; //Issue #1134, not needed
+	local StackedUIIconData StackedClassIcon; // Variable for issue #1134
 
 	UnitReference = UnitRef;
 
 	Unit = XComGameState_Unit(`XCOMHISTORY.GetGameStateForObjectID(UnitRef.ObjectID));
-	
-	FactionState = Unit.GetResistanceFaction();
+
+	//FactionState = Unit.GetResistanceFaction(); //Issue #1134, not needed
 
 	if(Unit.bCaptured)
 	{
@@ -154,7 +155,11 @@ simulated function UpdateData(optional StateObjectReference UnitRef)
 		AS_SetUnitWill(-1, "");
 	}
 
-	AS_SetFactionIcon(FactionState.GetFactionIcon());
+	// Start Issue #1134
+	StackedClassIcon = Unit.GetStackedClassIcon();
+	if (StackedClassIcon.Images.Length > 0)
+		AS_SetFactionIcon(StackedClassIcon);
+	// End Issue #1134
 
 	EnableNavigation(); // bsg-nlong (1.24.17): We will always want navigation now that the bond icon is a thing
 	if( bCanPromote )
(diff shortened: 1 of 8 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -15465,6 +15465,65 @@ private function XComLWTuple TriggerSoldierRankEvent(const int Rank, const name
...
+
+    Tuple.Data[0].kind = XComLWTVArrayStrings;
+    Tuple.Data[0].as = CustomIcon.Images;
+    Tuple.Data[1].kind = XComLWTVBool;
+    Tuple.Data[1].b = CustomIcon.bInvert;
+
+    `XEVENTMGR.TriggerEvent('OverrideResistanceFactionIcon', Tuple, self, none);
+
+	CustomIcon.bInvert = Tuple.Data[1].b;
+	CustomIcon.Images.Length = 0;
```
Xymanek, line 15517: Please fix the indentation

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -15465,6 +15465,65 @@ private function XComLWTuple TriggerSoldierRankEvent(const int Rank, const name
...
+// Start Issue #1134
+//
+// Function that returns the current unit's faction icon. It sends a
+// `OverrideResistanceFactionIcon` event with a tuple that allows listeners to override the
+// default structure of the faction icon. The tuple data takes the form:
+//
+//   {ID: "OverrideResistanceFactionIcon", Data: [inout array<string> Images, inout bool bInvertImage]}
+//
+// Due to the irregularities of how StackedUIIconData is consumed, strings applied to
+// the Images array should not start with 'img:///'.
```
Xymanek, line 15477: There is no need for these 2 comment "parts". Please put everything under/into `HL-Docs`

Xymanek (changes requested):
In many places where you changed `FactionState.GetFactionIcon()` to `UnitState.GetResistanceFactionIcon()`, said code is wrapped into `if (FactionState != none)` - rending the feature useless for non-faction soldiers.

Instead, you should store the result of `UnitState.GetResistanceFactionIcon()` and change the checks to `if (FactionIcon.Images.Length > 0)`

FearTheBunnies:
> In many places where you changed `FactionState.GetFactionIcon()` to `UnitState.GetResistanceFactionIcon()`, said code is wrapped into `if (FactionState != none)` - rending the feature useless for non-faction soldiers.
> 
> Instead, you should store the result of `UnitState.GetResistanceFactionIcon()` and change the checks to `if (FactionIcon.Images.Length > 0)`

Not opposed to changing the check, but was curious your thoughts on it.  Originally that icon was only related to faction heroes, so this change would be expanding the functionality and I am wanting to make sure we're good with that.

Xymanek:
While I understand the history behind the current setup, I think it's a disservice to the overall flexibility of the system - IMO they should've just converted all icons to stacked (albeit with a single item for traditional class icons). As such, I think it's a good idea to allow the stacked icons to be usable for normal soldiers

FearTheBunnies:
Updated the logic to make the event more generic so that non-hero soldiers can get the same benefit for having a stacked class icon instead of a flat class icon if a class author so wishes it.  Also cleaned up the other couple of small asks.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAfterAction_ListItem.uc
```diff
@@ -63,13 +63,11 @@ simulated function UpdateData(optional StateObjectReference UnitRef)
 	local XComGameState_Unit Unit, Bondmate;
 	local SoldierBond BondData;
 	local StateObjectReference BondmateRef;
-	local XComGameState_ResistanceFaction FactionState;
+	local StackedUIIconData StackedClassIcon; // Variable for issue #1134
```
Xymanek, line 66: Please do not remove lines from base game code - simply comment them out (same goes for other instances)

Xymanek (changes requested):

FearTheBunnies:
Flattened the commit and fixed documentation.

Xymanek (approved):

PR #1139, opened 2022-02-27
Right click melee priority #1138

### LINKED ISSUE #1138: Request: Right Click Melee should smartly prioritize which melee skill to choose
In the class `X2AbilityTrigger_EndOfMove` there is a static method which is used by `XComTacticalInput` when the user clicks the right mouse button that will select the melee ability to trigger for the unit.  Current logic is a greedy algorithm that will just take the first melee skill it finds.  Normally, that is fine when a unit has 1 melee skill, but it gets awkward for units with more than 1.  For example, a Templar with Rend and Subdue (from Nonlethal Combat mod) will prioritize Subdue over Rend when hitting the right click due to the order in which abilities are added to the available ability list.

To solve for this, I would like to expand the selection logic to allow for the selection to be based upon Shot HUD Priority, where it scans left -> right to find the first melee skill worth using.  There is a potential for ties to still occur, but it will stop the issue where Subdue is "overriding" Rend/Sword Slash on the right click UI.  I will also give it the ability for a custom priority to be assigned for specific use-case scenarios (e.g.: Using the Combat Knife's Knife Fighter if it is available instead of Sword Slash, prioritizing Kinetic Strike instead of Shield Bash, etc.) depending on what mod authors are wanting for their skills.

FearTheBunnies:
This is a pull request to address right-click prioritization when choosing melee skills.

Resolves #1138

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
@@ -1,0 +1,51 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2AbilityTrigger_EndOfMove.uc
+//  AUTHOR:  DavidBurchanowski  --  9/3/2015
+//  PURPOSE: Sentinal object to indicate that a given ability can be activated from the pathing pawn melee UI
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+
+class X2AbilityTrigger_EndOfMove extends X2AbilityTrigger;
+
+static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
+{
+	local X2AbilityTrigger Trigger;
+
+	if (Template == none)
+		return false;
+
+	foreach Template.AbilityTriggers(Trigger)
+	{
+		if(X2AbilityTrigger_EndOfMove(Trigger) != none)
+		{
+			return true;
+		}
+	}
+
+	return false;
+}
+
+static function XComGameState_Ability GetAvailableEndOfMoveAbilityForUnit(XComGameState_Unit UnitState)
+{
+	local XComGameStateHistory History;
+	local GameRulesCache_Unit UnitCache;
+	local XComGameState_Ability AbilityState;
+	local int ActionIndex;
+
+	if(`TACTICALRULES.GetGameRulesCache_Unit(UnitState.GetReference(), UnitCache))
+	{
+		History = `XCOMHISTORY;
+		for(ActionIndex = 0; ActionIndex < UnitCache.AvailableActions.Length; ActionIndex++)
+		{
+			AbilityState = XComGameState_Ability(History.GetGameStateForObjectID(UnitCache.AvailableActions[ActionIndex].AbilityObjectRef.ObjectID));
+			if(AbilityState != none && AbilityHasEndOfMoveTrigger(AbilityState.GetMyTemplate()))
+			{
+				return AbilityState;
+			}
+		}
+	}
+
+	return None;
+}
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
@@ -9,6 +9,20 @@
 
 class X2AbilityTrigger_EndOfMove extends X2AbilityTrigger;
 
+// Properties for Issue #1138.
+// Adds priority weighting identical to Shot HUD Priority so that the left-most HUD melee
+// ability can be picked up and prioritized for use. 
+
+// 0 is the highest priority as that corresponds to the left-most slot in the Shot HUD.
+
+// Each priority bucket is base 100_000 so that when Shot HUD priority is included it should not push into the next bucket
+const HIGH_PRIORITY_MELEE		= 100000;
+const STANDARD_PRIORITY_MELEE	= 200000;
+const LOW_PRIORITY_MELEE		= 300000;
+
+// Custom priorities can be set manually using RightClickMeleePriority;
+var int RightClickMeleePriority;
+
 static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
 {
 	local X2AbilityTrigger Trigger;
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -9,6 +9,12 @@
 
 class X2AbilityTrigger_EndOfMove extends X2AbilityTrigger;
 
+// RIGHT CLICK MELEE PRIORITIES low -> high == highest priority -> lowest priority
+// SECONDARY PRIORITY: Shot HUD Priority low -> high == left -> right in the UI bar
```
Xymanek, line 13: This makes no sense at all to me

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -9,6 +9,12 @@
...
+const SHOT_HUD_PRIORITY	= 1000;
+
+var int RClickMeleePriority;
```
Xymanek, line 16:
```suggestion
var int RightClickMeleePriority;
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -27,25 +33,99 @@ static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
 	return false;
 }
 
+// Start Issue #1138
+/// HL-Docs: feature:GetAvailableEndOfMoveAbilityForUnit; issue:1138; tags:ui,tactical
+/// By default the game was using a greedy algorithm for selecting which melee
+/// skill to use when the user right-clicks on the mouse to do a dash-melee attack
+/// on an enemy unit.  This upgrades the logic for that selection process to now
```
Xymanek, line 40:
```suggestion
/// on an enemy unit. This upgrades the logic for that selection process to now
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -27,25 +33,99 @@ static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
...
+/// prioritize all melee skills based upon the Ability Template's Shot HUD Priority.
+///
+/// Custom prioritization can also be granted by modifying `RClickMeleePriority` when
+/// using `Template.AbilityTriggers.AddItem(new class'X2AbilityTrigger_EndOfMove');`
+/// to enable an ability to be a dashing melee ability.
+///
+/// ```
+/// local X2AbilityTrigger_EndOfMove EndOfMoveTrigger;
+/// EndOfMoveTrigger = new class'X2AbilityTrigger_EndOfMove';
```
Xymanek, line 49:
```suggestion
/// local X2AbilityTrigger_EndOfMove EndOfMoveTrigger;
///
/// EndOfMoveTrigger = new class'X2AbilityTrigger_EndOfMove';
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -27,25 +33,99 @@ static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
...
+/// By default the game was using a greedy algorithm for selecting which melee
+/// skill to use when the user right-clicks on the mouse to do a dash-melee attack
+/// on an enemy unit.  This upgrades the logic for that selection process to now
+/// prioritize all melee skills based upon the Ability Template's Shot HUD Priority.
+///
+/// Custom prioritization can also be granted by modifying `RClickMeleePriority` when
+/// using `Template.AbilityTriggers.AddItem(new class'X2AbilityTrigger_EndOfMove');`
+/// to enable an ability to be a dashing melee ability.
+///
+/// ```
```
Xymanek, line 47:
```suggestion
/// ```unrealscript
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -27,25 +33,99 @@ static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
...
 		}
 	}
 
-	return None;
-}
\ No newline at end of file
+	return PriorityAbility;
+}
+
+static function X2AbilityTrigger_EndOfMove GetEndOfMoveTrigger(X2AbilityTemplate Template)
```
Xymanek, line 109:
```suggestion
static private function X2AbilityTrigger_EndOfMove GetEndOfMoveTrigger(X2AbilityTemplate Template)
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -27,25 +33,99 @@ static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
...
+	local X2AbilityTrigger Trigger;
+
+	if (Template == none)
+		return none;
+
+	foreach Template.AbilityTriggers(Trigger)
+	{
+		if(X2AbilityTrigger_EndOfMove(Trigger) != none)
+		{
+			return X2AbilityTrigger_EndOfMove(Trigger);
```
Xymanek, line 120: Please avoid doing the same cast twice

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -27,25 +33,99 @@ static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
...
 	local GameRulesCache_Unit UnitCache;
-	local XComGameState_Ability AbilityState;
-	local int ActionIndex;
+	local AvailableAction Action;
+	local XComGameState_Ability AbilityState, PriorityAbility;
+	local X2AbilityTrigger_EndOfMove EndOfMoveTrigger;
+	local int CurrentPriority, AbilityPriority;
+
+	PriorityAbility = none;
+	CurrentPriority = 2147483647; // Max INT
```
Xymanek, line 63: Both of these are useless:

> ```uc
> PriorityAbility = none;
> ```

That's the default value of object variables anyway

> ```uc
> CurrentPriority = 2147483647; // Max INT
> ```

Never relevant due to `if(PriorityAbility == none`

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -27,25 +33,99 @@ static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
...
 static function XComGameState_Ability GetAvailableEndOfMoveAbilityForUnit(XComGameState_Unit UnitState)
 {
 	local XComGameStateHistory History;
 	local GameRulesCache_Unit UnitCache;
-	local XComGameState_Ability AbilityState;
-	local int ActionIndex;
+	local AvailableAction Action;
+	local XComGameState_Ability AbilityState, PriorityAbility;
+	local X2AbilityTrigger_EndOfMove EndOfMoveTrigger;
+	local int CurrentPriority, AbilityPriority;
```
Xymanek, line 60: Please comment out base game variables (which are no longer used) and add a separate `// Variables for Issue #1138` block

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -27,25 +33,99 @@ static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
...
+			{
+				continue;
+			}
+
+			// Calculate priority
+			// If default value then use SHOT_HUD_PRIORITY, otherwise grab its priority
+			AbilityPriority = EndOfMoveTrigger.RClickMeleePriority >= 0 ? EndOfMoveTrigger.RClickMeleePriority : const.SHOT_HUD_PRIORITY;
+			
+			// Apply the shot HUD's priority weight to give a secondary sort for prioritization
+			AbilityPriority += Action.ShotHUDPriority;
```
Xymanek, line 95: This is not a secondary sort, this is a modifier to primary sort

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -27,25 +33,99 @@ static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
...
+			EndOfMoveTrigger = GetEndOfMoveTrigger(AbilityState.GetMyTemplate());
+			// Not right-click enabled, so move on
+			if(EndOfMoveTrigger == none)
+			{
+				continue;
+			}
+
+			// Calculate priority
+			// If default value then use SHOT_HUD_PRIORITY, otherwise grab its priority
+			AbilityPriority = EndOfMoveTrigger.RClickMeleePriority >= 0 ? EndOfMoveTrigger.RClickMeleePriority : const.SHOT_HUD_PRIORITY;
```
Xymanek, line 92: Several questions:

1) Why `SHOT_HUD_PRIORITY` was chosen as the name for default right-click-melee priority? Seems confusing/misleading to me
2) Why are negative custom explicit priorities not supported? e.g. in Nonlethal Combat case, wouldn't it make sense to set Subdue to something negative (very low) so that any other melee ability trumps it?

Xymanek (changes requested):
Several overall questions:

1) What happens if same ability has multiple `X2AbilityTrigger_EndOfMove` triggers?
2) What happens if the different dashing melee abilities have different target conditions? E.g. pretty sure I saw dashing medkit somewhere - but the current code doesn't support checking the (potential) target

FearTheBunnies:
> Several overall questions:
> 
> 1. What happens if same ability has multiple `X2AbilityTrigger_EndOfMove` triggers?
> 2. What happens if the different dashing melee abilities have different target conditions? E.g. pretty sure I saw dashing medkit somewhere - but the current code doesn't support checking the (potential) target

1) I am going to address this directly, and loop through to find the highest priority option available.  It will make patching less intrusive for mod makers (not accounting for conflicting patching).  In theory there should only ever be 1 on a skill, but if a mod adds a second one because they want a higher priority for that specific skill they can.
2) That scenario is handled by 
```
	// Is the action even available?
	if(Action.AvailableCode != 'AA_Success')
	{
		continue;
	}
```
That will filter down to the correct movement skill based upon targets, etc.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
FearTheBunnies: Negative values are not supported because it is mirroring the ShotHUDPriority scale, meaning that 0 will be the highest priority due to wanting the "left-most" skill in the Shot HUD.  I have tweaked the comments and naming conventions to make this clearer.

Xymanek:
Better compare: https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/compare/0cb9ce172246e7197cbb6c2d2dcdbdb6d8892d8a...aa8d059bee77d4b8d8565e49ec6f99004ebe7635

FearTheBunnies:
Flattened the commits to have a distinct add commit and a distinct edit commit

Iridar:
Per Discord discussion, I propose using a DLC hook for this, with the following algorithm:

1. `GetAvailableEndOfMoveAbilityForUnit()` gathers Ability States of abilities eligible for end-of-move activation (by checking trigger on the template)
2. (optional / pending discussion) Sorts the array of Ability States by Shot HUD Priority.
3. Runs the DLC hook, passing it the Unit State and `out array` of Ability States.
4. DLC hooks in mods, if any, apply their own arbitrary sorting logic to the array.
5. `GetAvailableEndOfMoveAbilityForUnit()` chooses whichever Ability State is in the 0th index of the array.

Would also need to handle the case if a mod decides to wipe the array entirely. I see two options:
1. Store the 0th Ability State before passing the array to the hook, and then use that Ability State if the array returns empty, so we can activate _something_.
2. Forbid using End of Move melee at all in those conditions, assuming the array was wiped with a purpose.

FearTheBunnies:
Closing this pull request because better approach is in a different branch.

PR #1141, opened 2022-04-03
Issue #1140 - fix bugs introduced by Issue #783

### LINKED ISSUE #1140: Faction Heroes are created without an origin country
# The Problem

Without Highlander, Reapers and Templars roll a random country for the purposes of name and biography, just like for regular soldiers, and then their country gets overridden to their faction countries for the sake of country flag.

With Highlander, Reapers and Templars are generated with their faction countries right away, so their names default to USA names, and their biography is missing a country name entirely, presumably because their faction countries do not have a localized name.

# The Suspect

#783 added faction hero XGCharacterGenerator classes into Highlander, and adjusted their logic slightly so that the new DLC hook, `ModifyGeneratedUnitAppearance`, can work more consistently. 

For example, for Reapers, the following code:

```
function TSoldier CreateTSoldier(optional name CharacterTemplateName, optional EGender eForceGender, optional name nmCountry = '', optional int iRace = -1, optional name ArmorName)
{
	local X2SoldierClassTemplateManager ClassMgr;
	local X2SoldierClassTemplate ClassTemplate;

	kSoldier = super.CreateTSoldier('ReaperSoldier', eForceGender, nmCountry, iRace, ArmorName);
	SetCountry('Country_Reaper');
	ClassMgr = class'X2SoldierClassTemplateManager'.static.GetSoldierClassTemplateManager();
	ClassTemplate = ClassMgr.FindSoldierClassTemplate('Reaper');
	kSoldier.strNickName = GenerateNickname(ClassTemplate, kSoldier.kAppearance.iGender);

	return kSoldier;
}
```
was replaced with
```
function TSoldier CreateTSoldier(optional name CharacterTemplateName, optional EGender eForceGender, optional name nmCountry = '', optional int iRace = -1, optional name ArmorName)
{
	kSoldier = super.CreateTSoldier('ReaperSoldier', eForceGender, nmCountry, iRace, ArmorName);
	return kSoldier;
}
```

The overall idea of the refactor was to move calling `super.CreateTSoldier()` to the very end of the function, so that `ModifyGeneratedUnitAppearance`, which is called at the very end of `XGCharacterGenerator::CreateTSoldier()`, has the final say on the properties of the generated soldier.

This had the assumption that `super.CreateTSoldier()` would call `SetCountry()`, which in the case of Reaper character generator was changed to:
```
function SetCountry(name nmCountry)
{
	kSoldier.nmCountry = 'Country_Reaper';
	kSoldier.kAppearance.nmFlag = kSoldier.nmCountry; // needs to be copied here for pawns -- jboswell
}
```
This assumption has missed the fact that `XGCharacterGenerator::CreateTSoldier()` sets `BioCountryName = kSoldier.nmCountry;`, which at that time and place would be the randomly generated country. 

The original `XGCharacterGenerator_Reaper::CreateTSoldier()` would override `kSoldier.nmCountry` later, but it would not affect the `BioCountryName`.

With the Highlander changes, the `BioCountryName` is set to faction country, creating the problem.

Iridar:
Fixes #1140

A slight refactor to the code of XGCharacterGenerator to address the issue introduced by #783.

An explanation on what is being changed and why.

# No Highlander

When the game generates a faction hero, the code is being called in this order:

`XGCharacterGenerator_Reaper::CreateTSoldier()` calls `XGCharacterGenerator::CreateTSoldier()`, which does the following:

1. Generate a random country and set it to the TSoldier using `SetCountry()` function.
2. Generate other soldier's properties based on that country, like Race, Voice, Name and BioCountryName, which will be stored to be used later when generating unit's bio.
3. The created TSoldier struct is then returned to the `XGCharacterGenerator_Reaper::CreateTSoldier()`, which calls `SetCountry()` again, this time with the faction country, replacing the values in `kSoldier.nmCountry` and `kSoldier.kAppearance.nmFlag` with the faction country, (but not BioCountryName, which still stores the previously randomly generated country), and generates a new nickname, appropriate for the faction.

The result is a soldier whose first and last name, as well as their BioCountry are randomly generated, but the flag country and nickname are appropriate for their faction country.

# With Highlander

Highlander overrides the `SetCountry()` function in faction hero character generators, so when the `XGCharacterGenerator::CreateTSoldier()` calls it with the randomly generated country, the `SetCountry()` sets a faction country anyway, which is then used to generate Race, Voice, Name and is set to BioCountryName. 

This creates a lot of issues; the generated name always defaults to USA names, and the country name is entirely absent from the generated unit's bio, presumably because faction countries are missing localized name.

# Proposed Fix

Refactor the `XGCharacterGenerator::CreateTSoldier()` so it uses the randomly generated country for the purposes of generating Race, Voice, Name and BioCountryName. It will still call the overridden `XGCharacterGenerator_Reaper::SetCountry()`, which will assign the faction country to `kSoldier.nmCountry` and `kSoldier.kAppearance.nmFlag`, achieving the desired result. 

The only hiccup in this plan is that without the Highlander, the randomly generated country is used for the `SetRace()` function, while with the proposed fix it will be called with the faction country.

So the bandaid solution is to add a `SetRace_CH()` function which takes the randomly generated country as an argument.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator.uc
@@ -399,9 +399,17 @@ function TSoldier CreateTSoldier( optional name CharacterTemplateName, optional 
 
 	SetVoice(CharacterTemplateName, nmCountry);
 	SetAttitude();
-	GenerateName( kSoldier.kAppearance.iGender, kSoldier.nmCountry, kSoldier.strFirstName, kSoldier.strLastName, kSoldier.kAppearance.iRace );
 
-	BioCountryName = kSoldier.nmCountry;
+	// Start Issue #1140
+	// Pass nmCountry to GenerateName and BioCountryName instead of kSoldier.nmCountry.
+	// This is done to fix the bug caused by #783, which refactored XGCharacterGenerator classes for faction heroes,
+	// which resulted in using their faction countries to generate character name and biography as opposed to using
+	// a random country, like they did before #783 was implemented.
+	GenerateName( kSoldier.kAppearance.iGender, nmCountry, kSoldier.strFirstName, kSoldier.strLastName, kSoldier.kAppearance.iRace );
+	//GenerateName( kSoldier.kAppearance.iGender, kSoldier.nmCountry, kSoldier.strFirstName, kSoldier.strLastName, kSoldier.kAppearance.iRace );
+	BioCountryName = nmCountry;
+	//BioCountryName = kSoldier.nmCountry;
+	// End Issue #1140
 
 	// Start issue #783
 	ModifyGeneratedUnitAppearance(CharacterTemplateName, eForceGender, nmCountry, iRace, ArmorName);
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGCharacterGenerator.uc
@@ -373,7 +373,8 @@ function TSoldier CreateTSoldier( optional name CharacterTemplateName, optional 
 		nmCountry = PickOriginCountry();
 	
 	SetCountry(nmCountry);
-	SetRace(iRace);
+	//SetRace(iRace);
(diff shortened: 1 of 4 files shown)


Iridar:
After I merged this, it occurred to me I might have forgotten to actually test it, so I tested it, and it's a good thing, cuz I'm pretty sure "Takumi Yamamoto" is not a valid Skirmisher name.

PR #1149, opened 2022-06-26
Create an option to make Fatigue and Injury Times Independent

### LINKED ISSUE #1148: Create an option to make Fatigue and Injury Times Independent
Currently, Recovering from injuries clears the soldier's shaken/tired status. This is obviously intended behavior, but it's a problem if the game is configured for low injury times and high fatigue times, as getting injured will for the most part make fatigue irrelevant. A simple bool config variable, or an event would help with that.

Grobobobo:
Fixes #1148

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameStateContext_HeadquartersOrder.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameStateContext_HeadquartersOrder.uc
@@ -1228,7 +1228,7 @@ static function IssueHeadquartersOrder(const out HeadquartersOrderInputContext U
 
 
 // Start issue #1148
-/// HL-Docs: feature:UIPopups_DisplayNewCardPopups; issue:1126; tags:strategy
+/// HL-Docs: feature:OverrideInjuryClearingFatigueBehavior; issue:1126; tags:strategy
 /// Allows overriding The game's behavior regarding Removing Fatigue on Healing
 /// Default: Vanilla behavior will be used.
 ///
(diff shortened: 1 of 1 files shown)


Iridar:
Looks good to me. Is this tested?

Xymanek:
How does the unit status UI behave?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameStateContext_HeadquartersOrder.uc
```diff
@@ -1224,4 +1224,32 @@ static function IssueHeadquartersOrder(const out HeadquartersOrderInputContext U
...
+
+// Start issue #1148
+/// HL-Docs: feature:UIPopups_DisplayNewCardPopups; issue:1126; tags:strategy
+/// Allows overriding The game's behavior regarding Removing Fatigue on Healing
+/// Default: Vanilla behavior will be used.
+///
+/// ```event
+/// EventID: OverrideInjuryClearingFatigueBehavior,
+/// EventData: [inout bool ShouldOverride],
+/// EventSource: XComGameState_Unit,
```
robojumper, line 1238:
```suggestion
/// EventSource: XComGameState_Unit (UnitState),
```

Grobobobo:
I've tested it, if you just put it to false, then the unit switches to the shaken status after recovering from injury.

Xymanek:
> the unit switches to the shaken status after recovering from injury

The problem with this is that the player has no idea when the soldier will actually be available for duty

Grobobobo:
> > the unit switches to the shaken status after recovering from injury
> 
> The problem with this is that the player has no idea when the soldier will actually be available for duty
I don't think it's that big of a problem, you can still clearly see the red will regardless, but even then
I imagine it'll be possible to modify the injury status message to also display shaken/tired time

Iridar:
A systemic change like that would've been probably better implemented as a config bool, I struggle to come up with a use case where you would want different behavior for different soldiers in this regard. 

> The problem with this is that the player has no idea when the soldier will actually be available for duty

I agree that this is a problem as well. 

However, I don't see either as grounds enough to deny this change, seeing as it preserves the vanilla behavior either way.

PR #1152, opened 2022-07-29
Rclick melee delegate 1138

### LINKED ISSUE #1138: Request: Right Click Melee should smartly prioritize which melee skill to choose

FearTheBunnies:
Uses a delegate based approach to expanding the functionality versus actually forcing a revised sorting algorithm.  Will allow other mods to hook into the right-click  melee determination logic and provide prioritization dynamically for which melee ability should be used. Closes #1138

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
@@ -287,9 +287,24 @@ struct OverrideHasHeightAdvantageStruct
 var protectedwrite array<OverrideHasHeightAdvantageStruct> OverrideHasHeightAdvantageCallbacks;
 // End Issue #851
 
+// Start Issue #1138
+struct PrioritizeRightClickMeleeStruct
+{
+	var delegate<PrioritizeRightClickMeleeDelegate> PrioritizeRightClickMeleeFn;
+	var int Priority;
+
+	structdefaultproperties
+	{
+		Priority = 50
+	}
+};
+var protectedwrite array<PrioritizeRightClickMeleeStruct> PrioritizeRightClickMeleeCallbacks;
+// End Issue #1138
+
 delegate EHLDelegateReturn ShouldDisplayMultiSlotItemInStrategyDelegate(XComGameState_Unit UnitState, XComGameState_Item ItemState, out int bDisplayItem, XComUnitPawn UnitPawn, optional XComGameState CheckGameState); // Issue #885
 delegate EHLDelegateReturn ShouldDisplayMultiSlotItemInTacticalDelegate(XComGameState_Unit UnitState, XComGameState_Item ItemState, out int bDisplayItem, XGUnit UnitVisualizer, optional XComGameState CheckGameState); // Issue #885
 delegate EHLDelegateReturn OverrideHasHeightAdvantageDelegate(XComGameState_Unit Attacker, XComGameState_Unit TargetUnit, out int bHasHeightAdvantage); // Issue #851
+delegate EHLDelegateReturn PrioritizeRightClickMeleeDelegate(XComGameState_Unit UnitState, out XComGameState_Ability PrioritizedMeleeAbility, optional XComGameState_BaseObject TargetObject); // Issue #1138
 
 // Start Issue #123
 simulated static function RebuildPerkContentCache() {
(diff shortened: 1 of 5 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -42,10 +89,45 @@ static function XComGameState_Ability GetAvailableEndOfMoveAbilityForUnit(XComGa
...
-}
\ No newline at end of file
+	// Start Issue #1138
+	if(MeleeAbilities.Length > 0)
+	{
+		PrioritizedMeleeAbility = FirstMeleeAbility;
+		CHHelpersObj = class'CHHelpers'.static.GetCDO();
+		CHHelpersObj.TriggerPrioritizeRightClickMelee(MeleeAbilities, PrioritizedMeleeAbility);
+
+		if(PrioritizedMeleeAbility == none)
```
Iridar, line 114: This logic seems weird to me. So you set

`PrioritizedMeleeAbility = FirstMeleeAbility;`

Then run the override delegates, and then you do

`if(PrioritizedMeleeAbility == none)`

but this will be true only if one of the delegates has explicitly set `PrioritizedMeleeAbility = none`, and otherwise it will remain equal to `FirstMeleeAbility`.

It just feels weird. Could you perhaps rewrite the logic to be more consistent?

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -9,6 +9,9 @@
 
 class X2AbilityTrigger_EndOfMove extends X2AbilityTrigger;
 
+// Issue #1138 Property for use by mods to support ability specific weighting in prioritization
+var int RightClickMeleePriority;
```
Iridar, line 13: You've added this value, but it doesn't seem to get used anywhere?

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -42,10 +89,45 @@ static function XComGameState_Ability GetAvailableEndOfMoveAbilityForUnit(XComGa
...
 
-	return None;
-}
\ No newline at end of file
+	// Start Issue #1138
+	if(MeleeAbilities.Length > 0)
+	{
+		PrioritizedMeleeAbility = FirstMeleeAbility;
+		CHHelpersObj = class'CHHelpers'.static.GetCDO();
+		CHHelpersObj.TriggerPrioritizeRightClickMelee(MeleeAbilities, PrioritizedMeleeAbility);
```
Iridar, line 112: You should send as much info as possible with the delegate trigger, specifically the source Unit State. There could be value in sending the  `AvailableAction` too (the `AvailableActions[ActionIndex]`). 

Though I suppose the delegates could get the Unit State from History using the `AbilityState.OwnerStateObject`, but I still think it's worth sending it as is.

There could also be value in moving the whole override code to the place in code where the mods will be able to override which right click ability is used based on the specific target.

Iridar (changes requested):
Either way this would be a useful functionality to have, so I hope you can address the requested changes so this can be added to CHL.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
FearTheBunnies: That value is meant to be used by partner mods to this delegate.  For example, this is the priority calculation from my companion mod:

`// Calculate priority
// If default value then use STANDARD_PRIORITY_MELEE, otherwise grab its priority
AbilityPriority = EndOfMoveTrigger.RightClickMeleePriority;
if(AbilityPriority == 0)
{
	AbilityPriority = const.STANDARD_PRIORITY_MELEE;
}
else if (AbilityPriority < 0)
{
	// We're using SHOT_HUD approach, so treat negative values as extremely high positive values
	AbilityPriority = (AbilityPriority * -1) + const.STANDARD_PRIORITY_MELEE + const.STANDARD_PRIORITY_MELEE;
}`

Iridar: That seems a bit situational. I don't think that's reason enough to add a new variable with unclear purpose into Highlander's code. You can always have a specific priority associated with a specific ability in your own mod, you don't need Highlander to store that information for you.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
FearTheBunnies: I tried to pass either GameRulesCache_Unit or array<AvailableActions> to the delegate, but for whatever reason CHHelpers threw an unknown type compilation error when I tried to use those as parameter types in the delegate. So instead I updated the documentation snippet to include the code needed to load the cache from the history so that folks can do what they want with it.

FearTheBunnies:
Simplified the process significantly by reducing a lot of the clunkiness and now passes UnitState to the delegate allowing for the end consumer to interact with it and the ruleset cache as much as the desire.  Should be much cleaner now.

Iridar (changes requested):
> Simplified the process significantly by reducing a lot of the clunkiness and now passes UnitState to the delegate allowing for the end consumer to interact with it and the ruleset cache as much as the desire. Should be much cleaner now.

It seems alright superficially, I haven't gone through it letter by letter or ran any tests yet. I still think the system would benefit from being able to override which ability to use depending on the target.

This would just mean copying the code from the `X2AbilityTrigger_EndOfMove` to:

 **1) [XComTacticalInput](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/73240bdf7b1b19d009a3962614ae3aa29969ea81/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComTacticalInput.uc#L3116)** - here you would have to retrieve the target State Object from History using `PathingPawn.LastTargetObject.GetReference().ObjectID`.

**2) [XComPathingPawn](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/73240bdf7b1b19d009a3962614ae3aa29969ea81/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComPathingPawn.uc#L515)** - you get `XComGameState_BaseObject TargetObject` right in the function arguments.

The `X2AbilityTrigger_EndOfMove` should still call the override though, just without the target state. You can make the target state object an optional argument for the delegate, or just explicitly pass `none` there. 

This is just so that the override can remain at least somewhat functional in case mod-added code also runs `X2AbilityTrigger_EndOfMove::GetAvailableEndOfMoveAbilityForUnit()`.

FearTheBunnies:
Updated to include the target object now.  Still tried to maintain a minimal footprint in the Highlander itself.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -28,12 +28,86 @@ static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
...
+///     local XComGameState_BaseObject TargetState;
+///     
+///     // To get the current state of the target from the history, use the reference ID. BUT! TargetObject may be NONE
+///     if(TargetObject != NONE)
+///     {
+///         TargetState = `XCOMHISTORY.GetGameStateForObjectID(TargetObject.GetReference().ObjectID);
+///     }
+///
+///     // To get the list of all available actions, load the unit cache from the tactical ruleset
+///     if(`TACTICALRULES.GetGameRulesCache_Unit(UnitState.GetReference(), UnitCache))
```
Iridar, line 76: Arguably shouldn't be here, leave it to individual mods to figure out how they want to find the Ability State they want to use.

Iridar (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
FearTheBunnies: *shrug* Fair enough. Only added it to be nice to folks consuming this.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityTrigger_EndOfMove.uc
```diff
@@ -28,12 +28,86 @@ static function bool AbilityHasEndOfMoveTrigger(X2AbilityTemplate Template)
 }
 
 static function XComGameState_Ability GetAvailableEndOfMoveAbilityForUnit(XComGameState_Unit UnitState)
+{
+	// Begin Issue #1138
+	// Forward the work to the new endpoint to maintain BC
+	return GetAvailableEndOfMoveAbilityForUnitWithTarget(UnitState, none);
```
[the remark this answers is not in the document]
FearTheBunnies: I tweaked the implementation a little by finally realizing there are actually two things at play here. First thing is allowing a mod to fix the "bug" of the right-click priority choosing abilities that are non-intuitive.  The second thing was enhancing the selection ability to allow for the target to be provided increasing the flexibility of the choice.

X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
```diff
@@ -949,6 +964,95 @@ simulated function TriggerOverrideHasHeightAdvantage(XComGameState_Unit Attacker
 }
 // End Issue #851
 
+// Start Issue #1138
+/// HL-Docs: ref:PrioritizeRightClickMelee
+/// # Delegate Priority
+/// You can optionally specify callback Priority. The default Priority is 50.
+///```unrealscript
+///CHHelpersObj.AddPrioritizeRightClickMeleeCallback(PrioritizeRightClickMelee, 45);
+///```
```
Tedster59, line 973: Discord formatting?

Iridar: HL-Docs can render UC code in a similar fashion, example: [this](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2BodyPartTemplateManager.uc#L65-L117) -> [renders like this](https://x2communitycore.github.io/X2WOTCCommunityHighlander/misc/BodyPartTemplateNames/)

Iridar:
Moved the override trigger after the default logic so that the override could allow for a more informed decision without directly replicating base game logic.

Iridar:
This is tested and ready to go, just waiting for @FearTheBunnies to squash the commit with my adjustments and it'll be ready to merge.

Issue #1036, opened 2021-07-11
Allow mods to override which alerts Can be triggered outside of player vision


Grobobobo:
Use Cases are various types of AI to AI activations. Currently mods like Yellow alert 2.0 and AI to AI use an MCO for that purpose.

Iridar:
This issue was resolved in https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/1037, closing.

Issue #620, opened 2019-09-10
Sound is heard by allied units too, not just enemies


RedDoby:
In OnAbilityActivated in XComGameState_Unit I think it would be better to change the following 
GetEnemiesInRange(SoundTileLocation, SoundRange, Enemies);

To create a new function that includes all allies as well as enemies; for mods that use yellow alert such as LWOTC and yellow alert gameplay.  It makes sense that an alien firing guns and throwing grenades would alert other aliens to the position.

RedDoby:
One other thing to note.  This issue depends on Issue #606 in order for the alerts to register.

PR #1165, opened 2023-04-20
Consider sound from allied units, not just enemies

RedDoby:
#620
This enhancement replaces the function `GetEnemiesInRange` with `GetUnitsinRange` in `OnAbilityActivated` which retrieves all units in range that are not civilians, so that sound alerts can be added to alert data for any unit with an ability that produces sound, even from those on the same team. Note this change does nothing by itself when the alerted unit is outside of the vision of XCom. If sound alerts are enabled using #1036 then the alert data will be added to the game history when outside the vision of XCom.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
@@ -997,7 +997,7 @@ static final function array<SoldierClassAbilityType> RebuildSoldierClassAbilityT
 
 /// HL-Docs: ref:ConsiderAlliesforSoundAlerts
 // start issue #620
-// Copied from XComGameState_Unit::GetEnemiesInRange, except will include all units on the alien team within
+// Copied from XComGameState_Unit::GetEnemiesInRange, except will include all units within
 // the specified range.
 static function GetUnitsInRange(TTile kLocation, int nMeters, out array<StateObjectReference> OutEnemies)
 {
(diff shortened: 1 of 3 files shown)


RustyDios:
The Highlander tends to not REMOVE functions due to the backwards compatibility measures for other mods.
What you should do is run the previous GetEnemiesInRange through your new GetUnitsInRange function.

RedDoby:
> The Highlander tends to not REMOVE functions due to the backwards compatibility measures for other mods. What you should do is run the previous GetEnemiesInRange through your new GetUnitsInRange function.
> 
> EDIT :: Scratch that, I see you're not editing the original function, but replacing it where it is called. I need to learn to read
> 
> However does this change mean that now EVERY perk in the game would have to filter out unwanted Allied units?

No, this is strictly for determining `eAC_TakingFire `and `eAC_DetectedSound` alerts and gaining knowledge of the unit creating the sound. It doesn't affect any other part of a perk than that.

RedDoby:
@RustyDios Can you move the changed file X2WOTCCommunityHighlander/X2WOTCCommunityHighlander.x2proj to #1166. I put it on the wrong PR. It is a simple change just adding the XML for the new source file XComGameState_AIUnitData.

Xymanek:
This smells of a backwards compatibility issue, but I don't know enough about the AI/alert system to judge it - some one else would need to chime in.

Regardless, you should update the title of the PR, since adding a function is not the actual purpose of this PR (so the current name is a bit misleading)

RedDoby:
I can always just have a bool to enable the change, that way the original function stays in tact.

RustyDios:
> @RustyDios Can you move the changed file X2WOTCCommunityHighlander/X2WOTCCommunityHighlander.x2proj to #1166. I put it on the wrong PR. It is a simple change just adding the XML for the new source file XGAIPatrolGroup.uc.

I don't know how to do that, or even if I can tbh .. I'm not very git aware

RedDoby:
> > @RustyDios Can you move the changed file X2WOTCCommunityHighlander/X2WOTCCommunityHighlander.x2proj to #1166. I put it on the wrong PR. It is a simple change just adding the XML for the new source file XGAIPatrolGroup.uc.
> 
> I don't know how to do that, or even if I can tbh .. I'm not very git aware

No worries. I will just change the file locally and send another commit.

PR #1168, opened 2023-04-27
Override patrol destinations

### LINKED ISSUE #718: Allow mods to add additional patrol causes in CheckForOverrideDestination()
IN class XGAIPatrolGroup, the function bool CheckForOverrideDestination() determines if a new ThrottlingBeacon alert should be set.  For Yellow Alert Gameplay I would like to have the option to add an additional alert cause to override the new throttling beacon being set, otherwise it interferes with the pod job alerts that are set by YAG.  I believe that LWOTC will also benefit from this change.  In Long War 2 Highlander they added this to the function:
```AlertData.AlertCause == eAC_MapwideAlert_Peaceful ||
                // LWS: If yellow alert is enabled we'll patrol to any valid alert, whichever is highest priority.
				// This is necessary for patrol movement to function: they need *some* destination.
                (class'Helpers_LW'.static.YellowAlertEnabled() && Unit.GetCurrentStat(eStat_AlertLevel) > 0))
```
I need to be able to add an additional alert cause that I use for default PodJob alerts.  
```
		AlertData.AlertCause == eAC_MapwideAlert_Peaceful ||
               //For Alerts set by the pod job manager in YAG
		AlertData.AlertCause == eAC_UNUSED_3))
```
You can see [Here](https://github.com/RedDoby/Yellow-Alert-Gameplay/blob/master/Src/YellowAlert/Classes/YellowAlert_XGAIPatrolGroup.uc) that I have to make a MCO just in YAG.  This would be one less class that I need to override.

RedDoby:
Closes #718 
Allow mods to override patrol locations from other alert data causes. Listeners can set `AllowThisCause` to `true` to allow this AlertCause to be used as a patrol destination for this AI group.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIPatrolGroup.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIPatrolGroup.uc
@@ -1,0 +1,613 @@
+//---------------------------------------------------------------------------------------
+//  *********   FIRAXIS SOURCE CODE   ******************
+//  FILE:    XGAIPatrolGroup.uc    
+//  AUTHOR:  Alex Cheng  --  11/6/2013
+//  PURPOSE: XGAIPatrolGroup:   Alien Pod with tabs on prev and next nodes to path to.
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+class XGAIPatrolGroup extends XGAIGroup
+	native(AI);
+
+var	  array<vector> m_arrNextDest; // Immediate destination list, rebuilt every turn.
+var	  array<vector> m_arrActualDest;
+var   TTile m_kRegroupDest; // Central location to have units regroup at before resuming green alert movement.
+var	  bool m_bRegrouping;
+
+var	  bool m_bRebuildDestinations; // Set on turn init, and after these destinations have been moved to.
+var   vector m_vCurrNode; // Location of next centralized group destination.  (may be offset from actual node due to interference.)
+
+var	  array<int> m_arrUnitsToMove;
+var	  array<TTile> m_arrGroupPath; // the path that the group leader will take
+
+var bool bDisableGroupMove;
+//------------------------------------------------------------------------------------------------
+event NativeInit()
+{
+	SetTickIsDisabled(false);// debug only
+}
+
+function Init(array<int> arrObjIDs, float InInfluenceRange, bool bRefresh=false )
+{
+	BaseInit(arrObjIDs, bRefresh);
+	NativeInit();
+	InfluenceRange = InInfluenceRange;
+}
+
+event InitializePostSpawn(XComGameState NewGameState, bool bRefresh=false)
+{
+	Super.InitializePostSpawn( NewGameState, bRefresh );
+
+	if (m_kPlayer != None && m_kPlayer.m_kNav.m_arrPatrolGroups.Find(self) == -1)
+	{
+		m_kPlayer.m_kNav.m_arrPatrolGroups.AddItem(self);
+	}
+}
+
+//------------------------------------------------------------------------------------------------
+function InitTurn()
+{
+	super.InitTurn();
+	m_bRebuildDestinations=true;
+	UpdateRegroup();
+	m_arrGroupPath.Length = 0;
+	bDisableGroupMove = false;
+}
+
+//------------------------------------------------------------------------------------------------
+function UpdateRegroup()
+{
+	local XComGameState_Unit kUnitState;
+	local int iObjID, iIdx, iMaxTileRadius;
+	local TTile kDest;
+	local vector vLoc;
+	iIdx = 0;
+	if (m_arrLastAlertLevel.Length == m_arrUnitIDs.Length)
+	{
+		// Check alert levels if it is time to start regrouping.
+		if (!m_bRegrouping)
+		{
+			foreach m_arrUnitIDs(iObjID)
+			{
+				kUnitState = XComGameState_Unit(`XCOMHISTORY.GetGameStateForObjectID(iObjID));	
+				if (kUnitState.GetCurrentStat(eStat_AlertLevel) == 0 && m_arrLastAlertLevel[iIdx] == 2)
+				{
+					m_bRegrouping = true;
+				}
+				kDest.X += kUnitState.TileLocation.X;
+				kDest.Y += kUnitState.TileLocation.Y;
+				kDest.Z += kUnitState.TileLocation.Z;
+			}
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIPatrolGroup.uc
```diff
@@ -133,9 +145,20 @@ function bool CheckForOverrideDestination(out vector CurrDestination, int Object
 		AIData = XComGameState_AIUnitData(History.GetGameStateForObjectID(AIDataID));
 		if( AIData.GetPriorityAlertData(AlertData) )
 		{
+			OverrideTuple = new class'XComLWTuple';
```
Iridar, line 148: The code you have now will probably work and be sufficient for your purpose, but there's a number of ways it can be improved.

1. The current Highlander policy regarding override events and Tuples is to form the Tuple and trigger the event in a separate private function, **[for example](https://github.com/long-war-2/X2WOTCCommunityHighlander/blob/b35a07f90f5786b517941026fc6005744adc31e1/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc#L1120)**.

2. Your override event allows considering more alert causes, but it doesn't allow to *not* consider certain alert causes. I know nothing about AI and the alert system, so I'm not sure how much value would be in that, but allowing it wouldn't take anything away from you want to do, so there's no reason not to do it. Normally it would be done like this:

```
OverrideTuple.Data[1].Kind = XComLWTVBool;
OverrideTuple.Data[1].b = AlertData.AlertCause == eAC_ThrottlingBeacon || AlertData.AlertCause == eAC_MapwideAlert_Hostile || AlertData.AlertCause == eAC_MapwideAlert_Peaceful;

TriggerEvent

return OverrideTuple.Data[1].b;
```

3. You could probably send more information with the event, such as using the `XComGameState_AIUnitData AIData` as `EventSource`. Again, no idea how much value is in that.
4. There could be value in allowing to override the `CurrDestination` with the event as well. You could do 
```
CurrDestination = `XWORLD.GetPositionFromTileCoordinates(AlertData.AlertLocation);
```
before you trigger the override the event, then send the value of the `CurrDestination` in the Tuple, and then take it back from the Tuple. 

Addressing the (1) is mandatory, so please do that, as for the other stuff, they're just suggestions, other people who know more about AI / alerts are welcome to weigh in on that.

Iridar (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIPatrolGroup.uc
RedDoby: Thanks for the suggestions, I opted to go with the basic changes. This is just eliminating an MCO and I don't see anyone else using this.

Iridar:
That's a bit disappointing, because right the event seems of very limited use and does not seem to do what its name suggests it does. I.e. you don't get to override the destination, you're overriding whether a particular alert cause is considered. At least that's how it looks to me. But alright, if that's what you need.

Just one final request at this point: please squash the two last commits, so that are only two commits in this PR, the first one where you add the new file to source, and second where you add the event itself, and I'll merge this.

RedDoby:
> That's a bit disappointing, because right the event seems of very limited use and does not seem to do what its name suggests it does. I.e. you don't get to override the destination, you're overriding whether a particular alert cause is considered. At least that's how it looks to me. But alright, if that's what you need.
> 
> Just one final request at this point: please squash the two last commits, so that are only two commits in this PR, the first one where you add the new file to source, and second where you add the event itself, and I'll merge this.

Done

Iridar:
Uhh still seeing three commits in this PR

RedDoby:
> Uhh still seeing three commits in this PR

Now I have 6 commits. I am not sure what is happening or how to fix this. SIgh..

Iridar:
> > Uhh still seeing three commits in this PR
> 
> Now I have 6 commits. I am not sure what is happening or how to fix this. SIgh..

You need to do an interactive rebase and squash all commits besides the one where you add new file to source. In your case it would be:

git rebase -i HEAD~6

Then you'll need to edit a couple of documents, comments will explain. Once you're done, force push.

RedDoby:
> > > Uhh still seeing three commits in this PR
> > 
> > 
> > Now I have 6 commits. I am not sure what is happening or how to fix this. SIgh..
> 
> You need to do an interactive rebase and squash all commits besides the one where you add new file to source. In your case it would be:
> 
> git rebase -i HEAD~6
> 
> Then you'll need to edit a couple of documents, comments will explain. Once you're done, force push.

I am going around in circles trying to google how to run this command you suggested. Is it from the windows command line or is there some other tool? Maybe you can point me in the right direction.

Iridar:
> > > > Uhh still seeing three commits in this PR
> > > 
> > > 
> > > Now I have 6 commits. I am not sure what is happening or how to fix this. SIgh..
> > 
> > 
> > You need to do an interactive rebase and squash all commits besides the one where you add new file to source. In your case it would be:
> > git rebase -i HEAD~6
> > Then you'll need to edit a couple of documents, comments will explain. Once you're done, force push.
> 
> I am going around in circles trying to google how to run this command you suggested. Is it from the windows command line or is there some other tool? Maybe you can point me in the right direction.

What git client are you using?

RedDoby:
> > > > > Uhh still seeing three commits in this PR
> > > > 
> > > > 
> > > > Now I have 6 commits. I am not sure what is happening or how to fix this. SIgh..
> > > 
> > > 
> > > You need to do an interactive rebase and squash all commits besides the one where you add new file to source. In your case it would be:
> > > git rebase -i HEAD~6
> > > Then you'll need to edit a couple of documents, comments will explain. Once you're done, force push.
> > 
> > 
> > I am going around in circles trying to google how to run this command you suggested. Is it from the windows command line or is there some other tool? Maybe you can point me in the right direction.
> 
> What git client are you using?

I was using GitHub. I found git bash and gitextensions and installed them both. Now I can navigate to the local repository in GitBash but when I try and run the command you suggested it states "fatal: this operation must be run in a work tree".

Iridar:
> I was using GitHub. I found git bash and gitextensions and installed them both. Now I can navigate to the local repository in GitBash but when I try and run the command you suggested it states "fatal: this operation must be run in a work tree".

Yeah as far as I know you can't do elaborate things like squashing specific commits via github interface. I don't really know anything about git extensions or git bash, but if you ping me on Discord, I can explain how to use Source Tree, it's a standalone app with a proper graphical interface for using git.

Iridar:
Though at this point it might be easier to just squash these myself, let me try that.

Iridar (approved):

RedDoby:
> Though at this point it might be easier to just squash these myself, let me try that.

Sweet. Thank you.

PR #1169, opened 2023-04-27
Add option to send sound alerts data to allies

### LINKED ISSUE #620: Sound is heard by allied units too, not just enemies

RedDoby:
Closes #620 
Allow option to override original function in `OnABilityActivated` to pass sound alerts from allies to other allies.


X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
```diff
@@ -994,3 +997,14 @@ static final function array<SoldierClassAbilityType> RebuildSoldierClassAbilityT
 	return AbilityTypes;
 }
 // End Issue #815
+
+/// HL-Docs: ref:ConsiderAlliesforSoundAlerts
+// start issue #620
+static function bool ConsiderAlliesforSoundAlerts()
```
Iridar, line 1003: Is there are a particular reason you created a function to check the value of a config variable? 

You could just do `class'CHHelpers'.default.bConsiderAlliesforSoundAlerts` wherever you need to check this. 

But even if using a function, it can be simplified to:

```
static function bool ConsiderAlliesforSoundAlerts()
{
    return default.bConsiderAlliesforSoundAlerts;
}
```
Also, your function has no return value for when `default.bConsiderAlliesforSoundAlerts` is `false`. It will still return `false` as that is the default value for `bool` values, but it's not a proper way to do things.

Iridar (changes requested):

X2WOTCCommunityHighlander/Config/XComGame.ini
```diff
@@ -32,13 +32,13 @@ bEnableVersionDisplay=true
 ;;; HL-Docs: ref:ArmorEquipRollDLCPartChance
 ;+CosmeticDLCNamesUnaffectedByRoll=MyDLCName
 
-;;; HL-Docs: ref:DisableExtraLOSCheckForSmoke
```
Iridar, line 35: What are these three lines concerning issue #669 are doing here?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -10902,6 +10919,47 @@ function GetUnitsInRangeOnTeam(ETeam Team, TTile kLocation, int nMeters, out arr
...
+
+			RadiiSumSquared = fDistSq;
+			if( UnitHearingRadius != 0 )
+			{
+				RadiiSumSquared = Square(AudioDistanceRadius + UnitHearingRadius);
+			}
+
+			if( VSizeSq(vLoc - vCenter) < RadiiSumSquared )
+			{
+				OutEnemies.AddItem(kUnit.GetReference());
```
Iridar, line 10956: Since you're collecting units, not enemies, would make sense for the array to be called OutUnits rather than OutEnemies.

Iridar:
Please remove the unrelated changes to issue #669 HL-Docs from this PR. It's actually breaking the docs, btw. Other than that, should be good to go.

Issue #1035, opened 2021-07-10
WotC bug + Solution : Problems Editing Covert Action Soldier Loadouts with a Controller


kdm2k6:
**General problem**: The issue is that, while using a controller, you can edit the loadout of the 1st soldier slot for any given Covert action; however, you can not edit the loadout of the 2nd, 3rd ... soldier slot. I have come across this problem while testing LWotC; however, I believe it is a WotC bug. Additionally, I have confirmed the solution by using a modified highlander.

In order to replicate this bug do the following :

1. Go to the Covert Actions screen.
2. Choose a Covert Action which requires at least 2 soldiers.
3. Using the A button, add a soldier to each soldier slot.
4. Press the Y button while the 1st soldier slot is active; notice that the loadout screen for that soldier is brought up. This is working as intended.
5. Go back and press the Y button while the 2nd soldier slot is active; notice that the loadout screen for the 1st soldier is brought up. This is a problem.

**Why this occurs** : Within the `UIScreen`, `UICovertActions`, the code : `SlotContainer.Navigator.SelectedIndex` is used to determine which particular slot within the slot container is currently selected. This can be seen within `OnUnrealCommand`, where the A button :

1. Determines which `UICovertActionStaffSlot` is selected via the call : `SlotContainer.Navigator.SelectedIndex`
2. Calls `HandleClick `on this slot, sending in the parameter "theButton" via : `SlotContainer.ActionSlots[SlotContainer.Navigator.SelectedIndex].HandleClick("theButton")`.
3. `UICovertActionStaffSlot.HandleClick` then determines if this slot should be filled with a soldier, or emptied, depending upon its current state.

Unfortunately in several locations, including the `EditSoldierLoadout` function, the selected slot is determined via `Navigator.SelectedIndex` rather than `SlotContainer.Navigator.SelectedIndex`. This results in the index value 0 always being returned and the 1st soldier slot being "dealt with". I never tested exactly what `Navigator.SelectedIndex` referenced, but I would guess the screen's navigator is pointing to the slot container itself.

**Solution** : There are 3 problematic lines of code within `UICovertActions`.

1.] Within the function : `RefreshNavigation`.
`if( SlotContainer.ActionSlots[Navigator.SelectedIndex].CanEditLoadout() )` should become
`if( SlotContainer.ActionSlots[SlotContainer.Navigator.SelectedIndex].CanEditLoadout() )`

[...]

PR #1170, opened 2023-05-09
Fix editing covert action soldier loadouts with a controller

araujorm:
While using a controller, you can edit the loadout of the 1st soldier slot for any given covert action; however, you can not edit the loadout of the 2nd or 3rd.

As discovered by [kdm2k6](https://github.com/kdm2k6) and shared in issue https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/issues/1035 this is a WOTC upstream bug, however they also discovered the solution.
Since they never submitted a pull request, I made one and tested it, but otherwise the full credit for the solution should go to them.

UICovertActions.uc file from SDK is needed to be added, and changed with the code to the working solution.

I just tested a build based on this and it works (with unmodified LWOTC).

Fixes:
https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/issues/1035
https://github.com/long-war-2/lwotc/issues/1378

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActions.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActions.uc
@@ -463,7 +463,9 @@ simulated function RefreshNavigation()
 	//bsg-jneal (3.10.17): add edit loadout navhelp if on an editable soldier
 	if( `ISCONTROLLERACTIVE && bIsSelectingSlots && SlotContainer.ActionSlots.Length > 0)
 	{
-		if( SlotContainer.ActionSlots[Navigator.SelectedIndex].CanEditLoadout() )
+		/// HL-Docs: ref:Bugfixes; issue:1035
+		/// Use `SlotContainer.Navigator.SelectedIndex` instead of `Navigator.SelectedIndex` to allow controller users editing all soldiers' loadouts.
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActions.uc
```diff
@@ -463,7 +463,7 @@ simulated function RefreshNavigation()
 	//bsg-jneal (3.10.17): add edit loadout navhelp if on an editable soldier
 	if( `ISCONTROLLERACTIVE && bIsSelectingSlots && SlotContainer.ActionSlots.Length > 0)
 	{
-		if( SlotContainer.ActionSlots[Navigator.SelectedIndex].CanEditLoadout() )
+		if( SlotContainer.ActionSlots[SlotContainer.Navigator.SelectedIndex].CanEditLoadout() )
```
Iridar, line 466: All changes within Highlander have to be documented. Please add the following lines above each line you have changed.
```
/// HL-Docs: ref:Bugfixes; issue:1035
/// Use `SlotContainer.Navigator.SelectedIndex` instead of `Navigator.SelectedIndex` to allow controller users editing all soldiers' loadouts.
```

Iridar (changes requested):

Iridar (approved):
LGTM.

PR #1173, opened 2023-05-14
Add an event to localize new Item Categories

### LINKED ISSUE #1112: Add a way to localize new item categories
There's a CHL event too localization for new weapon categories, but no such thing for item categories. 

Nothing is stopping mods from creating a custom X2ItemTemplate and overriding `function string GetLocalizedCategory()`, but should that really be required for such a small thing?

Iridar:
Fixes #1112


Iridar:
In retrospect, using the event system for this is a huge overkill, especially in the presence of the `Localize()` function, but let's keep the event for the sake of consistency with similar implementation in `X2WeaponTemplate'.

PR #1179, opened 2023-05-14
Fix TriggerOverrideProjectileInstance()

### LINKED ISSUE #1053: XGUnitNativeBase::TriggerOverrideProjectileInstance is bugged
[Relevant code.](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/de17827ec72dd9662170c96b8f74c41d534bf212/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnitNativeBase.uc#L1635)

Original issue: #829

The docs mention that the purpose of this override is to allow listeners to override parameters of the original function, but the override obviously doesn't function like that, as none of the override arguments are set as `out`, and are not updated after triggering the event.

Moreover, the `bPreventProjectileSpawning` parameter is not updated after triggering the event either, so the advertised functionality of being able to prevent the projectile from spawning doesn't work. 

This can be addressed in two ways:

1. Conservative - update `bPreventProjectileSpawning` and correct docs by removing the "override parameters" line.
2. Expansive - convert all override arguments to `out` and update all of them after triggering the event.

Suggestions, preference?

Iridar:
Fixes #1053

Decided to go with the "expansive" approach by making almost all of the stuff in Tuple `inout`.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnitNativeBase.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnitNativeBase.uc
@@ -1611,8 +1611,9 @@ function AddProjectileVolley(AnimNotify_FireWeaponVolley Notify)
 	}
 }
 
+// Start Issue #829
 /// HL-Docs: feature:OverrideProjectileInstance; issue:829; tags:tactical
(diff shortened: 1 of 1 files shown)


Iridar:
Note that this is technically a BC break, but I don't think there are any mods that actually use this functionality, so I figured making a second, expanded override event for the same thing would be excessive.

PR #1194, opened 2023-06-11
IsIncapacitated check to GatherUnitsToMove #1193

### LINKED ISSUE #1193: Enemy AI chokes on unconcious units
While not super common, if an AI team (Aliens, Raider Factions) has a unit put into the unconscious or ShadowbindUnconcious effects, the AI doesn't know how to process it, causing the AI turn to hang and then be skipped when the failsafe timeout is hit.  Mods such as  [No Unconscious AI](https://steamcommunity.com/sharedfiles/filedetails/?id=2761606669) attempt to fix this by stopping the Unconscious effect from being applied to the alien team, or just [removing it from the things that cause it](https://steamcommunity.com/sharedfiles/filedetails/?id=2574868490) (Muton melee, Stun Lancer melee, for example).  Ideally this should be fixed in Highlander so that the AI can just skip that specific unit's actions if it is unconscious.

Tepa00:
This allows the AI to skip incapacitated units in a group which has other units able to act.

fixes #1193

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIPlayer.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIPlayer.uc
@@ -897,7 +897,7 @@ simulated function GatherUnitsToMove()
 		//
 		// Issue #510 - Removed the DecisionStartHistoryIndex check as it prevents pod leaders from using
 		// any reflex actions they may have been granted.
-		// Issue #1193
+		
 		/// HL-Docs: ref:Bugfixes; issue:1193
 		/// Added `IsUnconcious` and `IsBleedingOut` check to skip AI units in group who are Unconcious or Bleeding out
 		/// as the AI processing hangs here until it times out, which causes the rest of the group to be skipped.
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIPlayer.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIPlayer.uc
@@ -899,7 +899,7 @@ simulated function GatherUnitsToMove()
 		// any reflex actions they may have been granted.
 		// Issue #1193 - Added IsIncapacitated check to skip units in group who are either Unconcious or Bleeding out
 		// as the AI processing hangs here until timeout and so the whole group gets skipped.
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIPlayer.uc
```diff
@@ -897,7 +897,9 @@ simulated function GatherUnitsToMove()
 		//
 		// Issue #510 - Removed the DecisionStartHistoryIndex check as it prevents pod leaders from using
 		// any reflex actions they may have been granted.
-		if(bDead || UnitState.bRemovedFromPlay || UnitState.NumAllActionPoints() == 0) // || kBehavior == None || kBehavior.DecisionStartHistoryIndex > kAIPlayerData.m_iLastEndTurnHistoryIndex)
+		// Issue #1193 - Added IsIncapacitated check to skip units in group who are either Unconcious or Bleeding out
+		// as the AI processing hangs here until timeout and so the whole group gets skipped.
+		if(bDead || UnitState.bRemovedFromPlay || UnitState.NumAllActionPoints() == 0 || UnitState.IsIncapacitated()) // || kBehavior == None || kBehavior.DecisionStartHistoryIndex > kAIPlayerData.m_iLastEndTurnHistoryIndex)
```
Iridar, line 902: Are you positive that `UnitState.IsIncapacitated()` returns true only if the unit is unconscious or bleeding out?

Tepa00: Yes

Iridar: It's not documented though. You're probably right, but I'd rather be extra careful. I think it would be better to add `UnitState.IsUnconscious()` and `UnitState.IsBleedingOut()`. Both are native functions, so the performance loss from calling two separate methods should be negligible.

Tepa00: Done.

Iridar (changes requested):
The docs still mention the `IsIncapacitated()`, please update. Also I think this should have the 

/// HL-Docs: ref:Bugfixes; issue:1193

tag.

Iridar (approved):

Iridar:
Approved, to be squash merged, ideally test first.

Iridar:
Tested, seems to be working.

PR #1204, opened 2023-06-24
Issue #1111 - Add localisation to Options menu for Shen's Last Gift

### LINKED ISSUE #1111: Shen's Last Gift cosmetic options slider is displayed as "DLC_3"
On the settings page, the game displays sliders to control the appearance frequency of cosmetic parts from a certain mod or DLC. 

Original code is arguably bugged, and does not display the slider if the DLC name is missing localization.

Highlander's issue #150 (I think?) fixes that problem, and makes the game display the slider using the DLC name itself if the localized DLC name is unavailable. 

This makes the entry for Shen's Last Gift appear as "DLC_3", which apparently scares some people into thinking it's a bug and a problem that needs to be solved ([example](https://github.com/long-war-2/lwotc/issues/1469)).

[Example Image](https://user-images.githubusercontent.com/11407231/140824512-b867fa17-822d-4e64-b845-d721779a0674.jpg)

Personally I don't think it's a problem worth scratching, but if anybody wants to fix it, it can be fixed by Highlander by shipping a localized DLC name for Shen's Last Gift. 

Keeping in mind that even showing a slider for the SLG is of arguable usefulness, since the only cosmetics added by that mod are for SPARKs, so unless the player has cosmetic mods that give SPARKs a whole new appearance, the slider will be useless. And even then, I'm not sure setting the slider to 0 would be useful, since SPARKs generated by the game may or may not use SLG body parts by default either way.

BlackDog86:
Fixes #1111 

"Shen's Last Gift" appears as "DLC_3" in the options menu. This PR adds some localisation text to all languages & mouseover text for all languages except CHN/CHT/KOR/JPN as I'm not familiar enough with the characters in these languages to add or remove stuff without (probably) completely changing the meaning. No attention has been given to whether the slider actually does anything or not in-game (all text assumes that the slider increases the chances of getting cosmetic SPARK options from the DLC in the same way as the other sliders).


Iridar:
I appreciate your enthusiasm, but the slight problem with this is this is the main Highlander is technically not the proper place to do it. If we had a Shen's Last Gift Highlander, that's where this should have gone. Until then, the best place for a fix like that is in some less strict community bugfix mod, such as the Core Collection Meta Mod. 

On the other hand, this is just localization, and not a code change, and will have literally no impact if the user doesn't have SLG, so maybe it's not a big deal?

Iridar:
Yeah, sorry, even if it's just a localization change, I don't want to set the precedent of adding things that are relevant only to a DLC into main Highlander. If ever have a Highlander for Shen's Last Gift, we'll fix it there. Until then, I'll fix it in Core Collection Meta Mod.

PR #1205, opened 2023-06-24
Add an override to force Euro-style date strings

### LINKED ISSUE #1191: X2StrategyGameRulesetDataStructures GetDateString() has no option to force EuroStyleData
X2StrategyGameRulesetDataStructures GetDateString() has no option to force EuroStyleData for the INT language for anyone using INT that is not located in the USA.
I propose a simple solution to add a config variable (around line 1083 with the other DateTime Handling variables) that will be read and actioned in the function at line 1627

BlackDog86:
Fixes #1191

Adds a simple config to XComGame.ini:
bForceEuroDateStrings (If true - INT Dates will be displayed in the Euro format DD/MM/YYYY)

This alters the output of GetDateString() which shows the date in various places in the game (memorial / logs etc.).  Further alterations are left to other mods / user discretion.

### DIFF
--- a/X2WOTCCommunityHighlander/Config/XComGame.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGame.ini
@@ -256,3 +256,7 @@ bEnableVersionDisplay=true
 ; destroying smoke trails (and similar) prematurely.
 ; Values below 5 (seconds) will be ignored.
 ;ProjectileParticleSystemExpirationDefaultOverride=60
+
+;;; HL-Docs: ref:EuroDateOverride
+; Uncomment the following line to set European date style for INT localization
+; bForceEuroDateStrings = true
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
@@ -220,6 +220,9 @@ var config bool DisableExtraLOSCheckForSmoke;
 var config bool DisableExtraLOSCheckForPoison;
 // End Issue #669
 
+// Variable for Issue #1191
+var config bool bForceEuroDateStrings;
+
 // Start Issue #720
 /// HL-Docs: ref:ProjectilePerformanceDrain
 struct ProjectileParticleSystemExpirationOverride
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
@@ -1,0 +1,3317 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2StrategyGameRulesetDataStructures.uc
+//  AUTHOR:  Ryan McFall  --  02/20/2013
+//  PURPOSE: Container class that holds data structures common to various aspects of the
+//           strategy game rules set
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+class X2StrategyGameRulesetDataStructures 
+	extends object 
+	native(Core) 	
+	dependson(XGTacticalGameCoreNativeBase, XComGameState)
+	config(GameData);
+
+// For appearance chance of persistent units
+struct native AppearChance
+{
+	var int Difficulty;
+	var int MissionCount;
+	var int PercentChance;
+};
+
+// For linking template names to narrative moments
+struct native TemplateNarrative
+{
+	var name TemplateName;
+	var string Narrative;
+};
+
+// Soldier Will mental states
+enum EMentalState
+{
+	eMentalState_Shaken,
+	eMentalState_Tired,
+	eMentalState_Ready,
+};
+
+// Soldier Com Int states
+enum ECombatIntelligence
+{
+	eComInt_Standard,
+	eComInt_AboveAverage,
+	eComInt_Gifted,
+	eComInt_Genius,
+	eComInt_Savant,
+};
+
+// Resistance Faction Influene states
+enum EFactionInfluence
+{
+	eFactionInfluence_Minimal,
+	eFactionInfluence_Respected,
+	eFactionInfluence_Influential,
+	eFactionInfluence_MAX,
+};
+
+// For facilities to define their staff slots
+struct native StaffSlotDefinition
+{
+	var name StaffSlotTemplateName; // Name of template
+	var name LinkedStaffSlotTemplateName; // Name of template of linked slot
+	var bool bStartsLocked; // Staff Slot must be unlocked
+};
+
+// Linked Templates
+struct native LinkedTemplates
+{
+	var name TemplateA;
+	var name TemplateB;
+};
+
+// Soldier Bond
+struct native SoldierBond
+{
+	var StateObjectReference Bondmate; // Who the bond is with
+	var float Compatibility; // How quickly cohesion increases
+	var int BondLevel; // discrete level of bond (levels have different benefits)
+	var int Cohesion; // raw score of bond
+	var int MostRecentCohesionGain; 
+	var bool bSoldierBondLevelUpAvailable;
+	var bool bBondJustBecameAvailable;
+};
+
+// Min/Max Float
+struct native MinMaxFloat
+{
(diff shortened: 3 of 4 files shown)


X2WOTCCommunityHighlander/X2WOTCCommunityHighlander.x2proj
Iridar, line 1: This is going to be a problem, the change to the .x2proj file should be in the same commit where you add the new file. 

The usual workflow is: 

1. Create branch
2. Add the file via Modbuddy
3. Close Modbuddy saving changes - that will apply the changes of adding the new file to the .x2proj file. Or maybe clicking File -> Save All is enough, but I usually restart Modbuddy to be sure.
4. Commit.
5. Continue working.

Iridar:
I'm confused, why did you have to make localization changes for this?

Iridar:
I take it you didn't realize you could do

```
`XEXPAND.ExpandString("<XGParam:IntValue1/><XGParam:StrValue1/!SeparatorString/><XGParam:IntValue2/><XGParam:StrValue1/!SeparatorString/><XGParam:IntValue0/>");
```

The goal is to make the date display configurable, right? As I understand it, the order of these tags determines the order in which day/month/year are displayed.

The strings with these tags are already localized. Meaning that anyone can go into their localization files and reorder the tags as they want. This can also be done as a standalone mod that will reorder these tags for the user. 

I see zero reason to involve Highlander in this.

X2WOTCCommunityHighlander/X2WOTCCommunityHighlander.x2proj
BlackDog86: OK didn't realise that - looks like probably gonna have to kill this branch altogether anyway so will take that under advisement.

BlackDog86:
Significantly reduced codebase for this PR & kept focus to the original issue. Hopefully this is better.

X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
```diff
@@ -220,6 +220,9 @@ var config bool DisableExtraLOSCheckForSmoke;
 var config bool DisableExtraLOSCheckForPoison;
 // End Issue #669
 
+/// HL-Docs: feature:AddINTEuroDateOverRide; issue:1191;
```
Iridar, line 223: Nitpicking, but just having

// Variable for Issue #1191

Would be enough. Stuff marked as HL-Docs will be rendered on the [Highlander documentation](https://x2communitycore.github.io/X2WOTCCommunityHighlander/) page, and there's no need for that in this case.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
```diff
@@ -1624,8 +1624,8 @@ static function string GetDateString(TDateTime kDateTime, optional bool bShortFo
 		Lang = GetLanguage();
 
 		kTag.StrValue1 = "/";
-
-		if  ( Lang == "FRA" || Lang == "ITA" || Lang == "ESN")
+	// issue #1191 - Add config option to force Euro date style with INT localisation
```
Iridar, line 1627: Since this is the actual code change with new logic, having a HL-Docs here would be reasonable.

X2WOTCCommunityHighlander/Config/XComGame.ini
```diff
@@ -256,3 +256,7 @@ bEnableVersionDisplay=true
 ; destroying smoke trails (and similar) prematurely.
 ; Values below 5 (seconds) will be ignored.
 ;ProjectileParticleSystemExpirationDefaultOverride=60
+
+; HL-Docs: feature:AddINTEuroDateOverRide; issue:1191;
```
Iridar, line 260: Super nitpicking, but "override" is one word, so camelcase for it would be "Override", not "OverRide".

Iridar (changes requested):
Sorry for nitpicking, but everything in Highlander has to be just right.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
BlackDog86: done

X2WOTCCommunityHighlander/Config/XComGame.ini
BlackDog86: done

X2WOTCCommunityHighlander/Config/XComGame.ini
```diff
@@ -256,3 +256,7 @@ bEnableVersionDisplay=true
 ; destroying smoke trails (and similar) prematurely.
 ; Values below 5 (seconds) will be ignored.
 ;ProjectileParticleSystemExpirationDefaultOverride=60
+
+; HL-Docs: feature:AddINTEuroDateOverride; issue:1191;
```
Iridar, line 260: HL-Docs in comments must be preceded by triple semicolon, or else they won't be picked up.

Also, since you already have an instance of HL-Docs that specifies feature and issue number, in other instances of HL-Docs you should just reference that feature name.

`/// HL-Docs: feature:AddINTEuroDateOverride; issue:1191`

So in the end you should have:

`;;; HL-Docs: ref:AddINTEuroDateOverride;`

I also recommend renaming the feature to `EuroDateOverride`, "Add" is redundant and "INT" doesn't actually have anything to do with the feature, as it's applied regardless.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
```diff
@@ -0,0 +1,3317 @@
...
+	{
+		kTag.StrValue0 = GetMonthString(,,kDateTime);
+		kTag.IntValue0 = kDateTime.m_iDay;
+		kTag.IntValue1 = kDateTime.m_iYear;
+		kTag.IntValue2 = kDateTime.m_iMonth;
+
+		Lang = GetLanguage();
+
+		kTag.StrValue1 = "/";
+		/// HL-Docs: feature:AddINTEuroDateOverride; issue:1191 - Add config option to force Euro date style with INT localisation
```
Iridar, line 1627: This should be split into two lines, the HL-Docs header and the comment itself.

X2WOTCCommunityHighlander/Config/XComGame.ini
BlackDog86: Update

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
BlackDog86: Done

X2WOTCCommunityHighlander/Config/XComGame.ini
```diff
@@ -256,3 +256,7 @@ bEnableVersionDisplay=true
 ; destroying smoke trails (and similar) prematurely.
 ; Values below 5 (seconds) will be ignored.
 ;ProjectileParticleSystemExpirationDefaultOverride=60
+
+;;; HL-Docs: feature:EuroDateOverride;
```
Iridar, line 260: Replace `feature` with `ref` please, only one HL-Docs entry per unique issue should have the `feature` keyword, followed by the issue number.

In other words, it's either:

/// HL-Docs: feature:EuroDateOverride; issue:1191

which can be only one, as I understand it, or:

/// HL-Docs: ref:EuroDateOverride;

Iridar (changes requested):

X2WOTCCommunityHighlander/Config/XComGame.ini
BlackDog86: Done

Iridar (approved):

Iridar:
@BlackDog86 
All good. Could you please squash the last two commits, so there are only two, one with new file, and one with the rest of the changes?

BlackDog86:
Squashed & comments updated

Iridar:
Sorry, I haven't caught this earlier, `class'CHHelpers'.default.bForceEuroDateStrings == true` is wasteful, `class'CHHelpers'.default.bForceEuroDateStrings` is enough, it's already a bool value.

Iridar (approved):

Issue #1082, opened 2021-10-18
Map plots for Avenger Defence missions not random


Dragon-32:
In my current campaign I have had 2 Avenger Defence missions, both were on the same plot ("plot_WLD_AvDef_Mount_EMM_01"). 

In my previous campaign I _think_ the single Avenger Defence mission I had was on the same plot (the Photobooth poster seems to back it up) but that mission was in August 2020.

The base game has four Avenger Defence plots:
Plot_WLD_AvDef_LgObj_Ravine
Plot_WLD_AvDef_LgObj_Stream
Plot_WLD_AvDef_MdObj_Ravine
Plot_WLD_AvDef_MdObj_Stream

I'm running two mods that add to that list, [Maps by Vozati [WotC Edition]](https://steamcommunity.com/sharedfiles/filedetails/?id=1132904326):
Plot_WLD_AvDef_Ravine_Voz_1C
Plot_WLD_AvDef_Large_Voz_1C

And [Even More Maps! [WOTC]](https://steamcommunity.com/sharedfiles/filedetails/?id=1139192059):
plot_WLD_AvDef_Suffer_EMM_01
plot_WLD_AvDef_Suffer_EMM_02
plot_WLD_AvDef_Flood_EMM_01
plot_WLD_AvDef_Mount_EMM_01

Assuming I'm right about my previous campaign then that's a 1 in 1000 chance of the same plot being used three times in a row, isn't it? Even if I'm wrong then it's a 1 in 100 for the same plot to be used for two missions, I think?

_I know. Someone complaining about RNG in an XCOM game..._

I posted the plot names in the order they appear in their respective XComPlots.ini (mod or base game)

My local EMM folder was created 7th August 2019, Vozati's 23rd October 2017.

So, I know [XCOM 2 load order](https://robojumper.github.io/too-real/load-order/) is weird but I think it's a bit too much of a coincidence that in all cases the plot chosen is the last in the list if the list is assembled in either Workshop ID or Subscription Date order.

I think there's enough here for someone to take a look at how plots are chosen for this mission type? See if it really is random.

Apologies if this is out of scope for the Highlander but I don't know enough about the game's innards to know.

It always seem to be the Xenoform biome too, maybe that's related...

h4ilst0rm:
I was intrigued and can cofirm this seems to be the case.
[...]

PR #1206, opened 2023-06-25
Fix plots not being marked used for CAD and UFO avenger missions

Tedster59:
Fixes Issue #1082.

Since nothing uses Seamless traveling anymore (per a discord message from robojumper), Missions that don't involve a dropship travel animation don't mark plots as used, leading to all UFOs and Chosen Avenger Defenses for a campaign being on the same plot.  This PR moves the MarkPlotsUsed code outside of the `bShowDropshipInteriorWhileGeneratingMap` check so that all missions have plots marked as used properly.


Tedster59:
Closing to more properly structure my own repos with different branches.

PR #1207, opened 2023-06-25
Add MarkPlotUsed call for missions that don't use the dropship

Tedster59:
Fixes issue #1082

Since nothing uses Seamless traveling anymore (per a discord message from robojumper), Missions that don't involve a dropship travel animation don't mark plots as used, leading to all UFOs and Chosen Avenger Defenses for a campaign being on the same plot. This PR moves the MarkPlotsUsed code outside of the bShowDropshipInteriorWhileGeneratingMap check so that all missions have plots marked as used properly.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
@@ -2974,7 +2974,9 @@ Begin:
 			sleep(0.0f);
 		}
 
-		//MarkPlotUsed(); // Tedster - commented out for Issue #1082
+		/// HL-Docs: ref:Bugfixes; issue:1082
+		/// MarkPlotUsed() - commented out original call and moved it outside of the if() statement
+		
 
 		`log("CreateTacticalGame: Dropship loaded");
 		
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
```diff
@@ -2974,7 +2974,7 @@ Begin:
 			sleep(0.0f);
 		}
 
-		MarkPlotUsed();
+		//MarkPlotUsed(); // Tedster - commented out for Issue #1082
```
Iridar, line 2977: Please alter the comment to:

// Issue #1082 - moved outside the if() statement.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
```diff
@@ -3000,6 +3000,10 @@ Begin:
 		`XTACTICALSOUNDMGR.StopHQMusicEvent();
 	}
 
+	// Start Issue #1082 - Move the MarkPlotUsed call outside of the if block so that missions that don't show the dropship interior are also handled
```
Iridar, line 3003: Please use the HL-Docs comments for this:

/// HL-Docs: ref:Bugfixes; issue:1082
/// Your comment here

Since the entire code change is just one line, Start and End tags are not necessary.

Iridar (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
Tedster59: Updated

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
Tedster59: Updated

PR #1212, opened 2023-06-26
Implement Better Positive Trait UI

### LINKED ISSUE #1081: No "gained positive trait" UI
#85 enabled the positive traits functionality, but the UI still reports them as negative:

image

image

BlackDog86:
Fixes #1081 

Summary:

1. Modified UIAlert to use the "TrainingComplete" alert for positive traits - This has the same set of UI elements as Alert_NegativeSoldierEvent for negative traits but with a neutral colour scheme & promotion-icon.

2. Updated XcomHQPresentationLayer to check whether traits are positive or negative before calling the notifybanner function (this is the piece which places the trait notifications on the bottom left of the screen). Adjusted colours and icons to make more sense.

3. Adjusted the UIAvengerHUD OnClickTraitIcon function so that if someone clicks on a soldier walking around the avenger with a negative or positive trait icon (like the bond icon), it takes them to the correct UI Panel for that trait. 

Note that I can't confirm this new code actually does anything - I can't remember seeing anyone walking around with trait icons bobbing around so probably the function is never called for, but anyway, the code now includes positive traits as well in case that behaviour is ever fixed in the future.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAlert.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAlert.uc
@@ -1242,8 +1242,13 @@ simulated function Name GetLibraryID()
 	case 'eAlert_RescueSoldier':					return 'XPACK_Alert_MissionSplash';
 
 	case 'eAlert_NegativeTraitAcquired':			return 'Alert_NegativeSoldierEvent';
-	case 'eAlert_PositiveTraitAcquired':			return 'Alert_AssignStaff';
+	/// HL-Docs: feature:PositiveTraitUI; issue:1081; tags:strategy,ui
+	/// Positive traits are a scrapped WOTC feature. The game had only negative traits on release, 
+  /// and no UI was created for a "positive trait acquired" event, even though X2TraitTemplate has an unused `bPositiveTrait` flag. 
+  /// This feature copies and adjusts the "soldier promoted" UI to make it function as a makeshift "positive trait acquired" UI. 
+  /// This popup will be displayed whenever a soldier acquires a trait with the `bPositiveTrait` flag set to `true` on its X2TraitTemplate.
 
+	case 'eAlert_PositiveTraitAcquired':			return 'Alert_TrainingComplete';
 	case 'eAlert_InspiredResearchAvailable':		return 'Alert_ItemAvailable';
 	case 'eAlert_BreakthroughResearchAvailable':	return 'Alert_ItemAvailable';
 	case 'eAlert_BreakthroughResearchComplete':		return 'Alert_ItemAvailable';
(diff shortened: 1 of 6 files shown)


RustyDios:
A quick glance, and I fully agree this looks like it fixes a lot of issues there are various places that have for example;
```
-var() bool							bHasSeenNegativeTraitPopup;
//Issue# 1081 - Alter bool name so it applies to both positive and negative traits
+var() bool							bHasSeenTraitPopup;
```

Which would break the CHL backwards compatibility policy on any mods previously reliant on the removed bool.
Rather than remove the word 'Negative' from everything, a better approach should be to duplicate the variables adding 'Positive' and make the code account for both.

robojumper:
`XComGameState_HeadquartersXCom` is a native class, so property additions are disallowed. Also in general, keep property names and function names and arguments the same -- even if they have "negativeTrait" in their name, simply keep that and add a comment that this does more than what the name implies.

BlackDog86:
> A quick glance, and I fully agree this looks like it fixes a lot of issues there are various places that have for example;
> 
> ```
> -var() bool							bHasSeenNegativeTraitPopup;
> //Issue# 1081 - Alter bool name so it applies to both positive and negative traits
> +var() bool							bHasSeenTraitPopup;
> ```
> 
> Which would break the CHL backwards compatibility policy on any mods previously reliant on the removed bool. Rather than remove the word 'Negative' from everything, a better approach should be to duplicate the variables adding 'Positive' and make the code account for both.

BC did cross my mind when I changed that - I don't think it matters to just change it back and comment that the function does 'more than what it says on the tin' as it were

BlackDog86:
Requested changes now implemented (no changes to XComGameState_HeadquartersXCom and function names kept the same for BC)

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAlert.uc
```diff
@@ -4062,10 +4064,15 @@ simulated function BuildSuperSoldierAlert()
 	LibraryPanel.MC.FunctionString("UpdateImage", class'X2StrategyGameRulesetDataStructures'.static.GetDynamicStringProperty(DisplayPropertySet, 'StaffPicture'));
 }
 
-simulated function BuildNegativeTraitAcquiredAlert()
+/// HL-Docs: bugfix:FixPositiveTraitUI; Issue:1081;
+/// Removed duplicate 'positive traits' function - now handle all traits in the same one with a bool to indicate whether trait is positive or negative
+/// Since Alert_NegativeSoldierEvent is a subclass of Alert_Trainingcomplete in flash, they take the same arguments
+/// This allows passing details to the panels using the same .uc code (but generating different outputs)
+/// All references to 'negative' traits in the function have been removed but the function name is preserved for BC
+simulated function BuildNegativeTraitAcquiredAlert(bool bPositiveTrait)
```
Iridar, line 4072: Without looking at anything else, you can't change function definitions of existing functions, that breaks the BC.

The only exception are private/final internal Highlander functions that are explicitly not covered by the BC policy.

Iridar (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAlert.uc
BlackDog86: Understood - I've now reworked the code so that all base game functions are left unmodified.

BlackDog86:
Are there any more changes required on this or is it just a timing thing?

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAlert.uc
```diff
@@ -4146,37 +4146,93 @@ simulated function BuildNegativeTraitAcquiredAlert()
...
+		ForceLocation.X += 50;
+		ForceLocation.Y -= 300;
+		ForceRotation.Yaw = 16384;
+		`HQPRES.CAMLookAtRoom(RoomState, `HQINTERPTIME, ForceLocation, ForceRotation);
+	}
 
 	// Send over to flash
-	LibraryPanel.MC.BeginFunctionOp("UpdateData");
-	LibraryPanel.MC.QueueString(m_strSoldierShakenHeader); //ATTENTION
+	LibraryPanel.MC.BeginFunctionOp("UpdateData");1
```
Iridar, line 4214: What is this stray `1` at the end of the line? Would this even compile?

Iridar (commented):
To be clear, this feature adds just a UI, right? But no way for units to actually acquire positive traits? Docs need to specify how mod authors should take advantage of this feature.

BlackDog86:
> To be clear, this feature adds just a UI, right? But no way for units to actually acquire positive traits? Docs need to specify how mod authors should take advantage of this feature.
> 
> And I'd love to see some screenshots of this UI in action from an example use case.

There are pictures of the updated UI in Issue #1081 - not sure how that 1 got in there either - it's probably a stray key-press from when I was fixing commits that had gotten a long way behind the master earlier this month.

BlackDog86:
Additionally, the base game only allows for one trait popup message to appear at a time. Personally I'd prefer if each trait brought its own popup message when its triggered but I didn't think that that was enough of a justification to comment out the block below. Also, if you use the 'starting traits' mod, it does mean it could bring up quite a few popups when you start a new campaign but I'm open to discussion on leaving the base game code as it is, or adjusting it.

```
> if(PopupCount > 0 && !XComHQ.bHasSeenNegativeTraitPopup)
> 				{
> 					XComHQ.bHasSeenNegativeTraitPopup = true;
> 					UINegativeTraitAlert(NewGameState, PopupUnit, StoredTraitName);
> 				}
```

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAvengerHud_SoldierStatusIndicatorContainer.uc
```diff
@@ -270,31 +270,31 @@ function UITraitIcon CreateTraitIcon(XComGameState_Unit Unit, bool bPositive)
 
 function OnClickTraitIcon(UITraitIcon Icon)
 {
-	local int UntiID;
```
Iridar, line 273: I appreciate the diligence, but fixing typos in vanilla code is outside the scope of this PR.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
```diff
@@ -1727,14 +1738,30 @@ function UISoldierBondConfirm(XComGameState_Unit UnitRef1, XComGameState_Unit Un
...
+
+	EventTemplateManager = class'X2EventListenerTemplateManager'.static.GetEventListenerTemplateManager();
+	TraitTemplate = X2TraitTemplate(EventTemplateManager.FindEventListenerTemplate(TraitTemplateName));
 
 	UnitState.AlertTraits.RemoveItem(TraitTemplateName);
 	UnitState.WorldMessageTraits.RemoveItem(TraitTemplateName);
 
-	BuildUIAlert(PropertySet, 'eAlert_NegativeTraitAcquired', None, 'OnNegativeTraitAcquired', "Geoscape_CrewMemberLevelledUp", false);
+	//Begin issue #1081
+	if (TraitTemplate.bPositiveTrait)
```
Iridar, line 1756: Add a none-check for the trait template pls.

Iridar (changes requested):
Looks mostly okay to me, but I have issues with it being a bit messy, like fixing typos or adding random empty spaces to lines.

I'm also nitpicking to docs. 

The point of HL-Docs in this case is to be mod author facing, so they can get a clean and concise explanation of what this CHL feature can do for them.

Something along the lines of:

_"Positive traits are a scrapped WOTC feature, on release the game had only negative traits, so there is no UI for "positive trait acquired", but X2TraitTemplate does have a `bPositiveTrait` flag. This feature repurposes "soldier promoted" UI to make it function as a makeshift "positive trait acquired UI". To make the game display it when the soldier acquires a positive trait, simply set bPositiveTrait = true; on your X2TraitTemplate."_

This all can be done as one piece of Docs somewhere, and the rest of the changes can be marked with standard non-Docs Start/End Issue tags.

BlackDog86:
OK hopefully this is a bit better.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAlert.uc
```diff
@@ -4146,38 +4149,94 @@ simulated function BuildNegativeTraitAcquiredAlert()
 	Button1.DisableNavigation();
 }
 
+// Begin Issue# 1081
```
Iridar, line 4152: Misplaced space, also while "begin" is used in some CHL comments, the convention is "Start".

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAlert.uc
```diff
@@ -1242,8 +1242,11 @@ simulated function Name GetLibraryID()
 	case 'eAlert_RescueSoldier':					return 'XPACK_Alert_MissionSplash';
 
 	case 'eAlert_NegativeTraitAcquired':			return 'Alert_NegativeSoldierEvent';
-	case 'eAlert_PositiveTraitAcquired':			return 'Alert_AssignStaff';
+	/// HL-Docs: feature:PositiveTraitUI; issue:1081; tags:strategy,ui
+	/// Positive traits are a scrapped WOTC feature. On release, the game was configured with only negative traits, so no UI was created for "positive trait acquired".
+	/// However, X2TraitTemplate does have a bPositiveTrait flag. This feature repurposes the "soldier promoted" UI to make it function as a makeshift
```
Iridar, line 1247: Feels like something is missing after the word "makeshift"?

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIAvengerHud_SoldierStatusIndicatorContainer.uc
```diff
@@ -287,14 +287,13 @@ function OnClickTraitIcon(UITraitIcon Icon)
 	EventTemplateManager = class'X2EventListenerTemplateManager'.static.GetEventListenerTemplateManager();
 	TraitTemplate = X2TraitTemplate(EventTemplateManager.FindEventListenerTemplate(Unit.AlertTraits[0]));
 	//while it's an array, we go with the base game's assumption of having only one trait to worry about when it changes				
-			
-	if(TraitTemplate.bPositiveTrait)
-		return; 
 	//end issue #85
 	
+	/// Issue #1081: Removed code which early exits the function if the user clicks on a positive trait
```
Iridar, line 292: In this case it would be preferable to comment out the original code rather than removing it.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersXCom.uc
```diff
@@ -162,6 +162,7 @@ var() bool							bHasSeenCovertActionRiskIntroPopup;
 var() bool							bHasReceivedResistanceOrderPopup;
 var() bool							bHasSeenCantChangeOrdersPopup;
 var() bool							bHasSeenSoldierBondPopup;
+// Single line for Issue #1081 - Note, this bool now also covers positive traits
```
Iridar, line 165: Since there's no actual changes here, I suggest making it an inline comment instead.

`var() bool bHasSeenNegativeTraitPopup; // Issue #1081 - this bool now also covers positive traits`

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
```diff
@@ -457,29 +457,38 @@ private function PostMissionNegativeTraitNotifies()
 		foreach UnitState.WorldMessageTraits(TraitName)
 		{
 			TraitTemplate = X2TraitTemplate(EventTemplateManager.FindEventListenerTemplate(TraitName));
-
+			
+			// Begin Issue# 1081
```
Iridar, line 461: Misplaced space, Begin -> Start

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
```diff
@@ -457,29 +457,38 @@ private function PostMissionNegativeTraitNotifies()
...
+						StoredTraitName = TraitName;
+					}
+			}
+			// Bring popupcount inside the foreach (unitstate)
+			if(PopupCount > 0 && !XComHQ.bHasSeenNegativeTraitPopup)
 				{
-					PopupCount++;
-					PopupUnit = UnitState;
-					StoredTraitName = TraitName;
+					XComHQ.bHasSeenNegativeTraitPopup = true;
```
Iridar, line 484: I'm a bit unclear, what is the purpose of this bool? Does this mean the player can see only one "negative trait acquired" popup per campaign? And what's the reasoning in making displaying a positive trait apply to it as well?

Iridar (commented):
Better, but needs work still.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
BlackDog86: OK so the way this bit works is, if multiple soldiers happen to get traits on a mission (or at the start of the game, if using the 'starting traits' mod), the UI displays the popup message like this only for the first one:
137698942-b2587684-60ad-4c30-bb3a-4063cc3d2c39
Then, any of the rest are just displayed on the LHS sidebar:
137698952-4b71c90b-b167-46ec-8e09-3a3ea674dffd

Personally I think the experience would be better if that bool was never set so that the trait popups display sequentially for each trait gained, rather than just once but it's a bit of a subjective change - for example, if a mod was made that gave traits to every soldier at the start of the campaign, then you could conceivably get 20 popups or something. We could set a limit of 6, which would allow displaying traits for every member of the squad on a mission but would prevent general UI spam in the event that something adds lots of traits at once.

Iridar: Yeah I have no issues with that approach, I'm not sure why Firaxis even put that functionality in place. Though conventional CHL practice would be to gate the change behind a config bool flag, and keep things as they were by default.

Iridar: Sorry, yes, push the new version with "allow multiple trait acquired popups" bool, defaulted to false.

BlackDog86:
This should do the trick :)

BlackDog86:
The comments look fine - I'll change the name of the bool value tonight. Thanks for the review.

PR #1214, opened 2023-06-29
Make Return Fire visualize after Covering Fire Overwatch

### LINKED ISSUE #1196: Return Fire and Covering Fire fight for visualization
If a soldier on Covering Fire Overwatch also has Return Fire, and an enemy shoots that soldier, then the soldier will respond with both Overwatch Shot and Return Fire Shot at the same time in short order in a sort of glitched animation.

Adding `Template.AssociatedPlayTiming = SPT_AfterSequential;` to the Return Fire Shot seems to fix this problem, but may or may not cause negative repercussions, keeping this change on ice for now until I do some playtesting myself.

Iridar:
Fixes #1196 

Test before merging.


Iridar:
Tested by having a bunch of soldiers with Covering Fire and Return Fire overwatch camp a bunch of enemies for quite a bit of time. Worked fine, the only weird interaction was the soldier return fire on a sectoid *after* being panicked by mindspin. 

Without this change it visualizes almost the same, so it's not strictly relevant to this issue.

PR #1219, opened 2023-07-04
Allow mods to override cost strings (#1218)

### LINKED ISSUE #1218: Allow mods to override cost strings
Currently, the only way for mods to change cost strings generated by the utility function [`UIUtilities_Strategy.GetStrategyCostString`](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/1dbcd73fcc5033c2d63dc2be526a206b8a44d576/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUtilities_Strategy.uc#L145) is either by an MCO (automatically covering all places where cost strings are used) or by implementing a ScreenListener that handles all known screens where cost strings are used.

By adding an event-based tuple override to the Highlander, a simple way to override the cost strings would be provided for mods without them having to rely on MCO'ing `UIUtilities_Strategy`.

fearlessiron:
Adds the event `OverrideStrategyCostString` that allows mods to override the cost strings created by the utility function
`UIUtilities_Strategy.GetStrategyCostString` which is used throughout XCOM 2 to create the strings specifying the artifacts and resources costs for items, facilities, covert actions, etc.

The event takes the form:
```event
EventID: OverrideStrategyCostString,
EventData: [
	in name ItemTemplateName,
	in int Quantity,
	in bool IsResourceCost,
	inout string CostString
],
EventSource: none,
NewGameState: none
```

The event is triggered individually for every component of the `StrategyCost` value. Both the name of the item template of the cost component and its are passed to the event. Futhermore, the information whether it constitutes a resource cost or an artifact cost is passed via `IsResourceCost`.

Resolves #1218

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUtilities_Strategy.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIUtilities_Strategy.uc
@@ -142,6 +142,44 @@ static function XComGameState_BlackMarket GetBlackMarket()
 	return BlackMarketState;
 }
 
+// Start Issue #1218
+/// HL-Docs: feature:OverrideStrategyCostString; issue:1218; tags:strategy,ui
+/// Fires an event that allows mods to override the strategy layer cost strings.
+/// Default: Vanilla behavior will be used.
+///
+/// ```event
+/// EventID: OverrideStrategyCostString,
+/// EventData: [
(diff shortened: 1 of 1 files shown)


Iridar:
Looks very good on the first glance, I'll do a proper review and some testing at a later date.

Iridar:
Tested, seems to work just fine. Good work.

PR #1220, opened 2023-07-10
Chosen Death Teleport Visualization Fix

### LINKED ISSUE #496: Chosen death teleport effect times out
image

Happens to me always when running with redscreens

BlackDog86:
Fixes #496 - Tested & confirmed working with minimal code changes.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_Chosen.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_Chosen.uc
@@ -1,0 +1,2895 @@
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+class X2Ability_Chosen extends X2Ability
+	config(GameData_SoldierSkills);
+
+var localized string HoloTargetEffectName;
+var localized string HoloTargetEffectDesc;
+
+var localized string KidnapHeader;
+var localized string KidnapTargetHeader;
+var localized string KidnapMessageBody;
+var localized string ExtractHeader;
+var localized string ExtractTargetHeader;
+var localized string ExtractMessageBody;
+var localized string DefeatedHeader;
+var localized string DefeatedTargetHeader;
+var localized string DefeatedMessageBody;
+
+var config float SOULSTEALER_HP_MOD;
+var config int AGILE_CHANCE;
+var config int ACHILLES_TO_HIT;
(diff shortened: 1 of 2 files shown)


Iridar:
What's the status here? Is this ready for review?

BlackDog86:
Yep, all good - I've tested it a few ways with different abilities and it seems to work just fine :)

Iridar:
Uhh why is X2Ability_CarryUnit in this PR?

Iridar:
Good thing I decided to test this before merging, because current implementation is causing bugged visualization.

https://youtu.be/va8IXk38H7Y

The Chosen plays a part of the standard rifle shot animation before teleporting away. Also it seems to be overlaid on top of the chosen death animation, because it horribly distorts his chest and shoulders. 

I don't know why exactly does this happen, but implementing a different solution should bypass the problem.

Iridar:
Since the original issue happens because the particle effect `"FX_Chosen_Teleport.P_Chosen_Teleport_Out_w_Sound"` never finishes playing, presumably because that's just how it is set up, there are two possible solutions here:

1. Create an alternative effect that *will* finish playing, allowing visualization to proceed as intended.
2. Create an `X2Action_TimedWait` that runs in parallel to the `X2Action_PlayEffect`, and make actions that previously waited on `X2Action_PlayEffect` wait on `X2Action_TimedWait` instead. Give it ~5 seconds or so.

Iridar:
On deeper look, it seems that build visualization function is just a horrible bugged mess.

Iridar:
So the firing animation was played by the build visualization function itself for no adequate reason. Once I got rid of that and tweaked the build vis in a few places, it's crispy smooth:

https://youtu.be/kv-MRWrBGcQ

Iridar:
I ended up refactoring most of the build visualization function for that ability. The original one was horribly defunct, and the only reason it was working at all is because the thing that was breaking it was broken as well. 

It appears it tried to play the Chosen weapon firing animation at the same time as the death animation was playing, and treated the animation action as if it *was* the death animation.

At least that's the only way I can make sense of it at all. 

Either way, I removed the animation action and added a couple delays to make sure it visualizes in a decent manner. 

According to testing by Black Dog, the actual delay tends to vary depending on how exactly the Chosen was killed, e.g. kills by melee or environmental explosive will have different timings.

Ideally I would want to have an X2Action that waits for the pawn to become fully idle before the teleport visualization can start, regardless of the death animation used, but there's no existing methodology for that, and would require bigger scope of Highlander changes that I'm comfortable with.

So I'll leave this as is for now, revisit it later if there are any complaints.

PR #1224, opened 2023-07-25
Add Exclusion for Burning Status on PutDownUnit()

### LINKED ISSUE #1223: Can't put down carried unit while on fire
>  being on fire stops you from setting down a carried unit.  And since carrying a unit means you can't hunker, you are stuck with no options other than movement or evac -- hoping you have enough HP to not burn up in the meanwhile (or can get a heal from someone else).

BlackDog86:
Fixes #1223

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_CarryUnit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_CarryUnit.uc
@@ -1,0 +1,410 @@
+class X2Ability_CarryUnit extends X2Ability
+	config(GameCore);
+
+var config float CARRY_UNIT_RANGE;
+var config int CARRY_UNIT_MOBILITY_ADJUST;
+var localized string CarryUnitEffectFriendlyName;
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_CarryUnit.uc
```diff
@@ -301,6 +301,9 @@ static function X2DataTemplate PutDownUnit()
 	// Special effects that get removed when being put down
 
 	SkipExclusions.AddItem(default.CarryUnitEffectName);
+	/// HL-Docs: ref:BugFixes; issue:1223
```
Iridar, line 304: Docs builder isn't happy with something, try `Bugfixes` instead of `BugFixes`.

BlackDog86: yep, that did it

Iridar:
Looks good. Did you gameplay test this?

BlackDog86:
> 

Yep, it works correctly. You can't pick up units while burning but you can definitely put them down :)

Issue #324, opened 2018-11-04
Errors reported in launch.log


Dragon-32:
Hello.

I ran a couple of missions and then checked my launch.log. There're a number of entries which appear to relate to the Highlander (did a grep). I have no idea if these are important or not. No obvious errors appeared in game.

```
[1186.43] Error: Redscreen: DisconnectAction called with a NULL ActionToDisconnect! Stack:
Script call stack:
	(XComGameState_AIGroup_84) XComGameState_AIGroup::MergeGroupMoveVisualization
	(XComGameStateContext_Ability_380) XComGameStateContext_Ability::MergeIntoVisualizationTree
	(XComGameStateVisualizationMgr_2) XComGameStateVisualizationMgr::BuildVisualizationFrame
	(XComGameStateVisualizationMgr_2) XComGameStateVisualizationMgr::EvaluateVisualizationState
	(XComGameStateVisualizationMgr_2) Idle::Tick
[1186.43] Warning: Redscreen: ConnectAction called with a NULL action to place.
Script callstack:
 Script call stack:
	(XComGameState_AIGroup_84) XComGameState_AIGroup::MergeGroupMoveVisualization
	(XComGameStateContext_Ability_380) XComGameStateContext_Ability::MergeIntoVisualizationTree
	(XComGameStateVisualizationMgr_2) XComGameStateVisualizationMgr::BuildVisualizationFrame
	(XComGameStateVisualizationMgr_2) XComGameStateVisualizationMgr::EvaluateVisualizationState
	(XComGameStateVisualizationMgr_2) Idle::Tick
[1186.44] Error: Redscreen: DisconnectAction called with a NULL ActionToDisconnect! Stack:
Script call stack:
	(XComGameState_AIGroup_84) XComGameState_AIGroup::MergeGroupMoveVisualization
	(XComGameStateContext_Ability_381) XComGameStateContext_Ability::MergeIntoVisualizationTree
	(XComGameStateVisualizationMgr_2) XComGameStateVisualizationMgr::BuildVisualizationFrame
	(XComGameStateVisualizationMgr_2) XComGameStateVisualizationMgr::EvaluateVisualizationState
	(XComGameStateVisualizationMgr_2) Idle::Tick
[1186.44] Warning: Redscreen: ConnectAction called with a NULL action to place.
Script callstack:
 Script call stack:
	(XComGameState_AIGroup_84) XComGameState_AIGroup::MergeGroupMoveVisualization
[...]

PR #1226, opened 2023-07-26
Fix a few log warnings

Iridar:
Tracking Issue #324


Iridar:
Note that `ValidateTierXHackRewards()` functions returning true even if the hack reward template is 'none' may or may not be leading to buggy behavior down the line, but fixing it is outside the scope of Issue #324. Something to keep an eye out in the future.

PR #1234, opened 2023-08-05
Cache hit chances and game state for sorting enemies in UITacticalHUD_Enemies

### LINKED ISSUE #1233: UITacticalHUD_Enemies UpdateVisibleEnemies should use cached hit chances when sorting final targets array
UpdateVisibleEnemies of UITacticalHUD_Enemies sorts the array of visible enemies by hit chance using a comparer delegate.

This sorting delegate determines the hit change, by calling GetHitChanceForObjectRef on object A and B, which is relatively slow, and calls it multiple times for the same objects, until the array is fully sorted.

The hit chance for the same object reference cannot change during one sorting pass, so pre-calculating the hit chances for visible enemies first. Then using that in the sorting delegate would decrease the number of calls drastically.

Before:
image

After:
image

Note that I used the override of UITacticalHUD_Enemies from 'WOTC_DisplayHitChange' to do my test, which was easier for now than to build a custom CHL.

The profiling was done on the exact same save, same conditions, same move. Reduced the number of calls in that one frame, from 378 to 48.

```
struct StateObjectReferenceWrapper
{
	var StateObjectReference Object;
	var int HitChance;
};

simulated function UpdateVisibleEnemies(int HistoryIndex)
{
	local array<StateObjectReferenceWrapper> CachedHitChanges;
	local StateObjectReferenceWrapper HitChangeWrapper;

	// ... snip

	if (kActiveUnit != none)
	{
		// ... snip

		iNumVisibleEnemies = m_arrTargets.Length;

		CachedHitChanges.Length = 0;
		for (i = 0; i < m_arrTargets.Length; ++i)
		{
			HitChangeWrapper.Object = m_arrTargets[i];
			HitChangeWrapper.HitChance = GetHitChanceForObjectRef(HitChangeWrapper.Object);
			CachedHitChanges.AddItem(HitChangeWrapper);
		}

		CachedHitChanges.Sort(SortEnemies2);

		m_arrTargets.Length = 0;
		for (i = 0; i < CachedHitChanges.Length; ++i)
		{
			m_arrTargets.AddItem(CachedHitChanges[i].Object);
		}

		UpdateVisuals(HistoryIndex);
	}
}

private function int SortEnemies2(StateObjectReferenceWrapper ObjectA, StateObjectReferenceWrapper ObjectB)
{
    // you get the point :)
}	
```

Not sure if the above implementation is ideal, that's the best I could do with my limited UnrealScript knowledge.
Maybe it's also possible with a class variable, then using CachedHitChanges.Find('Object', ObjectA) (which is a hashtable lookup?) inside SortEnemies.

discord: https://discord.com/channels/165245941664710656/165245941664710656/1137333029837209630

remcoros:
Full implementation for what I mentioned in the issue.

Not sure about the name for 'StateObjectReferenceHitChange'

fixes #1233

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Enemies.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Enemies.uc
@@ -478,20 +478,15 @@ function RefreshShine(optional bool bMoveAbovePercent = false)
 	}
 }
 
-// Start Issue #1233 if inheritors of this class still call this function, use a slightly less efficient, but still faster version
 simulated function int SortEnemies(StateObjectReference ObjectA, StateObjectReference ObjectB)
 {
 	local XComGameState_Destructible DestructibleTargetA, DestructibleTargetB;
 	local XComGameStateHistory History;
 	local int HitChanceA, HitChanceB;
-	local StateObjectReferenceHitChange ObjectAHitChange;
-	local StateObjectReferenceHitChange ObjectBHitChange;
 
-	ObjectAHitChange = m_arrTargetsUnsorted[m_arrTargetsUnsorted.Find('Object', ObjectA)];
-	ObjectBHitChange = m_arrTargetsUnsorted[m_arrTargetsUnsorted.Find('Object', ObjectB)];
-
-	DestructibleTargetA = XComGameState_Destructible(ObjectAHitChange.GameState);
-	DestructibleTargetB = XComGameState_Destructible(ObjectBHitChange.GameState);
+	History = `XCOMHISTORY; 
+	DestructibleTargetA = XComGameState_Destructible(History.GetGameStateForObjectID(ObjectA.ObjectID));
+	DestructibleTargetB = XComGameState_Destructible(History.GetGameStateForObjectID(ObjectB.ObjectID));
 
 	//Push the destructible enemies to the back of the list.
 	if( DestructibleTargetA != none && DestructibleTargetB == none ) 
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Enemies.uc
```diff
@@ -441,15 +478,47 @@ function RefreshShine(optional bool bMoveAbovePercent = false)
 	}
 }
 
+// Start Issue #1233 if inheritors of this class still call this function, use a slightly less efficient, but still faster version
```
robojumper, line 481: So this assumes that `UpdateVisibleEnemies` is not overridden by that subclass, otherwise `m_arrTargetsUnsorted` will not be initialized with the targets to be sorted, and I think it's very likely that subclasses that call `SortTargets` themselves will override `UpdateVisibleEnemies` since that's the original caller location...

remcoros: good point!

what to do here, just not touch that function and let any overrides deal with it? (I know only one mod that actually overrides _Enemies: Extended Information, and it does not override those functions.

robojumper: I admit I haven't quite kept up with the story around these mods, but the copy of Perfect Information I have does that. I think the best approach would be extracting the enhanced sort code to its own function and leaving the existing comparison delegate untouched -- mods can either fix the performance issue themselves or call our sort function conditionally depending on whether the CHL is installed.

Iridar (commented):
General logic seems fine to me, but I have issues with names of the new struct and function - they need the CH suffix. And I'd like to test it a bit. Since it took me forever to review this, I think the author is moved on by this point, so I'll probably resubmit this PR with some minor changes myself a bit later.

remcoros:
> General logic seems fine to me, but I have issues with names of the new struct and function - they need the CH suffix. And I'd like to test it a bit. Since it took me forever to review this, I think the author is moved on by this point, so I'll probably resubmit this PR with some minor changes myself a bit later.

I've moved on (for now), but can still support and/or answer questions to the best of my knowledge, also about the other PR (I see that was merged). I don't have a working dev environment for UE3 at the moment, so any changes have to be done/tested by someone else, but happy to reply to any questions here or hit me up on discord.

Iridar:
That's good to hear and understandable.

Iridar:
Closing in favor of #1387 (my alternative implementation with additional features)

Issue #1108, opened 2021-12-16
Unequipping BIT / GREMLIN should remove the cosmetic unit


Iridar:
This bug doesn't occur in vanilla, because GREMLINs and BITs are hard set to be the only secondary weapons available to their respective classes, but many mods make it possible to equip and unequip them at will, which does not immediately nuke the cosmetic unit, so the BIT / GREMLIN will still hang around the loadout screen, despite the unit not having one equipped already.

Since this is a fairly universal bug, it should probably be fixed in Highlander rather than incorporated individually into every mod that makes it possible to equip something other than a BIT / GREMLIN.

PR #1244, opened 2023-08-17
Destroy cosmetic pawns when items are unequipped 

Tedster59:
Fixes Issue #1108 by creating a new function `DestroyCosmeticPawn_CH` that is modeled on the existing `GetCosmeticPawn` function, but destroys the requested cosmetic pawn instead of returning it.  This function is now called by `RemoveItemFromInventory` on `XComGameState_Unit`

New PR to use proper rebased branch with cleaned up commits.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIPawnMgr.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIPawnMgr.uc
@@ -259,6 +259,54 @@ simulated function XComUnitPawn GetCosmeticPawnInternal(out array<PawnInfo> Pawn
 	return PawnStore[StoreIdx].Cosmetics[CosmeticSlot].Pawn;
 }
 
+/// Start issue #1108 - new function to destroy cosmetic pawns.  INTERNAL CHL API, not covered by BC policy.
+simulated final function bool DestroyCosmeticPawn_CH(int CosmeticSlot, int UnitRef, bool bUsePhotoboothPawns = false)
+{
+	local int PawnInfoIndex;
+
+	if (bUsePhotoboothPawns)
+	{
+		PawnInfoIndex = PhotoboothPawns.Find('PawnRef', UnitRef);
+		if (PawnInfoIndex != -1)
+		{
+			return DestroyCosmeticPawnInternal_CH(PhotoboothPawns, CosmeticSlot, PawnInfoIndex);
+		}
+	}
+	else
+	{
+		PawnInfoIndex = Pawns.Find('PawnRef', UnitRef);
+		if (PawnInfoIndex != -1)
+		{
+			return DestroyCosmeticPawnInternal_CH(Pawns, CosmeticSlot, PawnInfoIndex);
+		}
+
+		PawnInfoIndex = CinematicPawns.Find('PawnRef', UnitRef);
+		if (PawnInfoIndex != -1)
+		{
+			return DestroyCosmeticPawnInternal_CH(CinematicPawns, CosmeticSlot, PawnInfoIndex);
+		}
+	}
+
+	return false;
+}
+
+// Helper function used by DestroyCosmeticPawn_CH. Also internal only, and not covered by BC policy.
+simulated final function bool DestroyCosmeticPawnInternal_CH(out array<PawnInfo> PawnStore, int CosmeticSlot, int StoreIdx)
+{
+	local bool bSuccess;
+
+	if (CosmeticSlot >= PawnStore[StoreIdx].Cosmetics.Length)
+		return false;
+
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UIPawnMgr.uc
```diff
@@ -259,7 +259,8 @@ simulated function XComUnitPawn GetCosmeticPawnInternal(out array<PawnInfo> Pawn
 	return PawnStore[StoreIdx].Cosmetics[CosmeticSlot].Pawn;
 }
 
-/// Start issue #1108 - new function to destroy cosmetic pawns.  INTERNAL CHL API, not covered by BC policy.
+// Start issue #1108
+// Destroy a cosmetic pawn associated with a particular inventory slot on a particular unit, if there is one.
```
Iridar, line 263: It's preferable to keep comments to the point, just explaining what the function does, giving its history unless it's relevant is unnecessary. 

I for one got mislead by "use the new function", which can be interpreted as there's an old version of `DestroyCosmeticPawn()`.

I also decided to remove the mention of "internal CHL API", because that would imply we don't want mods using this new function. I don't see any harm in keeping it open, though I confess I can't come up with a potential use case on the spot.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIPawnMgr.uc
```diff
@@ -290,8 +291,7 @@ simulated final function bool DestroyCosmeticPawn_CH(int CosmeticSlot, int UnitR
 	return false;
 }
 
-// Helper function used by DestroyCosmeticPawn_CH. Also internal only, and not covered by BC policy.
-simulated final function bool DestroyCosmeticPawnInternal_CH(out array<PawnInfo> PawnStore, int CosmeticSlot, int StoreIdx)
+simulated private function bool DestroyCosmeticPawnInternal_CH(out array<PawnInfo> PawnStore, int CosmeticSlot, int StoreIdx)
```
Iridar, line 294: Now that's an actual internal CHL API function, but in this case just adding `private` is enough to imply that, per Xym's guidelines.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -8485,9 +8485,6 @@ simulated function bool RemoveItemFromInventory(XComGameState_Item Item, optiona
 	local X2ArmorTemplate ArmorTemplate;
 	local int RemoveIndex;
 
-	// Variable for Issue #1108
-	local UIPawnMgr PawnMgr;
```
Iridar: Since you're not gonna be none-checking the UIPawnMgr - and the user has bigger problems then accessed-none log warnings if UIPawnManager is missing - I decided to get rid of the local var for code brevity.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -8512,15 +8509,14 @@ simulated function bool RemoveItemFromInventory(XComGameState_Item Item, optiona
 			{
 				`RedScreen("Attempt to remove item" @ Item.GetMyTemplateName() @ "properly may have failed due to OnUnequippedFn -jbouscher @gameplay");
 			}
-		}		
+		}	
+		
+		/// HL-Docs: ref:Bugfixes; issue:1108
```
Iridar, line 8514: In general we do want to ship most submissions with a HL-Docs entry, one was missing from this PR.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
```diff
@@ -8512,15 +8509,14 @@ simulated function bool RemoveItemFromInventory(XComGameState_Item Item, optiona
...
+		/// If there is a cosmetic pawn associated with the unequipped item item, remove it.
+		`PRESBASE.GetUIPawnMgr().DestroyCosmeticPawn_CH(Item.InventorySlot, self.ObjectID);
```
Iridar, line 8516: As I understand it, there are two main presentation layers, HQPRES for strategy and PRES for tactical. Both of them extend PRESBASE. Using PRESBASE macro will yield you either HQPRES or PRES depending on where you are. 

Since `GetUIPawnMgr()` lives in PRESBASE, and it is technically possible to unequip items in tactical, I think it's more appropriate to use PRESBASE here.

Iridar (commented):
@Tedster59 I've made some adjustments to the PR, see the comments for details. I tested and it still works fine. Please take a look to see if you agree with my changes, and if all is good, I'll squash merge this.

Iridar:
Approved by Tedster, merging.

PR #1247, opened 2023-08-19
Introduce CHDLCHookManager to retrieve DLCInfos with specific implemented hooks.

### LINKED ISSUE #212: Improve performance on DLC Hooks (X2DownloadableContentInfo)
Adding a DLC Hook to the [X2DownloadableContentInfo](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2DownloadableContentInfo.uc) class is a simple way to expose hooks to Mods. This is an example call:

https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/47a00149c83d86b9826da64866313192cb41b18c/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHumanPawn.uc#L967-L971

The problem is that performance becomes worse with every mod installed, since almost all mods include an X2DownloadableContentInfo file. If we limited the function calls to only those classes that actually implement the function, we could improve performance dramatically and make DLC hooks more viable in code segments that are called often. We would incur a small overhead for filtering the classes in the first place, but if we cache the classes, this should improve performance quite a bit.

Whether a class implements a function can be checked by 

    `CONTENT.RequestGameArchetype("Package.Class:Function") != none

There is one problem: If users create an "abstract" X2DownloadableContentInfo subclass and let functions in this class call functions that are overridden in subclasses, it will break, because the subclass functions don't really exist. I don't know of any mod/base-game mechanism that does it this way, but a warning would be appropriate. Alternatively, `IsA` checks can be used to automatically include subclasses.

remcoros:
fixes #212

It's still missing proper docs and comments. but this is the idea. I've split up the commits for easier reviewing.


X2WOTCCommunityHighlander/Src/XComGame/Classes/CHDLCHookManager.uc
```diff
@@ -0,0 +1,71 @@
...
+
+			if (FindObject(QualifiedName, class'Function') != none)
+			{
+				DLCInfosWithMethod.AddItem(CurrentDLC);
+				break;
+			}
+
+			CurrentClass = class'XComEngine'.static.GetClassDefaultObject(CurrentClass).ObjectArchetype.Class;
+		}
+		until(CurrentClass == class'X2DownloadableContentInfo' || CurrentClass == class'Object' || CurrentClass == none);
```
robojumper, line 67: Instead of looping with a variable of type `Class` and going up the class hierarchy, you could loop with a variable of type `X2DownloadableContentInfo` and go up the `ObjectArchetype` hierarchy -- that way you can still access `.Class` where you need it, but don't need to go back from class to X2DLCInfo default object.

X2WOTCCommunityHighlander/Src/XComGame/Classes/UIArmory_Loadout.uc
```diff
@@ -841,6 +840,7 @@ simulated function string GetDisabledReason(XComGameState_Item Item, EInventoryS
 	//start of Issue #50: add hook to UI to show disabled reason, if possible
 	//start of Issue #114: added ItemState of what's being looked at for more expansive disabling purposes
 	//issue #127: hook now fires all the time instead of a specific use case scenario
+	DLCInfos = `DLCHOOKMGR.GetDLCInfos('CanAddItemToInventory_CH_Improved');
```
robojumper, line 843: This hook has a bit of a troubled history where the default implementation of `CanAddItemToInventory_CH_Improved` calls previous APIs too, so this needs special attention similarly to `UpdateHumanPawnMeshMaterial`.

robojumper: This wasn't updated by the way -- maybe it'd make more sense to check these dependencies internally in `CHDLCHookManager` so that a single call with `CanAddItemToInventory_CH_Improved` returns all classes that have one of the three functions? That'd also obviate the need for the double/triple loops and the "// don't call when we already called it above" behavior. The multiple loops also currently inadvertently change the [DLC Run Order](https://x2communitycore.github.io/X2WOTCCommunityHighlander/misc/DLCRunOrder/) IIUC.

remcoros: I missed this one, just forced pushed. But it might be better to move this into CHDLCHookManager, yes, I'll think it over tomorrow :)

remcoros: I guess force-pushing doesn't play nice with reviews, as this seems now out of sync. ah well.

remcoros:
I found another issue I am undoing now:

There are a few calls to GetDLCInfos(true) (instead of false), to retrieve only the DLCS that are new in the current campaign. For example in XComOnlineEventMgr, it is used to trigger "OnLoadedSavedGame", but only for DLCs which where not in the curent campaigns histroy.

I will undo that change so that it uses the normal GetDLCInfos(true) again.

I suppose it's not really worth optimizing these cases, since they are rarely/used only once?

remcoros:
Another question: I need some guidance on documenting/commenting on these changes.

- Does CHDLCHookManager need a "HL-Docs: feature:" entry? (Maybe not, because it's only an internal API?)
- Any other places I should add more comments/issue references?

robojumper:
Personally I would go for a documentation page that documents the intent and some implementation details, explicitly stating that:

1. The Highlander does this performance optimization
2. Any observable difference compared to calling the hooks in all DLCInfo classes in run order is a bug.

remcoros:
Added some documentation.

Should I do a rebase --interactive to make the history clean again, or will this get squashed into one commit anyway?

Iridar:
It's gonna get squashed. The only time we use more than one commit per PR is when an existing base game class is added to source.

Iridar:
Also docs fail to build atm in this PR.

X2WOTCCommunityHighlander/Src/XComGame/Classes/CHDLCHookManager.uc
```diff
@@ -0,0 +1,195 @@
+/// HL-Docs: feature:CHDLCHookManager; issue:212; tags:misc
```
robojumper, line 1:
```suggestion
/// HL-Docs: feature:CHDLCHookManager; issue:212; tags:
```

docs workflow is failing because `misc` is not an explicitly assigned tag, `misc` gets applied if there's not exactly one of the {`strategy`, `tactical`} tags.

remcoros: yeah found that after pushing (I now ran make_docs locally, and also fixed a slight markdown formatting issue). I forced push it quickly, but you were faster :D

remcoros:
> It's gonna get squashed. The only time we use more than one commit per PR is when an existing base game class is added to source.

There is one added file. So I'll do a final rebase after this is considered 'done'.

Iridar:
> There is one added file

But this is not base game source code file, it's entirely new, so it doesn't matter.

remcoros:
> > There is one added file
> 
> But this is not base game source code file, it's entirely new, so it doesn't matter.

[XComPresentationLayerBase.uc](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/1247/files#diff-604b57a119e845ce63ed3631c1c023ef92d13ca684d24aa8d6a2cb03a3bd71ed)

this one from base game got added.

Iridar:
Ah, I didn't notice. Then yes.

Iridar:
Implemented in #1385

PR #1248, opened 2023-08-20
skip triggering DrawDebugLabels during game state building

### LINKED ISSUE #1245: Lag spikes caused by triggering events in DrawDebugLabels()
#490 added a `TriggerEvent()` into the **[DrawDebugLabels() function](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/2bd66af629f4a4ba665b69dd2984cafac37027d5/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComTacticalController.uc#L1020C3-L1020C3)**, which causes noticeable lag spikes in tactical under certain conditions.

According to tests done by @remcoros with the profiler, it doesn't matter how many event triggers there are or if any EventData / EventSource are being passed. Event name or if there are any listeners to it does not appear to matter either.

This bug is currently being investigated further, but if no better solution is found, the final option is to gate it behind a config variable, disabled by default. It would be a BC break, but considering it (A) causes noticeable lag and (B) is purely for debug anyway, I feel perfectly fine about.

remcoros:
fixes #1245

As noted in the original issue, this makes it so that there is a slight delay (~ 20 frames in my profiles) during which no DrawDebugLabels events are triggered, and thus, mods using it temporarily don't draw debug labels.

I did a test using LWOTC with 'LWDebugPodJobs' console command toggled on. But the effect is not really noticeable, since it quickly catches up (~300 ms. later, and that is with the profiler ON).

Personally, I don't think we need an additional config switch to fix this and always draw.


remcoros:
Another solution would be to gate the TriggerEvent behind a config switch, and just NEVER call it if it is false. So that when the game is idling, we also do not trigger this event.

Thoughts?

Iridar:
What's the performance cost of triggering the event when it's gated behind ``XCOMGAME.GameRuleset.BuildingLatentGameState` check? If it's at least slightly significant even by a huge stretch, I suggest locking it behind a config bool.

Sidenote, ``GAMERULES.BuildingLatentGameState` should be ever so slightly.

remcoros:
The 'cost' of triggering it (during an idle frame) is near zero (depending on what mods are listening):

image

Checking for a config switch or BuildingLatentGameState is exactly the same (it's just checking a var).

So it's more a matter of if this event should always be triggered every frame, even for "regular players". Or gate it behind a config switch and instruct mod devs to enable this config (which will be a slight backward incompatible change).

With the config switch, we could still check BuildingLatentGameState to avoid some lag for mod devs.

Iridar:
> With the config switch, we could still check BuildingLatentGameState to avoid some lag for mod devs.

Personally I'd prefer doing it this way, because it's gonna be a waste of processing for 99.999999999% CHL mod users out there. Though like I mentioned on discord, if it were up to me, I'd gate every piece of logic that's only useful for debugging behind `final release` check.

remcoros:
Added the config option. I think that's a good compromise, this doesn't force mod devs to compile a custom CHL in non-final-release.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComTacticalController.uc
```diff
@@ -1017,7 +1017,16 @@ function DrawDebugLabels(Canvas kCanvas)
 		// Start Issue #490
 		//
 		// Allow mods to draw their own debug labels to the screen.
-		`XEVENTMGR.TriggerEvent('DrawDebugLabels', kCanvas, self);
+
+		/// HL-Docs: feature:DrawDebugLabels; issue:1245; tags:tactical
```
Iridar, line 1021: Please add a `// Start Issue #1245` above this line.

Also, this arguably doesn't require a HL-Docs entry. The only contributor value in having one is the mention of not triggering events during latent game state building, which is not exactly relevant to drawing debug labels itself. 

IMO mark this as `ref:Bugfixes`, cuz arguably the performance degradation is a bug.

Iridar (changes requested):
Other than the docs nitpick, this is good to go.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComTacticalController.uc
remcoros: The "HL-Docs: feature DrawDebugLabels" is related to the ini setting in XComGame.ini (;;; HL-Docs: ref:DrawDebugLabels' ). I copied the pattern used for bDisableCivilianPreMoveYell.

I can change it to ref:BugFixes here, but what should the new entry in XComGame.ini then have?

Iridar: Just a standard comment with the issue number would be enough, in my opinion. Alternatively, if Draw Debug Labels feature itself has a HL-Docs entry, you can add that one instead.

remcoros: I'll have a look. Ideally, anyone looking at the docs for 'DrawDebugLabels' should see that there is an ini option to be enabled before it will work.

X2WOTCCommunityHighlander/Src/XComGame/Classes/CHHelpers.uc
```diff
@@ -245,6 +245,9 @@ var config float ProjectileParticleSystemExpirationDefaultOverride;
 // Variable for Issue #817 - put the string here, as UIOptionsPCScreen is native and cannot have new variables.
 var localized string m_strFramerateSmoothingDisabledReason;
 
+// Variable added Issue #1245 - enables the triggering of the DrawDebugLabels CHL event
```
Iridar, line 248: Missed "for" here, though I'd prefer just:

// Variable for Issue ...

X2WOTCCommunityHighlander/Config/XComGame.ini
```diff
@@ -268,3 +268,7 @@ bEnableVersionDisplay=true
 ;;; HL-Docs: ref:EuroDateOverride
 ; Uncomment the following line to set European date style for INT localization
 ; bForceEuroDateStrings = true
+
+;;; HL-Docs: ref:DrawDebugLabels
+; Uncomment the following line to enable the DrawDebugLabels CHL event
+; bDrawDebugLabels = true
```
Iridar, line 274: Minor nitpick, in the future please try to always leave an empty line at the end of each file, it makes diffs cleaner.

Iridar:
This is good to go and should be squash merged.

PR #1251, opened 2023-08-21
fix ValidateTier1HackRewards/ValidateTier2HackRewards, when HackRewar…

remcoros:
…dTemplate is none, it should return false

This was introduced in https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/1226
I commented on it, ~~but Iridar might have missed it.~~ I actually missed his reply!

I don't agree with it being out of scope though. It's uc behavior to fall through that if statement with a warning, and still accesses the variable (returning a default value of false). Changing the behavior to fix a log warning is not supposed to happen here :)


Iridar:
Agh I see the problem only now.

`if( !HackRewardTemplate.IsHackRewardCurrentlyPossible() )`

So they have flipped logic in this check. Then yes, it should have been changed in my initial PR.

Iridar:
Sorry, I decided to fix this by committing directly to main branch, since I wanted to first revert the original changes to this file, which meant your PR would have to be rebased first and redo changes to the file, it was easier to just handle it myself. Thanks.

remcoros:
no problem, much cleaner this way.

PR #1254, opened 2023-08-23
Add a check to XCGS_Unit.HasLoadout to verify if the item in the loadout is a valid item

### LINKED ISSUE #1253: XGS_Unit.HasLoadout returns false on missing item templates, resulting in infinitely duplicated items over time.
This started as an investigation of a profile RatherIncoherent send us of his campaign dealing with very severe perf. problems.

The culprit is this line from Spark Arsenal:

[XComGame.X2ItemTemplateManager]
; "EvacFlare" is the item in LWOTC which grants the Evac ability.
+Loadouts=(LoadoutName="RequiredSpark", Items[0]=(Item="IRI_Spark_XPad"), Items[1] = (Item="EvacFlare"))

EvacFlare is a lwotc only item.

In UISquadSelect (https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/X2WOTCCommunityHighlander/Src/XComGame/Classes/UISquadSelect.uc#L753), 'HasLoadout ' and 'ApplyInventoryLoadout' are used to determine if a default loadout should be applied to the unit.

As stated in the comments of 'ApplyInventoryLoadout' this method should only be called once on a unit.

BUT

Looking at the implementation of HasLoadout (https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/master/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc#L8701) it does not actually check if the name of the items in the loadout are actually existing item templates. 

So in this case, because EvacFlare is not a known item (only in lwotc), HasLoadout returns false and the required loadout (in this case the XPad item) is added over and over again.

Iridar already knows about this. 

To prevent mod makers not to shoot themself in the foot when trying to support additional items which may or may not exist in the users campaign, I propose changing the HasLoadout method to check if the given item name is a valid item template name, and skip that item if it does not even exist.

remcoros:
fixes #1253

Testing method:

I changed the following in LWOTC config:

```
-Loadouts=(LoadoutName="RequiredSoldier", Items[0]=(Item="XPad"))
+Loadouts=(LoadoutName="RequiredSoldier", Items[0]=(Item="XPad"), Items[1]=(Item="EvacFlare"), Items[2]=(Item="ZZDummyItem"))
```

Added two log statements where it is checking the item, and if the item was skipped.

When opening Squad Select, this happens:

image

verified that it skipped the non-existing dummy item.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Unit.uc
@@ -8720,6 +8720,12 @@ simulated function bool HasLoadout(name LoadoutName, optional XComGameState Chec
 	{
 		foreach Loadout.Items(LoadoutItem)
 		{
+			// Issue #1253 skip non-existing items, so that it not fails the HasLoadout test if the item template does not exist
+			if (ItemTemplateManager.FindItemTemplate(LoadoutItem.Item) == none)
+			{
+				continue;
+			}
+
 			if(!HasItemOfTemplateType(LoadoutItem.Item, CheckGameState))
 			{
 				return false;


Iridar:
Arguably, the original behavior is correct. Does the unit has the loadout with specified items equipped? No? Then return false. Whether the item actually exists or not is beside the point. With your suggested changes, you can create a loadout with random gibberish, and as far as HasLoadoutEquipped() is concerned, the unit will have it equipped, which doesn't seem right.

remcoros:
The original devs just didn't anticipate the fact that mod makers could use this to add items of other mods. Let's call it future insight.

Another fix would be to make EquipLoadout idempotent, but this is just easier.

It also doesn't undermine current functionality, so I'd say it's pretty safe.

Iridar:
I think the onus is on mod creators to create loadouts that do not risk containing items that don't exist. Push comes to shove, `Loadouts` is just a config array and can be dynamically patched through script.

RustyDios:
If only there was a "the loadout was attempted and partially succeeded" option (redscreen?)
I think the issue is the loadout equipping an item the user already has *again*. (and again, and again)
It should not continually add items for partially completed loadouts.

But I also agree, the onus is on mod creators to ensure loadouts are correct and the items exist in the first place.

remcoros:
I don't see what the cons are. It's a slight improved to make it idempotent and prevent future mistakes, there's no side effects. Maybe a redscreen for a warning.

chrishayesmu:
> I think the onus is on mod creators to create loadouts that do not risk containing items that don't exist.

The problem with this approach is that the consequences of not doing this are nearly invisible. As near as the modder can tell, everything is working properly. It's just a rake waiting to be stepped on.

Iridar:
This is a bit artificial, but here's an example:

1. Mod A adds some items.
2. Mod B adds a loadout that contains those items, and has some logic that wants to do something if the loadout is equipped. Mod B does not check if mod A is loaded or if items in the loadout actually exist.
3. Mod B checks if unit HasLoadout(), which returns true, causing an unintended interaction. 

Once again, I cannot possibly sign off on a function checking if a loadout is equipped returning true if items in the loadout literally do not exist. It is obviously a very bad idea and completely counter to the intended logic of such a function.

Are you currently wearing a Galablerb. Hmm let me check. Galablerbs do not exist. Therefore I am not wearing one. 

^ Seems only logical.

Once again, the onus is on mods to police their loadouts. Not every little thing needs to be fixed in Highlander. This function is working as intended.

>  It's just a rake waiting to be stepped on.

And I stepped on it today, and it was 100% my fault.

remcoros:
Ok I see your point now. 
Just thinking how to prevent this, as this can go on unnoticed for quiet a while.

RustyDios:
I think it just needs logging/redscreening ... not skipping .. 
The current reason it goes unnoticed, is because there seems to be nothing to indicate "item C" of the loadout does not exist

Iridar:
I don't Highlander needs to do anything about this issue. If you really want to, you can add a redscreen/log warning when EquipLoadout tries to equip a non-existent item. You can also add some loadout validation with redscreens and warnings into OnPostTemplatesCreated.

remcoros:
Well, I actually changed my mind a bit.

The source (and CHL docs) clearly state that `ApplyInventoryLoadout` should only be called ONCE on a unit.

image

image

The only way to verify this is using `HasLoadout`

image

Currently, due to a minor config change/error HasLoadout breaks the idempotent contract between HasLoadout/EquipInventoryLoadout and can lead to duplicate EquipInventoryLoadout calls. Which is unintended in my opinion

I DO understand why you'd not want to take such change, I just hope you see my point that this was just an oversight of how those methods work. (Ideally, it should remember the named loadout on the unit, instead of relying on what items are in the loadout)

remcoros:
> There are 33 thousand ways mods can break the game by shipping incorrect config.

fair enough 👍

PR #1266, opened 2023-10-21
Fix auto comment formatting

robojumper:
Untested attempt to fix bad formatting in https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/1263#issuecomment-1773874239 - I can't be bothered to replicate the setup in a new repo, so I guess we can merge this and test afterwards?


Iridar:
Fine by me.

PR #1275, opened 2023-10-22
Allow over-riding weapon base damage for purposes of flyovers

### LINKED ISSUE #1274: X2Effects which modify a weapon's base damage type do not visualise correctly
In X2Action_ApplyWeaponDamageToUnit, the following code runs which is responsible for feeding the damage type to the flyover:

```
//Set up a damage type
    if (WeaponTemplate != none)
    {
        DamageTypeName = WeaponTemplate.BaseDamage.DamageType;
        if (DamageTypeName == '')
        {
            DamageTypeName = WeaponTemplate.DamageTypeTemplateName;
        }
    }
    else if (TickContext != none || WorldEffectsContext != none)
    {
        for (DmgIndex = 0; DmgIndex < UnitState.DamageResults.Length; ++DmgIndex)
        {
            if (UnitState.DamageResults[DmgIndex].Context == StateChangeContext)
            {
                LookupEffect = UnitState.DamageResults[DmgIndex].SourceEffect.EffectRef;
                SourceEffect = class'X2Effect'.static.GetX2Effect(LookupEffect);
                DamageEffect = SourceEffect;
                if (SourceEffect.DamageTypes.Length > 0)
                    DamageTypeName = SourceEffect.DamageTypes[0];
                m_iDamage = UnitState.DamageResults[DmgIndex].DamageAmount;
                break;
            }
        }
    }
    else
    {
        DamageTypeName = class'X2Item_DefaultDamageTypes'.default.DefaultDamageType;
    }
```

This code 'prioritises' the weapon's base damage type & if it isn't able to find one, then looks for the damage type of the effect. It would be useful, for the purposes of effects which apply pure fire/poison/psi damage for the effect damage types to be fed into the flyover even when the effect is directly bound to a weapon.

BlackDog86:
Fixes #1274

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
@@ -200,7 +200,16 @@ function Init()
 		}
 	}
 	// End Issue #326
-
+	
+	/// HL-Docs: feature:OverRideWeaponBaseDamageForFlyover; issue:1274; tags:tactical
+	/// This feature allows modders to specify a boolean value inside X2Effect_ApplyWeaponDamage 
+	/// which will force the damage flyover to use the damage type of the damage effect applied instead
+	/// of the base damage of the weapon. This allows normal weapons to use the Psi flyover, for example
+	if (X2Effect_ApplyWeaponDamage(OriginatingEffect).bIgnoreWeaponBaseDamageTypeForFlyover)
+		{
+		DamageTypeName = X2Effect_ApplyWeaponDamage(OriginatingEffect).EffectDamageValue.DamageType;
+		}
+	
 	bWasHit = false;
 	IsKnockback = false;
 	m_vHitLocation = UnitPawn.GetHeadshotLocation();
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
@@ -19,6 +19,8 @@ var bool    bIgnoreArmor;
 var bool	bBypassSustainEffects;
 var array<name> HideVisualizationOfResultsAdditional;
 
+// Issue #1274
+var bool	bIgnoreWeaponBaseDamageTypeForFlyover;
 // Issue #321
 var config bool NO_MINIMUM_DAMAGE;
 // Issue #743


Iridar:
I think the displayed damage type on the flyover for applied damage should match the damage type actually applied to the target. 

The trick is that an ability can apply damage of several types at the same time, which are gathered from different sources in X2Effect_ApplyWeaponDamage.

The `var array<name> DamageTypes;` lives in `X2Effect`, and if the target is immune to one of the damage types listed in this array, then the effect cannot be applied.

This functions a bit differently for X2Effect_ApplyWeaponDamage. Normally it has the `DamageTypes` array empty, and instead its various functions keep a local array of applied damage types going. It does start with `AppliedDamageTypes = DamageTypes`, and is then gradually appended by damage values gathered from different sources, unless the target is immune to that particular damage type.

I assume the rule of "effect won't be applied if the target is immune to a damage type listed in `DamageTypes`" still holds true for X2Effect_ApplyWeaponDamage, but once that threshold is passed, the effect can gather applied damage types from different sources, ignoring sources that add damage types the target is immune to.

For example, let's take an X2Effect_ApplyWeaponDamage that is set up to apply weapon damage, attached to a sword, and also its own `EffectDamageValue`, that has the `fire` damage type. Then the effect will normally apply both sources of damage summed up together, but if the target is immune to one of these, then only the other one will be applied.

And even if the target is immune to both, the effect will still be applied, the target just won't take any damage.

Unless, of course, either of these damage types is manually added to effect's DamageTypes array.

To summarize, an X2Effect_AWD can potentially apply damage of multiple types, and has its own logic for gathering the damage types. 

Then the question is: which of the damage types should be displayed in flyover?

Well let's think about what the flyover *can* actually display, which is either generic damage or psionic damage. 

So I think the task boils down to checking whether the effect is applying psionic damage or not. If yes, then display the psionic flyover. If not, then we don't care what other damage types are, the flyover will be the same either way.

All that said, to determine whether the effect applies psionic damage or not, we have to repeat the effect's process of gathering the effects either way. 

One huge huge caveat is that a custom X2Effect_ApplyWeaponDamage could have completely arbitrary logic for gathering applied damage types, so we can't reliably make a one size fits all solution that way.

What we can do is check what damage type has actually been applied to the unit. However, since the visualization happens after the effect has already been applied by state code, so we can examine the `var() array<DamageResult> DamageResults;` on the `XComGameState_Unit` we're visualizing taking damage.

```cs
struct native DamageResult
{
	var int DamageAmount, MitigationAmount, ShieldHP, Shred;
	var bool bImmuneToAllDamage;	//	if ALL of the damage being dealt was 0'd out due to immunity, it will be tracked here
	var bool bFreeKill;     //  free kill weapon upgrade forced death of the unit
	var name FreeKillAbilityName;
	var EffectAppliedData SourceEffect;
	var XComGameStateContext Context;
	var array<DamageModifierInfo> SpecialDamageFactors;
	var array<name> DamageTypes;
};
```
Of particular interest are `SourceEffect` - this is how we'll find the `DamageResults` entry relevant to the instance of damage we're visualizing and `DamageTypes` - the bingo we're looking for.

Therefore, the task is:

1. Get the visualized Unit State.
2. Get the originating X2Effect.
3. Iterate over `DamageResults` array on the Unit State until you find the entry the originating effect is responsible for.
4. Check if the `DamageTypes` array on the Damage Result has psionic damage type in it or not. If yes, show psionic damage flyover. If not, show default damage flyover.

I think this is the optimal solution that covers all bases, but discussion is welcome.

BlackDog86:
Seems reasonable to me - I'll try to tackle this over the next few days :)

BlackDog86:
Hey, so I've been looking into this a bit - basically, X2Action_ApplyWeaponDamageToUnit already does something pretty similar to what you're suggesting here - see below:
```
//Set up a damage type
	if (WeaponTemplate != none)
	{
		DamageTypeName = WeaponTemplate.BaseDamage.DamageType;
		if (DamageTypeName == '')
		{
			DamageTypeName = WeaponTemplate.DamageTypeTemplateName;
		}
	}
	else if (TickContext != none || WorldEffectsContext != none)
	{
		for (DmgIndex = 0; DmgIndex < UnitState.DamageResults.Length; ++DmgIndex)
		{
			if (UnitState.DamageResults[DmgIndex].Context == StateChangeContext)
			{
				LookupEffect = UnitState.DamageResults[DmgIndex].SourceEffect.EffectRef;
				SourceEffect = class'X2Effect'.static.GetX2Effect(LookupEffect);
				DamageEffect = SourceEffect;
				if (SourceEffect.DamageTypes.Length > 0)
					DamageTypeName = SourceEffect.DamageTypes[0];
				m_iDamage = UnitState.DamageResults[DmgIndex].DamageAmount;
				break;
			}
		}
	}

	// Start Issue #326: If we still don't have a damage type, try to pull from the manually configured
	/// HL-Docs: ref:Bugfixes; issue:326
	/// Allow damage flyovers from weapon-less Psi abilities to use the Psi damage popup
	// Effect Damage from X2Effect_ApplyWeaponDamage (PsiBombStage2, mod abilities)
	if (DamageTypeName == '')
	{
		if (X2Effect_ApplyWeaponDamage(OriginatingEffect) != none)
		{
			DamageTypeName = X2Effect_ApplyWeaponDamage(OriginatingEffect).EffectDamageValue.DamageType;
		}

		if (DamageTypeName == '')
		{
			DamageTypeName = class'X2Item_DefaultDamageTypes'.default.DefaultDamageType;
		}
	}
	// End Issue #326
```
Basically, the bit inside the 'else/if' which runs for damage types which either tick, or come from the world, does almost exactly what you're suggesting - checks that for each instance of damage on the unit, if it matches the context of the thing activating it, get the originating X2Effect, takes the first element of the damagetypes array associated with that effect and assigns the damagetypename (which decides what type of flyover is used) based on that. The difficulty is that if we re-arrange it to adjust the behaviour such that it gets the damage type from the X2Effect and use that preferentially for the flyover, then abilities like rend (where the damage type on the effect is 'melee' but the damage type on the weapon is 'psi') will display incorrectly. So really, there probably needs to be some kind of flag within X2Effect its-self saying 'ignore base weapon damage for flyovers' - then when grabbing the X2Effects associated with the damage, if that flag isn't set, we skip assigning the damagetypename there and let it fall back to the weapon template instead. Does that make any sense?

PR #1284, opened 2023-10-29
Make X2Effect_BonusWeaponDamage show bonus damage only if it's actually going to be added

### LINKED ISSUE #612: Conventional weapon breakthrough interacts badly with Psi Amps
This actually affects any conventional weapon that ignores base damage. Basically the preview damage includes the breakthrough's +1 damage, but the actual application of that damage does not. So there's an inconsistency from the player's perspective.

The following LWOTC save demonstrates the problem:

[save_BUG Soulfire preview incorrect.zip](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/files/3563256/save_BUG.Soulfire.preview.incorrect.zip)

Soulfire damage preview says it will do 5-7 damage, but in fact it will only do 4-6.

The offending code is in `X2Effect_BonusWeaponDamage`, which seems to deliberately treat the actual damage differently from the damage preview. There seems to be no valid reason for it to do so.

Iridar:
Fixes #612


Iridar:
Tested, works.

Iridar:
This PR caused a bug so I've reverted it for 1.27.1. See #1289 for context.

PR #1285, opened 2023-10-29
Apply weapon customization when equipping unmodified weapons

### LINKED ISSUE #609: `XCGS_Unit:EquipOldItems` does not apply color and pattern to unmodified items
See https://github.com/long-war-2/lwotc/issues/420

Sending a soldier to a covert action without ambush risks unequips a colored weapon, stacks it with the existing stack in HQ. Re-equipping decides that a new copy needs to be made from the infinite stack, but doesn't apply customization to it.

Iridar:
Fixes #609 

This PR does fix the outlined issue, but I need more tests to determine it doesn't go overboard. For example, when working with non-infinite weapons that have weapon upgrades applied to them, but no upgrade slots, e.g. TLP weapons or Chosen XCOM weapons.


Iridar:
I think this is fine, but I'll let it stew for a bit in case @robojumper decides to give it a look.

Iridar:
One thing that has me slightly concerned is using weapon pattern for weapons that are supposed to armor appearance. Like, wouldn't it make sense to use armor pattern for them? But this is what existing code does, so I'll cargo cult this for now. This can be changed later if need be.

PR #1287, opened 2023-11-02
Fix ultrasonic lure expiry duration

### LINKED ISSUE #1286: Ultrasonic lure duration is actually permanent
The duration of ultrasonic lure is said to be 8 turns. However, when you play it in the game and look at the turn counter in F1, turns left will be 8 turns forever.

According to Zelfana
> actually, after checking the source code that is exactly how it works
> which is to say, it's bugged
> the "when to count down" setting is eGameRule_TacticalGameStart by default and they didn't set it to be eGameRule_PlayerTurnBegin instead

BlackDog86:
Fixes #1286

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StatusEffects.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StatusEffects.uc
@@ -2485,8 +2485,8 @@ static function X2Effect_Persistent CreateUltrasonicLureTargetStatusEffect()
 
 	UltrasonicLureTargetEffect = new class'X2Effect_Persistent';
 	UltrasonicLureTargetEffect.EffectName = default.UltrasonicLureName;
-	UltrasonicLureTargetEffect.DuplicateResponse = eDupe_Ignore;
-	UltrasonicLureTargetEffect.BuildPersistentEffect(default.ULTRASONICLURE_TURNS,, true);
+	UltrasonicLureTargetEffect.DuplicateResponse = eDupe_Ignore;	
+	UltrasonicLureTargetEffect.BuildPersistentEffect(default.ULTRASONICLURE_TURNS,, true,,eGameRule_PlayerTurnBegin); //Issue #1286 - Add tick on start of player turn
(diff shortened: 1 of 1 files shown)


Iridar:
Did you test this?

BlackDog86:
Not yet but I checked the arguments for buildpersistenteffect & it definitely defaults to eGameRule_TacticalGameStart if you don't specify an eGameRule as the 5th argument passing into the fn (basically what Zelf mentioned in the original comments). I'll run a quick test tonight to confirm & post a screenshot.

Iridar:
Yes, please do a test with and without the fix.

BlackDog86:
This turned out to be slightly different than I initially anticipated. So, throwing the lure paints a dummy persistent effect on the hostile units in the AOE:
```
static function X2Effect_Persistent CreateUltrasonicLureTargetStatusEffect()
{
	local X2Effect_Persistent UltrasonicLureTargetEffect;

	UltrasonicLureTargetEffect = new class'X2Effect_Persistent';
	UltrasonicLureTargetEffect.EffectName = default.UltrasonicLureName;
	UltrasonicLureTargetEffect.DuplicateResponse = eDupe_Ignore;	
	UltrasonicLureTargetEffect.BuildPersistentEffect(default.ULTRASONICLURE_TURNS,, true,,eGameRule_PlayerTurnBegin); //Issue #1286 - Add tick on start of player turn
	UltrasonicLureTargetEffect.SetDisplayInfo(ePerkBuff_Passive, default.UltrasonicLureFriendlyName, default.UltrasonicLureFriendlyDesc, "img:///UILibrary_PerkIcons.UIPerk_mark" );
	UltrasonicLureTargetEffect.bRemoveWhenTargetDies = true;
	return UltrasonicLureTargetEffect;
}
```

And then applies an effect on all lost units through X2Effect_AlertTheLost. X2Effect_AlertTheLost which is what governs the lost moving towards the painted units has infinite duration so it doesn't make any sense to have the persistent effect applied on the unit to not be infinite either. In the base game, you'd never see this but with mods which allow you see the effects on the unit (e.g. YAF1 and "Show More Buff Details", the poor code relating to the dummy persistent effect is made visible. PR will be altered to update this dummy effect to have infinite duration to match X2Effect_AlertTheLost.

PR #1293, opened 2023-12-19
Fix issue with pillar ability's cover not updating properly on units spawned next to the actor 

### LINKED ISSUE #1288: Templars' Pillar doesn't give cover when spawned directly adjacent to units
Seems like an oversight, it works fine when you spawn a pillar and then move a unit to it, but if spawned right near a unit, it will not provide cover right away.

BlackDog86:
Fixes #1288

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Pillar.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Pillar.uc
@@ -1,0 +1,16 @@
+class X2Effect_Pillar extends X2Effect_SpawnDestructible;
+
+function int GetStartingNumTurns(const out EffectAppliedData ApplyEffectParameters)
+{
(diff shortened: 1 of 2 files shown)


Iridar (changes requested):
This PR is missing .x2proj change for X2Effect_Pillar.

BlackDog86:
Should be done now

Iridar:
Sorry, but it's in the wrong commit now. For PRs that add new files to Highlander, the first commit must have the new file and change to .x2proj file (it helps to restart modbuddy before making the commit so that changes actually apply to the .x2proj file), and the second commit should have the rest of the PR.

BlackDog86:
Commits re-arranged

BlackDog86:
Needs more work - issue was more complex than first thought

Iridar:
In testing this PR doesn't actually fix the issue. The latest commit includes other things I tried, but none of them help either. I'm closing it for now, feel free to reopen later if a solution is found.

PR #1294, opened 2023-12-19
Updated Lure status effect so it ticks properly & isn't removed when the source unit dies

### LINKED ISSUE #1286: Ultrasonic lure duration is actually permanent

BlackDog86:
Fixes #1286

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StatusEffects.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StatusEffects.uc
@@ -2486,8 +2486,12 @@ static function X2Effect_Persistent CreateUltrasonicLureTargetStatusEffect()
 	UltrasonicLureTargetEffect = new class'X2Effect_Persistent';
 	UltrasonicLureTargetEffect.EffectName = default.UltrasonicLureName;
 	UltrasonicLureTargetEffect.DuplicateResponse = eDupe_Ignore;
-	UltrasonicLureTargetEffect.BuildPersistentEffect(default.ULTRASONICLURE_TURNS,, true);
-	UltrasonicLureTargetEffect.SetDisplayInfo(ePerkBuff_Passive, default.UltrasonicLureFriendlyName, default.UltrasonicLureFriendlyDesc, "img:///UILibrary_PerkIcons.UIPerk_mark" );
+	// Start Issue #1286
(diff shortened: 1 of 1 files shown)


BlackDog86:
image
image

Tested and now ticks properly - this issue actually only affects a couple of niche circumstances because the base game includes the ability as a passive (not a debuff) - the "Lost Lure Status Icon" mod changes this to a penalty so the tick then displayed in YAF1 (hence is invisible normally).

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StatusEffects.uc
```diff
@@ -2486,8 +2486,8 @@ static function X2Effect_Persistent CreateUltrasonicLureTargetStatusEffect()
 	UltrasonicLureTargetEffect = new class'X2Effect_Persistent';
 	UltrasonicLureTargetEffect.EffectName = default.UltrasonicLureName;
 	UltrasonicLureTargetEffect.DuplicateResponse = eDupe_Ignore;
-	UltrasonicLureTargetEffect.BuildPersistentEffect(default.ULTRASONICLURE_TURNS,, true);
-	UltrasonicLureTargetEffect.SetDisplayInfo(ePerkBuff_Passive, default.UltrasonicLureFriendlyName, default.UltrasonicLureFriendlyDesc, "img:///UILibrary_PerkIcons.UIPerk_mark" );
+	UltrasonicLureTargetEffect.BuildPersistentEffect(default.ULTRASONICLURE_TURNS,false,false,,eGameRule_PlayerTurnBegin); //Issue #1286 - Keep effect when source dies, tick on turn begin. Adjust icon and displayinfo to display as debuff rather than passive.
```
Iridar, line 2489: Please use Start Issue and End Issue tags when changes span multiple lines, and please add a HL-Docs ref:bugfixes entry.

Iridar (changes requested):

BlackDog86:
Docs tags updated

Iridar (approved):

PR #1307, opened 2024-02-27
Fix X2Effect_BonusWeaponDamage damage preview

### LINKED ISSUE #612: Conventional weapon breakthrough interacts badly with Psi Amps

Iridar:
Fixes #612 
Requires #1306

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_BonusWeaponDamage.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_BonusWeaponDamage.uc
@@ -18,10 +18,16 @@ var int BonusDmg;
 
 // Start Issue #612
 /// HL-Docs: ref:Bugfixes; issue:612
-/// Use `GetAttackingDamageModifier_CH()` to gain access to the damage effect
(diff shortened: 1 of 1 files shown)


Iridar:
I realized that custom damage effects may not always call the new `GetAttackingDamageModifier_CH()`, so it's preferable to use it only for the purposes of damage preview. I brought back the original implementation of `GetAttackingDamageModifier()` so that in cases like that it can still apply the bonus damage correctly, even if it will not be displayed in the damage preview.

Iridar:
In testing the new implementation doesn't work, to be fixed.

PR #1314, opened 2024-02-29
Allow Stun Lance to target Mind Controlled units

### LINKED ISSUE #1312: Stun Lancer's Stun Lance can't target MC'd or hacked units

copyrite:
Fixes #1312

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_StunLancer.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_StunLancer.uc
@@ -1,0 +1,283 @@
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+class X2Ability_StunLancer extends X2Ability;
+
+var private name StunImpairingAbilityName;
+
+static function array<X2DataTemplate> CreateTemplates()
+{
+	local array<X2DataTemplate> Templates;
+	
+	Templates.AddItem(CreateStunLanceAbility());
+	Templates.AddItem(CreateStunLancerImpairingEffectAbility());
+	
+	return Templates;
+}
+
+static function X2AbilityTemplate CreateStunLanceAbility(optional Name AbilityName = 'StunLance')
+{
+	local X2AbilityTemplate                 Template;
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_StunLancer.uc
```diff
@@ -51,6 +51,10 @@ static function X2AbilityTemplate CreateStunLanceAbility(optional Name AbilityNa
 	TargetPropertyCondition = new class'X2Condition_UnitProperty';	
 	TargetPropertyCondition.ExcludeDead = true;                     //Can't target dead
 	TargetPropertyCondition.ExcludeFriendlyToSource = true;         //Can't target friendlies
+
+	/// HL-Docs: ref:Bugfixes; issue:1312
+	/// Allow Stun Lance to target Mind Controlled units (currently on the same team as the Stun Lancer)
+	TargetPropertyCondition.TreatMindControlledSquadmateAsHostile = true;
```
Iridar, line 57: Please add an empty line after this.

Iridar (changes requested):
Not 100% comfortable changing a function that could be potentially getting used by mods to create their own versions of Stun Lance where the changed behavior may or may not be necessary. This change is on the edge between a bugfix and a BC break.

OTOH `TreatMindControlledSquadmateAsHostile` is a really weird flag you would expect to be `true` by default.

copyrite:
I expect this PR to have the same merge conflict as #1313, the project file changes all landed on the same line in alphabetical order. Therefore I'd wait for that one to be merged first.

copyrite:
I'll also repeat here the consideration I brought up on Discord:

I'm not completely sure if there's a mod compatibility risk related to AI. For example, let's say there's a Sectoid and Stun Lancer in play. Sectoid Mind Controls an XCOM soldier. Before the fix, the same XCOM soldier is not a valid target for Stun Lance, but after the fix, it is.

Testing this in a vanilla game both with and without highlander, and verifying the available moves with x2allowselectall, I found that the Lancer will not attack the mind controlled target even if it has the option to do so. But if this feels like it'd cause unwanted changes downstream, I'm not bummed if this PR gets rejected. After all the situation where the player would get burnt by this interaction is very niche; take 2 MCs (one of which is a Lancer) and make them fight against each other.

Issue #1316, opened 2024-03-08
Add hook to allow adjusting UFO interception time


Tedster59:
the LW2 highlander and X2CommunityHighlander featured a `PostUFOSetInterceptionTime` event that allowed LW2 to adjust the UFO interception time since LW2 doesn't use the calendar system.  It's line 115 in [XComGameState_UFO](https://github.com/X2CommunityCore/X2CommunityHighlander/blob/1d3a290f9f38d610dcc7bb1bef95f78df335a0d8/X2CommunityHighlander/Src/XComGame/Classes/XComGameState_UFO.uc#L115C27-L115C53) in the original Highlander.  Somehow this was missed in the WOTC Highlander, and it will be useful for LWoTC.

Iridar:
Closing as this was fixed by #1317

@Tedster59 in the future, please start PR description with `Fixes #<issuenumber>`, that will allow the Issue to close automatically when the PR is merged.

PR #1317, opened 2024-03-08
Re-add event to allow mods to adjust UFO interception time

Tedster59:
Fixes issue #1316.

This event existed in the LW2 Highlander as well as the [non-WOTC X2CommunityHighlander](https://github.com/X2CommunityCore/X2CommunityHighlander/blob/1d3a290f9f38d610dcc7bb1bef95f78df335a0d8/X2CommunityHighlander/Src/XComGame/Classes/XComGameState_UFO.uc#L115C27-L115C53). This pull request re-adds it and documents it so LWoTC can use this event for its original purpose of adjusting the UFO interception time for systems such as Long War that do not use the base game mission calendar.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_UFO.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_UFO.uc
@@ -1,0 +1,650 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    XComGameState_UFO.uc
+//  AUTHOR:  Ryan McFall  --  02/18/2014
+//  PURPOSE: This object represents the instance data for a UFO
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+class XComGameState_UFO extends XComGameState_Airship native(Core);
+
+var() TDateTime						SpawnTime; // The time this UFO appeared on the map
+var() TDateTime						InterceptionTime; // The time this UFO will begin its attack on the Avenger
+var() TDateTime						NextVisibilityTime; // The next time this UFO will become visible
+var() bool							bChasingAvenger; // If the UFO is currently chasing XCOM
+var() bool							bDoesInterceptionSucceed; // The UFO chasing XCOM will result in an Avenger Defense mission
+var() bool							bIsVisible; // Is the UFO currently shown on the map
+var() bool							bNeedsUFOInboundPopup;
+var() bool							bChaseActivated;
+var() bool							bGoldenPath; // Spawned by Golden Path mission
+
+var() config int					MinDisplaySpeed; // Min speed used for UFO Inbound popup
+var() config int					MaxDisplaySpeed; // Max speed used for UFO Inbound popup
+
+var() config int					MinInvisibleHours; // Min hours the UFO will remain visible at a time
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_UFO.uc
```diff
@@ -108,6 +108,22 @@ function SetInterceptionTime()
...
+	///
+	/// Listen to this event with `ELD_Immediate` deferral
+	///
+	/// ```event
+	/// EventID: PostUFOSetInterceptionTime,
+	/// EventData: XComGameState_UFO,
+	/// EventSource: XComGameState_UFO,
+	/// NewGameState: none
+	/// ```
+
```
Iridar, line 125: The empty line before the event trigger is unnecessary, and I would add Start / End Issue #### tags before the start of the docs and after the event trigger.

Also the docs could be formulated a bit better, like:

> The `PostUFOSetInterceptionTime` event allows mods to modify the state of a UFO
> after it has been created, but before the gamestate is submitted. 
> To use this event, listen to it with `ELD_Immediate` deferral and edit the state object received.

There's also this issue to consider: https://www.reddit.com/r/xcom2mods/wiki/index/events#wiki_important_note

In short, when listening to events using ELD_Immediate, the state objects passed as EventSource / EventData are taken from History. To modify the current version of the state object, listeners must explicitly get it from the NewGameState you should be passing with the event, otherwise you would be modifying History.

Iridar (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_UFO.uc
Tedster59: Ok, this has been refactored and updated to shift the call up the stack and pass in NewGameState to the event.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_UFO.uc
```diff
@@ -86,6 +86,27 @@ function SetInterceptionChance(XComGameState NewGameState, bool bGoldenPathSpawn
...
 	SetInterceptionTime();
+
+	// Start Issue #1316
+	/// HL-Docs: feature:PostUFOSetInterceptionTime; issue:1316; tags:strategy
+	/// This event allows mods to modify the state of a UFO 
+	/// after it has been created but before the gamestate is submitted.
+	/// To use this event, listen to it with `ELD_Immediate`
+	/// and edit the state object received in the NewGameState.
+	///
+	/// local XComGameState_UFO            UFOState;
```
Iridar, line 97: Please use ```unrealscript <code> ``` tags around the example code, for example:

```
/// ```unrealscript
/// float GetPreDefaultAttackingDamageModifier_CH(
///         XComGameState_Effect EffectState,
///         XComGameState_Unit SourceUnit,
///         Damageable Target,
///         XComGameState_Ability AbilityState,
///         const out EffectAppliedData ApplyEffectParameters,
///         float CurrentDamage,
///         XComGameState NewGameState)
/// ```
```

Also, nitpicking, but I'd adjust docs a little:

"To use this event, listen to it with the `ELD_Immediate` deferral, and make sure to get the latest version of the UFO state object from the NewGameState, for example: <example code>".

Iridar (changes requested):

PR #1321, opened 2024-03-25
Allow hit results other than `eHit_Success` to display flyovers when using bCombineFlyovers

### LINKED ISSUE #1320: X2Action_ApplyWeaponDamageToUnit::bCombineFlyovers incorrectly handles hit results
# The Problem

I'm not sure what's the purpose behind `bCombineFlyovers`, the only ability in the entire game that uses it is Ionic Storm, and I can't be bothered to trace the code to figure it out. 

But if it's set to `true`, it will check `HitResults[i] == eHit_Success`, ignoring other hit results, like `eHit_Crit`, and in cases like that it will simply not create a flyover, as if the target was completely unaffected by the ability. 

So this will happen only when `bCombineFlyovers = true;` and the ability is not a guaranteed hit, which is not an issue for the base game, as, again, this flag is used only for Ionic Storm, which is a guaranteed hit, but for the sake of mod-added abilities that may want to use this flag, it might be worth fixing. 

# The Fix

Simply replacing `HitResults[i] == eHit_Success` with `class'XComGameStateContext_Ability'.static.IsHitResultHit(HitResults[i])` in the relevant spot should do the trick.

Iridar:
Fixes #1320

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
@@ -884,7 +884,9 @@ simulated state Executing
 				m_iShredded = 0;
 				for( i = 0; i < HitResults.Length && i < DamageResults.Length; i++ ) // some abilities damage the same target multiple times
 				{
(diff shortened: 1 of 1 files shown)


Iridar:
This PR allows crits to display, but misses/grazes/etc will still not show, that would require separate handling.

PR #1330, opened 2024-05-02
Remove cap from bonus damage added by rupture and make rupture apply after attack's damage

### LINKED ISSUE #1299: Rupture is applied before attack damage is calculated
[Relevant code](https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/673de4faaf5b89077f2fead508ec6f6f5b5e2219/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc#L150C3-L153C4)

It's not perfectly clear if this is a bug or intended behavior, since only the Rupture ability uses the rupture mechanic. The ability description implies the Rupture Shot itself shouldn't benefit from this bonus damage.

> A Rupture shot deals critical damage and ensures that the target takes an additional +3 damage from all attacks in the future.

For comparison, nor Shredder nor Holo Target provide any benefit to attacks that apply them. It would make sense for Rupture to behave the same.

Iridar:
Fixes #1299

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_ApplyWeaponDamage.uc
@@ -146,13 +146,17 @@ simulated protected function OnEffectAdded(const out EffectAppliedData ApplyEffe
 				}
 			}
 		}
-		
+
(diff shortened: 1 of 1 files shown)


Tedster59:
This change appears to be causing misses to do damage to ruptured targets.  I set up a scenario with a 0% chance to hit target and they are taking 1 damage

image

Another test using a weapon that does not have built in rupture as well.
image

PR #1336, opened 2024-05-02
Issue #1299 - remove unused local variable

Iridar:
Addendum for Issue #1299 - forgot to comment out the local variable `RuptureCap` which no longer has a use.


RustyDios:
Other than the warning in the build for an unused variable this has potentially fatal knock-on effects for any classes that extend X2effect_applyweapondamage and expect to use/inherit the value from the base class

PR #1339, opened 2024-05-02
Make guaranteed crit abilities display as having 100% crit chance

### LINKED ISSUE #1298: Guaranteed crit abilities do not display as guaranteed crits in the UI
For example, Rupture (`'BulletShred'`) always crits, but you would never know that from the in-game UI, only the ability description.

Iridar:
Fixes #1298

# The Problem

There are two main components to `X2AbilityToHitCalc_StandardAim`, the `GetHitChance()` which is used both for hit chance preview in UI and in the second component, `InternalRollForAbilityHit()`, which does the actual hit roll. 

The `bHitsAreCrits` bool flag is not processed in `GetHitChance()` at all, only in the roll function, where it will replace `eHit_Success` hit result with `eHit_Crit`. So outside the ability description, the player has no way of knowing that the ability will crit.

Here's how it looks in-game:

# Proposed Fix

Give the `bHitsAreCrits`  the same treatment as `bGuaranteedHit`:

1. At the start of `GetHitChance()`, `super.AddModifier()` is called to add 100 to crit chance using the super function from `X2AbilityToHitCalc` class.
2. `AddModifier()` function itself in `X2AbilityToHitCalc_StandardAim` is adjusted to disallow adding crit modifiers if the `bHitsAreCrits` flag is set. 

As the result, the ability will have one and only one crit chance modifier that equals to 100, and all other crit chance modifiers, such as from flanking, will be ignored. So in-game UI for `bHitsAreCrits` abilities will always display exactly 100%.

Here's how it looks:

# Addressing Concerns

This change does not alter handling of `bHitsAreCrits` in `InternalRollForAbilityHit()`, so nothing changes functionality-wise. Custom ToHitCalcs that override `GetHitChance()` or `AddModifier()` , at worst, will display incorrect crit chance, which they would already be doing in regards to `bHitsAreCrits`  abilities.

# Extended Information

Extended Information mod seems to correctly process the changed logic:

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityToHitCalc_StandardAim.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2AbilityToHitCalc_StandardAim.uc
@@ -373,6 +373,16 @@ protected function int GetHitChance(XComGameState_Ability kAbility, AvailableTar
 		AddModifier(100, AbilityTemplate.LocFriendlyName, m_ShotBreakdown, eHit_Success, bDebugLog);
 	}
 
+	// Start Issue #1298
+	/// HL-Docs: ref:Bugfixes; issue:1298
+	/// Add 100 crit chance to guaranteed crit abilities for the purposes of UI.
+	if (bHitsAreCrits)
+	{
+		//  call the super version to bypass our check to ignore crit chance mods for guaranteed crits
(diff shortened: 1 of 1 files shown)


Iridar:
Note: `bAllowCrit` is not being checked because `InternalRollForAbilityHit()` does not check it for `bHitsAreCrits` functionality either, so we can just assume `bHitsAreCrits` overrides `bAllowCrit`.

PR #1341, opened 2024-05-03
Fix weapon model glitching when it's dropped on unit death

### LINKED ISSUE #1290: Weapon glitches when dropped by killed units
Allegedly this bug is fixed by using the Chimera Squad version of `XGUnit::DropWeapon()` function.

Iridar:
Fixes #1290

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComWeapon.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComWeapon.uc
@@ -471,6 +471,34 @@ function ShowWeapon()
 	}
 }
 
+// Start Issue #1290
+// Hides the dropped weapon.
+final function DropWeapon()
+{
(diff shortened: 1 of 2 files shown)


Iridar:
Forgot to mention: ran some tests, including doing save/load, and confirmed the old weapon remains invisible, while the newly dropped weapon remains laying on the ground, though loading a save shifts the position of both the weapon and the corpse.

Iridar:
Addendum: if the weapon is dropped by an XCOM soldier on bleedout, and then brought back with Stabilize and Revival Protocols, they will spawn a new weapon out of thin air, while the dropped weapon will remain lying on the ground. Gonna have to make an additional PR to fix this.

Iridar:
Confirmed the stable CHL doesn't have this issue.

Iridar:
Ugh I don't have the willpower to deal with this right now, I'll just revert it.

PR #1345, opened 2024-05-11
Make it so the [!] icon in the PCS tab appears only if a PCS slot is empty.

### LINKED ISSUE #1348: NeedsAttention icon in the PCS tab ignores if there's a PCS already equipped.
For the weapon mods the little [!] icon appears in a soldier's list only if there's an empty slot for a mod.
The same isn't true for the PCS tab. After researching "Interchangeable PCSs" the orange [!] will remain permanently for every soldier that can equip PCSs. Once you finish that research, bHasNeurochipImplantsInInventory becomes the only condition for the [!] to show, which is practically always, unless you sell every extra PCS you can't immediately equip the moment you get it. 
This isn't exactly a bug but it's not useful to have the game constantly tell the player that they can swap a soldier's equipped PCS with one that might even be worse, just because there is an unused PCS in the player's inventory.

Larryturbo:
Fixes #1348 
For the weapon mods the little [!] icon appears in a soldier's list only if there's an empty slot for a mod.
The same isn't true for the PCS tab. After researching "Interchangeable PCSs" the orange [!] will remain there permanently for every soldier that can equip PCSs.
This isn't exactly a bug but it's not useful to have the game constantly tell the player that they can swap a soldier's equipped PCS with one that might even be worse, just because there is an unused PCS in the player's inventory.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIArmory_MainMenu.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIArmory_MainMenu.uc
@@ -177,14 +177,16 @@ simulated function UpdateData()
 
 	if( bEnableImplantsOption )
 	{
-		//Start issue #1348
 		/// HL-Docs: ref:Bugfixes; issue:1348
-		/// Add additional check so NeedsAttention doesn't trigger if a soldier's PCS slots are full with user equipped PCSs already.
+		/// Display the "needs attention" icon on the PCS button only if the soldier has an empty PCS slot.
 		if( PCSAvailabilityData.bHasNeurochipImplantsInInventory && PCSAvailabilityData.bHasCombatSimsSlotsAvailable && PCSAvailabilityData.bHasEmptyCombatSimSlot) 
+    {
 			PCSButton.NeedsAttention(true);
+    }
 		else
+    {
 			PCSButton.NeedsAttention(false);
-		// End Issue #1348
+    }
 	} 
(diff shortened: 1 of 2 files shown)


RustyDios:
surely this should be;
`PCSAvailabilityData.bNoCombatSimsEquipped = ( EquippedImplants.Length <= 0 );`

chrishayesmu:
> surely this should be;
> `PCSAvailabilityData.bNoCombatSimsEquipped = ( EquippedImplants.Length <= 0 );`

I think the logic is right, but the name is wrong. It should be something like `bHasEmptyCombatSimSlots`.

RustyDios:
which is what `PCSAvailabilityData.bHasCombatSimsSlotsAvailable` does?

RustyDios:
oh! see this is where the issue is because of the `|| XComHQ.bReusePCS;` right ?

Larryturbo:
> oh! see this is where the issue is because of the `|| XComHQ.bReusePCS;` right ?

Yes because of that, once you finish the research, bHasNeurochipImplantsInInventory becomes the only condition for the [!] to show for every single soldier, which is practically always, unless you go and sell every extra PCS you can't immediately equip. I didn't want to alter the original variable since it has other uses, bHasEmptyCombatSimSlots is a better name, but the logic is correct, in case some mod adds more slots or something.

Iridar (approved):
Tested and made some minor adjustments, mainly to docs.

PR #1346, opened 2024-05-11
Proposed fix for the persistent "Low Intel" warning after contacting all regions.

### LINKED ISSUE #1347: Persistent "Low Intel" warning after contacting all regions
The "Low Intel" warning shows up if your total intel is less than the contact cost of the next closest region. But if you've contacted all regions the cost defaults to 9999 and you get the warning despite not having more regions to contact.

Larryturbo:
Fixes #1347 
The "Low Intel" warning shows up if your total intel is less than the contact cost of the next closest region. But if you've contacted all regions the cost defaults to 9999 and you get the warning despite not having more regions to contact. I've added an additional check to make sure your current resistance contacts aren't equal to the total number of regions that can be contacted (probably 16 but I didn't want to hardcode it) before the warning is displayed.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
@@ -531,6 +531,7 @@ simulated function DisplayWarningPopups()
 	local XComGameState_HeadquartersXCom XComHQ;
 	local TDateTime StartDateTime, CurrentTime;
 	local int MinStaffRequired, NumStaff, MonthsDifference;
+	local array<XComGameState_WorldRegion> AllRegions; // For Issue #1347 
 
 	XComHQ = XComGameState_HeadquartersXCom(`XCOMHISTORY.GetSingleGameStateObjectForClass(class'XComGameState_HeadquartersXCom'));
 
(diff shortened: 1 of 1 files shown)


BlackDog86:
Please can you squash all commits into a single commit?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
```diff
@@ -541,7 +541,13 @@ simulated function DisplayWarningPopups()
 
 	if (!XComHQ.bHasSeenLowIntelPopup && XComHQ.GetIntel() < class'UIUtilities_Strategy'.static.GetMinimumContactCost() && XComHQ.IsContactResearched())
 	{
-		UILowIntel();
+		// Start Issue #1347
+		// Show warning only if there are still regions left to be contacted.
+		if(XComHQ.GetCurrentResContacts() < class'X2StrategyElement_DefaultMissionSources'.static.GetAllRegions().Length)
```
Iridar, line 546: I'm pretty sure you can't do `.Length()` on a dynamic array returned by a function. Stepped on these rakes too at some point. It's always just going to be zero.

Iridar (changes requested):

Iridar:
Good find, but the code needs to be adjusted and actually tested, because right now I believe it will the Low Intel warning to just show never, because `class'X2StrategyElement_DefaultMissionSources'.static.GetAllRegions().Length` will evaluate to zero. 

Highlander is sirius business, please make sure to properly test all submissions.

Don't worry about squashing, it can be done during the merging process.

Iridar:
I tested, and while this solution works for the Low Intel popup, it doesn't work for the Low Intel warning in the lower left corner, and probably several other UI elements that use the `GetMinimumContactCost()` function, which is where the problem is. Thus, I suggest fixing the issue in that function instead.

PR #1351, opened 2024-05-31
Properly trigger unit concealment loss when bForceNoSquadConcealment flag is set instead of not concealing in the first place.

### LINKED ISSUE #1319: ADVENT does not extract supplies from the Extract Supplies mission if XCOM starts unconcealed
> I'm going on an extract supplies mission where you mark crates for Avenger to pick up before Advent gets them - lost city, fairly late game. As I hit the map I realize that one of the Dark Events is still active - the one where you start the mission without concealment.
> 
> My first thought is how is this going to work? Will the crates start getting marked and taken away after the first round? After I first encounter Advent? Normally it happens right after you break concealment and the fighting starts....
> 
> ...but since I am not in concealment, there is no concealment to break, and Advent never marked any crates. I was able to take my time and get them all (even managed to lure a Sectopod away from the two crates he was standing next too before I blew him up). I guess its a combo of events that they never thought of when programming this mission type...

https://www.reddit.com/r/XCOM2/comments/1bgn9y1/well_thats_new_extract_supplies_mission/

BlackDog86:
Fixes #1319 

This is a preliminary fix for review & likely will need some changes but just creating a PR for review / conversation.

The code adjustment fixes the problem by starting the squad in concealment and then breaking it if the dark event / battledata flag bForceNoSquadConcealment is set, instead of not setting concealment on the squad in the first place. It works properly with phantom & reaper shadow but at the moment it creates a lot of flyovers on mission start which I suspect which looks a bit messy & I suspect will prevent this from being merged directly - I haven't sorted the comments out yet either!

Basically, on missions where the squad starts concealed and high alert is active you now get the following situation:
"High Alert Flyover"
"Concealed Flyover"
"Revealed Flyover"
(Optional if you have a ranger) "Phantom Ranger Concealed Flyover"

Normal mission start flyovers are not affected.

It would be ideal if we could come up with a way to suppress the visualiser when we use the SetSquadConcealment function but I don't really know how to do that at this point. Advice welcome!

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
@@ -1246,6 +1246,8 @@ function ApplyStartOfMatchConditions()
 	local X2HackRewardTemplate Template;
 	local XComGameState NewGameState;
 	local Name HackRewardName, ConcealmentName;
+	// Variable for Issue #1319
+	local XComGameState ConcealmentBrokenGameState;
 
 	History = `XCOMHISTORY;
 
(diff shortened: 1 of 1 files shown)


Iridar:
Since the issue is specifically with the mission, perhaps it's the mission that needs to be fixed? Perhaps changing the mission trigger from "XCOM breaks concealment" to "XCOM unit sighted". Either that or "or mission start if XCOM starts unconcealed". The latter is less friendly towards the player, but makes more sense with the mission narrative that goes "ADVENT already knows we're here".

BlackDog86:
Think I've found a workable solution to this - it seems the mission kismet which triggers the start of the crate pickup is set to trigger with the 'SquadConcealmentBroken' event, so we can simply get the event manager to fire off this event if the bForceNoSquadConcealment flag is set on the battle data. Preliminary testing suggests this seems to work. Will update shortly!

Iridar:
I'm gonna request some more testing done on this, specifically:
1. Produce the issue 
2. Confirm the fix addresses the issue.
3. Do a reasonable amount of testing to make sure the fix doesn't break other missions with the dark event / battle data flag present.

BlackDog86:
1. I can categorically confirm that without this fix & just CHL active, if you spawn the high alert dark event and attempt an Advent Supplies extraction, the AI will never collect any crates. 
2. I can also categorically confirm that triggering the 'SquadConcealmentBroken' event on mission start (or indeed by toggling concealment on & off via console command) fixes the problem and allows the mission to proceed normally (i.e. ADVENT starts marking crates immediately on the first turn. Obviously, this increases the difficulty of the mission but I think the argument that not triggering the start of crate collection with that DE active was an intentional design decision is a stretch!).
3. Tested a good few mission types so far (Hack workstation, Gather survivors, Sabotage transmitter) - no adverse effects so far.

Iridar:
I've updated the docs and code styling, please take a look and if you don't see anything objectionable, this is good to go.

BlackDog86:
Happy with your adjustments :)

PR #1355, opened 2024-07-10
Add event to override patrol behavior if pod is spotted in stealth

### LINKED ISSUE #1354: Add hook to allow overriding patrol behavior if pod seen by unrevealed xcom unit
Another LW2/X2CommunityHighlander feature that is not part of the WOTC Highlander.

Code is at https://github.com/X2CommunityCore/X2CommunityHighlander/blob/master/X2CommunityHighlander/Src/XComGame/Classes/XGAIPlayer.uc#L702

The code as is from the X2CommunityHighlander should work to update the matching function in `XGAIPlayer.uc`

The comment from LW2 config shows why this is needed:

> This adjusts an interesting vanilla setting in which patrolling AI (who had not sighted XCOM) would stop patrolling when a concealed XCOM unit sighted them)

Tedster59:
fixes #1354.

This event allows mods to override the default behavior of `XGAIPlayer.ShouldUnitPatrol`.

Default behavior ties the patrol behavior to `UnitState.IsUnrevealedAI` meaning that units spotted by concealed XCOM units stop patrolling. This event allows this behavior to be overridden by mods.  The original hook was present in the LW2 Highlander and X2CommunityHighlander, but was not ported to WOTCCommunityHighlander.

I have tested this hook with a dev build of LWOTC and it functions as expected.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIPlayer.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIPlayer.uc
@@ -750,11 +750,42 @@ static function bool IsMindControlled(XComGameState_Unit UnitState)
 // Update - green alert units and units that have not yet revealed should do their patrol movement.
 function bool ShouldUnitPatrol( XComGameState_Unit UnitState )
 {
+	// Variables for Issue #1354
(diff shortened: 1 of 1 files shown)


Tedster59:
Closing for code cleanup and making new PR

PR #1357, opened 2024-07-10
Update ShowPromotionUI to not cast as directly

### LINKED ISSUE #1356: ShowPromotionUI casts to subclasses of UIArmory_Promotion
`XComHQPresentationLayer` contains the following function:

`function ShowPromotionUI(StateObjectReference UnitRef, optional bool bInstantTransition)
{
    local UIArmory_Promotion PromotionUI;
    local XComGameState_Unit UnitState;

    UnitState = XComGameState_Unit(`XCOMHISTORY.GetGameStateForObjectID(UnitRef.ObjectID));

    // Start Issue #600: Replaced class literals with the local variables
    if (UnitState.IsResistanceHero() || ScreenStack.IsInStack(class'UIFacility_TrainingCenter'))
        PromotionUI = UIArmory_PromotionHero(ScreenStack.Push(
            Spawn(TriggerOverridePromotionUIClass(eCHLPST_Hero), self), Get3DMovie()));
    else if (UnitState.GetSoldierClassTemplateName() == 'PsiOperative')
        PromotionUI = UIArmory_PromotionPsiOp(ScreenStack.Push(
            Spawn(TriggerOverridePromotionUIClass(eCHLPST_PsiOp), self), Get3DMovie()));
    else
        PromotionUI = UIArmory_Promotion(ScreenStack.Push(
            Spawn(TriggerOverridePromotionUIClass(eCHLPST_Standard), self), Get3DMovie()));
    // End Issue #600
    
    PromotionUI.InitPromotion(UnitRef, bInstantTransition);
}`

While we fire an override event to allow mods to change the promotion UI, it is then casted back into `UIArmory_PromotionHero`, `UIArmory_PromotionPsiOp`, or `UIArmory_Promotion` depending on the type of unit. This means that changing the screen type for Psi Ops or Hero units will cause issues if the screen is not a subclass of the class being casted.

For example, trying to allow Psi Operatives to use Community Promotion Screen causes the screen to not initialize because it is being casted to `UIArmory_PromotionPsiOp`

image

Proposed fix: Change all these casts to the basic `UIArmory_Promotion` class instead. Tested and working for Psi Operative.

image-1

Tedster59:
Fixes #1356. Updates the function to cast only to UIArmory_Promotion instead of its subclasses, allowing Psi Operative to use Community Promotion Screen. Also adds missing brackets to if/else statements. Tested with LWOTC and also reviewed by RustyDios and Zelfana.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
@@ -1497,16 +1497,27 @@ function ShowPromotionUI(StateObjectReference UnitRef, optional bool bInstantTra
 
 	UnitState = XComGameState_Unit(`XCOMHISTORY.GetGameStateForObjectID(UnitRef.ObjectID));
 
+	/// HL-Docs: ref:Bugfixes; issue:1356
+	/// ShowPromotionUI now casts the spawned promotion screen to UIArmory_Promotion instead of its subclasses
+
 	// Start Issue #600: Replaced class literals with the local variables
 	if (UnitState.IsResistanceHero() || ScreenStack.IsInStack(class'UIFacility_TrainingCenter'))
-		PromotionUI = UIArmory_PromotionHero(ScreenStack.Push(
+	{
+		// Issue #1356 - Cast to UIArmory_Promotion instead of UIArmory_PromotionHero
+		PromotionUI = UIArmory_Promotion(ScreenStack.Push(
 			Spawn(TriggerOverridePromotionUIClass(eCHLPST_Hero), self), Get3DMovie()));
+	}
 	else if (UnitState.GetSoldierClassTemplateName() == 'PsiOperative')
-		PromotionUI = UIArmory_PromotionPsiOp(ScreenStack.Push(
+	{
+		// Issue #1356 - Cast to UIArmory_Promotion instead of UIArmory_PromotionPsiOp
+		PromotionUI = UIArmory_Promotion(ScreenStack.Push(
 			Spawn(TriggerOverridePromotionUIClass(eCHLPST_PsiOp), self), Get3DMovie()));
+	}
 	else
+	{
 		PromotionUI = UIArmory_Promotion(ScreenStack.Push(
 			Spawn(TriggerOverridePromotionUIClass(eCHLPST_Standard), self), Get3DMovie()));
+	}
 	// End Issue #600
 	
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
```diff
@@ -1497,6 +1497,7 @@ function ShowPromotionUI(StateObjectReference UnitRef, optional bool bInstantTra
 
 	UnitState = XComGameState_Unit(`XCOMHISTORY.GetGameStateForObjectID(UnitRef.ObjectID));
 
+	/*
```
Iridar, line 1500: This confused me a bit, it seems you commented out original block of code and put a new one nearby. That's fine for testing, but main repo needs to be a bit more tidy, so adjust individual lines instead.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
```diff
@@ -1508,6 +1509,30 @@ function ShowPromotionUI(StateObjectReference UnitRef, optional bool bInstantTra
 		PromotionUI = UIArmory_Promotion(ScreenStack.Push(
 			Spawn(TriggerOverridePromotionUIClass(eCHLPST_Standard), self), Get3DMovie()));
 	// End Issue #600
+	*/
+
+	/// HL-Docs: ref:Bugfixes; issue:1356
+	/// ShowPromotionUI no longer casts the resulting screen type to subclasses of UIArmory_Promotion
```
Iridar, line 1515: I think something like "ShowPromotionUI now casts the spawned promotion screen to UIArmory_Promotion instead of its subclasses" would be a bit more clear.

Iridar (changes requested):
Otherwise I don't mind the concept behind this change. Seems like the cast was unnecessary in the first place.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
Tedster59: Addressed

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComHQPresentationLayer.uc
Tedster59: Addressed

PR #1359, opened 2024-07-14
Make it so 2nd tier Alien Hunters weapons show up on the trophy wall even if player skipped straight to 3rd tier.

### LINKED ISSUE #1358: 2nd tier alien hunter weapons not showing up on the trophy wall
At script XComGameState_HuntersLodgeManager.uc.
The second and third tier versions of the alien hunter weapons show up independently. Meaning that if you skip the second tier and make the beam versions you only get the 1st and 3rd tiers on the wall.

Larryturbo:
Fixes #1358


Iridar:
I don't see how this is a bug? You didn't purchase the second tier, why should you get to see second tier? If you skip aircraft and go from cars straight to rockets, it doesn't automatically grant you aircraft, you can't build them on rocket factories. Closing until I hear some compelling aruments.

Larryturbo:
The argument is that if you first research the alien hunter weapons at beam tier you still get the basic tier. If you first build the axe at beam tier you should either only get the 3rd tier on the wall or all 3, currently you get the 1st and the 3rd, which is the least sensible way to have it.

Iridar:
In that case I'd be in favor of not granting the first tier.

PR #1361, opened 2024-07-17
Update Chinese localization

xgl1991714:


Iridar (approved):
Well I can't read Chinese, so I'll just assume everything's good here.

xgl1991714:
> Well I can't read Chinese, so I'll just assume everything's good here.

Thank you very much. No worries, I usually test what I submit locally first. Initially, I just intended to submit the translation for 'RuptureLabel,' but then I went ahead and translated everything. I'm a bit curious as to why there's a specific 'RuptureLabel' in X2WOTCCommunityHighlander. Some weapons with the Rupture property seem to use this translation.

PR #1363, opened 2024-07-24
Added a new EndBattleForceWin console function which finishes the tactical battle & completes all mission objectives

### LINKED ISSUE #386: Enhance console commands (or at least EndBattle)
At this point issues with commands such as `EndBattle` not working as expected in common use cases (such as trying to end a bugged mission in one's favour) are pretty well known. I've been thinking that maybe we should consider having the Highlander add a couple commands of its own for when vanilla console commands are unable to meet the common task (since the scripts the Firaxis cheats are in are reflected in native code), or bugfixing them when it is possible.

ie fixing the various soldier-related commands on the strategy layer to not check for `GetMyTemplateName() == 'Soldier' `, but instead use `IsSoldier()` whenever possible, or use a more complicated check if it's something that might break with exotic units such as SPARKs.

The relevant files for this are `XComHeadquartersCheatManager` and `XComTacticalCheatManager`, though I think there might be another file containing some old commands that still work with X2 regardless.

BlackDog86:
Fixes #386

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComTacticalCheatManager.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComTacticalCheatManager.uc
@@ -1106,6 +1106,35 @@ exec function TestClosestPointToCursorOnAxis()
 }
 //------------------------------------------------------------------------------------------------
 
+exec function EndBattleForceWin()
+{
+	// Start Issue #386
+	/// HL-Docs: feature:ImprovedEndBattle; issue:386; tags:tactical
+	/// Adds a new console command, 'EndBattleForceWin' which ends the tactical engagement and forces all mission objectives to completed 
+	local X2TacticalGameRuleset		Ruleset;
+	local XGBattle_SP				Battle;
+	local int						ObjectiveIndex;
+	local XComGameState_BattleData	BattleData;
+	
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComTacticalCheatManager.uc
```diff
@@ -1106,6 +1106,35 @@ exec function TestClosestPointToCursorOnAxis()
 }
 //------------------------------------------------------------------------------------------------
 
+exec function EndBattleForceWin()
+{
+	// Start Issue #386
+	/// HL-Docs: feature:ImprovedEndBattle; issue:386; tags:tactical
+	/// Add additional function which ends the tactical engagement while also forcing all objectives to be completed
```
robojumper, line 1113: This adds a *console command* and the documentation should mention the name of the console command.

robojumper (changes requested):
There's also some weird formatting/indentation going on here.

BlackDog86:
Updated

Iridar (commented):
Rejected.

BlackDog86:
In that case I would close issue #386 as well then. I've asked XPanD to include the function in his console commands mod instead.

PR #1365, opened 2024-07-27
Add docs - Torso validation

BlackDog86:
Adds docs for Issue #350


X2WOTCCommunityHighlander/Src/XComGame/Classes/UICustomize_Body.uc
```diff
@@ -69,6 +69,9 @@ simulated function UpdateData()
 	ResetMechaListItems();
 
 	super.UpdateData();
+	
+	/// HL-Docs: ref:Bugfixes; issue:350
+	/// Allow separation of arm body part types
```
Iridar, line 74: How does doing List.ClearItems() help in the task of "separation of arm body part types"?

BlackDog86: Copypaste error - corrected

Iridar:
I hate to nitpick, but current PR seems to touch two extra files without actually changing anything in them, and the HL-Docs entry for bugfixes is not very clear, ideally it should be something mod users can understand as a changelog entry on the bugfixes page.

PR #1367, opened 2024-07-27
Add docs - Gremlin customization

BlackDog86:
Adds docs for Issue #380

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
@@ -2309,15 +2309,16 @@ simulated function SpawnCosmeticUnitPawn(UIPawnMgr PawnMgr, EInventorySlot InvSl
 		return;
 	}
 
-	// Start Issue #380: Moved earlier because we want to SetAppearance() whether we create a new pawn or not
+	// Start Issue #380: 
+	/// HL-Docs: ref:Bugfixes; issue:380
+	/// Ensure that the appearence of cosmetic unit pawns is updated properly, even when a new pawn is not created
 	UseAppearance = OwningUnit.kAppearance;
 	UseAppearance.iArmorTint = UseAppearance.iWeaponTint;
 	UseAppearance.iArmorTintSecondary = UseAppearance.iArmorTintSecondary;
 	UseAppearance.nmPatterns = UseAppearance.nmWeaponPattern;
-	// End Issue #380
+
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
```diff
@@ -2308,15 +2308,17 @@ simulated function SpawnCosmeticUnitPawn(UIPawnMgr PawnMgr, EInventorySlot InvSl
...
 
-	// Start Issue #380: Moved earlier because we want to SetAppearance() whether we create a new pawn or not
+	// Start Issue #380: 
+	/// HL-Docs: ref:Bugfixes; issue:380
+	/// Moved earlier so we can call SetAppearance() whether a new pawn is created or not
 	UseAppearance = OwningUnit.kAppearance;
 	UseAppearance.iArmorTint = UseAppearance.iWeaponTint;
 	UseAppearance.iArmorTintSecondary = UseAppearance.iArmorTintSecondary;
 	UseAppearance.nmPatterns = UseAppearance.nmWeaponPattern;
-	// End Issue #380
```
Iridar, line 2316: I see no reason to remove these two issue tags.

Iridar (changes requested):

Iridar:
I'm sorry, but this still seems to remove Start Issue / End Issue tags in XComUnitPawn, what's up with that?

BlackDog86:
> I'm sorry, but this still seems to remove Start Issue / End Issue tags in XComUnitPawn, what's up with that?

I explained this in the comment above -

The original file had this:
```
issue # X
<some code>
end issue # X    <---- remove this 
issue # X again <----- remove this 
<some code>
end issue # X 
```

Then you end up with
```
issue #x
<some code>
end issue #x
```

Iridar:
I see. 

The purpose of ref:Bugfixes docs is to be user-facing. As in, it should be treated as an entry in a changelog that explains the bug that is being fixed to Highlander users. If it's just a comment that explains to Highlander maintainers and Highlander-using mod authors, it doesn't need a docs entry.

Something like "Moved earlier so we can call SetAppearance() whether a new pawn is created or not" doesn't really explain what is being fixed here, so if you could reword it a bit, it would be appreciated.

BlackDog86:
> I see.
> 
> The purpose of ref:Bugfixes docs is to be user-facing. As in, it should be treated as an entry in a changelog that explains the bug that is being fixed to Highlander users. If it's just a comment that explains to Highlander maintainers and Highlander-using mod authors, it doesn't need a docs entry.
> 
> Something like "Moved earlier so we can call SetAppearance() whether a new pawn is created or not" doesn't really explain what is being fixed here, so if you could reword it a bit, it would be appreciated.

Yep, I see what you mean. That should be done now.

PR #1371, opened 2024-08-03
Allow removal of lost units from AoE targetting profiles

### LINKED ISSUE #1369: Allow mods to adjust AI behaviour of AoE profiles to specifically exclude lost units
Currently, XGAIBehaviour does not have any specific means of excluding lost units from AoE targetting profiles. Setting bTargetTheLost in XComAI.ini allows units to explicitly **include** lost for abilities which wouldn't otherwise (this is only set for Harbor Wave and purifier flamethrower in the base game) but there is nothing which specifically excludes lost units when bTargetEnemy is set (which is thge default setting for all abilities if not explicity set in AI config).

Proposal would be to add an additional variable (bExpliciltyExludeLost or similar) to the AoETargetingInfo struct and set up some code to remove lost units from the valid targets UnitList array if that bool is set (similar to how the Deprioritised effects bit works in the GetUnfilteredAOETargetList function).

Advice on whether this would be a sensible approach would be greatly appreciated!

BlackDog86:
Fixes #1369 - Tested with modified version of alien pack with bIgnoreTheLost set. Was unable to get grenadiers or vipers to target lost - mission successful! :)

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIBehavior.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIBehavior.uc
@@ -48,8 +48,10 @@ struct native AoETargetingInfo
 	var bool bAreaSearchSpace;			// if true, also target adjacent points around our test locations.
 	var bool bRequireLoS;				// Require Line of Sight between the shooter and the target
 	var bool bCountAdjacentAsHits;		// if true, hitting adjacent tiles to targets count as a valid hit.
-	var bool bTestTargetEffectsApply;	// Only consider units that are valid for the MultiTarget effects on the ability.
+	var bool bTestTargetEffectsApply;	// Only consider units that are valid for the MultiTarget effects on the ability.	
 	var bool bIgnoreRepeatAttackList;   // Ignore restrictions on tiles that have already been used for AoEs.
+	// Variable for Issue #1369
+	var bool bIgnoreTheLost;			// Ignore lost units from AoE Targetting calculations	 
 
 	structdefaultproperties
 	{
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIBehavior.uc
```diff
@@ -48,8 +48,10 @@ struct native AoETargetingInfo
 	var bool bAreaSearchSpace;			// if true, also target adjacent points around our test locations.
 	var bool bRequireLoS;				// Require Line of Sight between the shooter and the target
 	var bool bCountAdjacentAsHits;		// if true, hitting adjacent tiles to targets count as a valid hit.
-	var bool bTestTargetEffectsApply;	// Only consider units that are valid for the MultiTarget effects on the ability.
+	var bool bTestTargetEffectsApply;	// Only consider units that are valid for the MultiTarget effects on the ability.	
 	var bool bIgnoreRepeatAttackList;   // Ignore restrictions on tiles that have already been used for AoEs.
+	// Variable for Issue #1369
+	var bool bIgnoreTheLost;			// Ignore lost units from AoE Targetting calculations
```
robojumper, line 54: This is a really iffy change - `AoETargetingInfo` is native, so this actually changes a native data type, which is generally not allowed. In this case it goes well since bools are represented as bitfields and this effectively doesn't change the layout but I think changes like this should have a better motivation than "could be useful for mods".

BlackDog86: Would there be any other obvious ways of approaching this?

PR #1373, opened 2024-08-10
Use List.ClearItems() instead of List.Itemcontainer.RemoveChildren() to prevent CTDs with large lists

### LINKED ISSUE #1372: Game crashes to desktop when a pose is selected in the photobooth, when large numbers of poses are in the list.
UIPhotoboothBase uses List.ItemContainer.RemoveChildren(); when a selection is made on this screen. It's a native call which is assumed to nuke the ItemContainer & its contents. However, if the item container is too large, too much information is allocated on the Flash stack at once and a CTD occurs.

Replacing the call with List.ClearItems(); in the OnCancel and OnPoseConfirm functions in UIPhotobooth base prevents the crash, presumably by more safely disposing of the container.

BlackDog86:
Fixes #1372

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIPhotoboothBase.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UIPhotoboothBase.uc
@@ -692,7 +692,8 @@ function OnCancel()
 		currentState = eUIPropagandaType_Graphics;
 		break;
 	}
-	List.ItemContainer.RemoveChildren();
+	//Issue #1372 - Replace List.ItemContainer.RemoveChildren() with List.ClearItems() to prevent CTDs with large lists
+	List.ClearItems();
 	NeedsPopulateData();
 }
 
@@ -751,7 +752,8 @@ function OnConfirmFormation()
 {
 	List.OnSelectionChanged = none;
 	currentState = eUIPropagandaType_Base;
-	List.ItemContainer.RemoveChildren();
+	//Issue #1372 - Replace List.ItemContainer.RemoveChildren() with List.ClearItems() to prevent CTDs with large lists
+	List.ClearItems();
 	NeedsPopulateData();
 }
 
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/UIPhotoboothBase.uc
```diff
@@ -2114,7 +2120,7 @@ function OnConfirmLayout()
 {
 	List.OnSelectionChanged = none;
 	currentState = eUIPropagandaType_Base;
-	List.ItemContainer.RemoveChildren();
+	List.ClearItems();
```
robojumper, line 2123: This is missing a change comment.

robojumper (changes requested):
One of these lines could probably use a Bugfixes comment like
```
/// HL-Docs: ref:Bugfixes; issue:1372
/// Fixed a potential crash when the Photobooth shows many list items.
```

Also, please add "Fixes #1372" to the PR description to link this PR to the related issue.

BlackDog86:
Changes made as requested

Iridar:
I assume this was tested and confirmed to fix the actual problem?

Tedster59:
@Iridar  - Currently released as a standalone invasive MCO until the Highlander release here: https://steamcommunity.com/sharedfiles/filedetails/?id=3307455200, and this was also identified as the issue via Robojumper/Crash dump investigation

BlackDog86:
> @Iridar - Currently released as a standalone invasive MCO until the Highlander release here: https://steamcommunity.com/sharedfiles/filedetails/?id=3307455200, and this was also identified as the issue via Robojumper/Crash dump investigation

That mod also does some other stuff with the UI - this fix literally just stops the crash, but leaves all the poses in a big list on the screen.

Iridar (approved):

Iridar:
I see. No reason to hold this up then. I checked and looks like these are the only places in code where `RemoveChildren()` is even called. It might be a good idea to add a comment that would caution against using it.

PR #1374, opened 2024-08-15
Set bConstantComplete to true for cleaned up particle systems

### LINKED ISSUE #1201: Grappling sometimes crashes the game
[Allegedly](https://www.reddit.com/r/Xcom/comments/13zxc4h/please_help_my_game_crashed_when_use_grapple/joggfca/?context=10000) this bug is caused by Highlander itself. 

I've experienced this crash myself, and I think it might be caused by a relatively recent addition. 

The first order of business is doing some extensive testing with the grapple with and without Highlander to confirm the problem, and ideally find reproduction steps.

If we can at least confirm the problem is caused by highlander, the next step is to keep rolling back through versions until we can identify the version that causes the crash. 

Then we can roll back issue commits one by one until we find the exact cause.

Tedster59:
Fixes #1201 based on crash dump analyzed by Robojumper and further discussion on discord, and implements Robojumper's suggested fix: https://discord.com/channels/165245941664710656/167912351763398657/1273396231011434517

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
@@ -463,8 +463,7 @@ function EndConstantProjectileEffects( )
 			// an access violation. It's unclear whether holes in the `ActiveComponents`
 			// can legitimately happen or indicate another bug, so the None check here
 			// is an additional defensive measure.
-
-			if(Projectiles[ Index ].ParticleEffectComponent != None)
+			if (Projectiles[ Index ].ParticleEffectComponent != None)
 			{
 				WorldInfo.MyEmitterPool.OnParticleSystemFinished(Projectiles[ Index ].ParticleEffectComponent);
 				CancelDelayedReturnToPoolPSC(Projectiles[ Index ].ParticleEffectComponent);
@@ -1170,7 +1169,7 @@ function OnParticleSystemFinished(ParticleSystemComponent PSystem)
 			// Allow the pool to reuse this Particle System's spot in the pool.
 			//WorldInfo.MyEmitterPool.OnParticleSystemFinished(Element.ParticleEffectComponent);
 			// Issue #1201 none check
-			if(Projectiles[i].ParticleEffectComponent != none)
+			if (Projectiles[i].ParticleEffectComponent != none)
 			{
 				WorldInfo.MyEmitterPool.OnParticleSystemFinished(Projectiles[i].ParticleEffectComponent);
 			}
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
@@ -455,8 +455,25 @@ function EndConstantProjectileEffects( )
 			/// if the PSC's `OnSystemFinished` delegate is replaced, then `EmitterPool::OnParticleSystemFinished()` must
 			/// be called for this PSC manually when the PSC is no longer needed, otherwise the same "memory leak" 
 			/// will occur. This applies to all PSCs using the Emitter Pool, not just those in `X2UnifiedProjectile`.
-			WorldInfo.MyEmitterPool.OnParticleSystemFinished(Projectiles[ Index ].ParticleEffectComponent);
-			CancelDelayedReturnToPoolPSC(Projectiles[ Index ].ParticleEffectComponent);
+
+			// Start Issue #1201
+			// `EmitterPool.OnParticleSystemFinished` checks that the passed PSC is
+			// contained in the `ActiveComponents` array before doing anything with it,
+			// but not that the `PSC` is not None, in which case the function crashes with
+			// an access violation. It's unclear whether holes in the `ActiveComponents`
+			// can legitimately happen or indicate another bug, so the None check here
+			// is an additional defensive measure.
+
+			if(Projectiles[ Index ].ParticleEffectComponent != None)
+			{
+				WorldInfo.MyEmitterPool.OnParticleSystemFinished(Projectiles[ Index ].ParticleEffectComponent);
+				CancelDelayedReturnToPoolPSC(Projectiles[ Index ].ParticleEffectComponent);
+			}
+			else
+			{
+				`REDSCREEN("CHL: OnParticleSystemFinished called with a None ParticleEffectComponent - this should never happen!" @ `showvar(Projectiles[ Index ].ParticleEffectComponent) @ GetScriptTrace());
+			}
+			// End issue #1201
 			// End Issue #720
 			Projectiles[ Index ].ParticleEffectComponent = none;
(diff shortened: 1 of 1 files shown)


robojumper (changes requested):
I think we should additionally guard all calls to `EmitterPool.OnParticleSystemFinished` with a `!= None` check, and a comment that says something to the effect of:

```
// `EmitterPool.OnParticleSystemFinished` checks that the passed PSC is
// contained in the `ActiveComponents` array before doing anything with it,
// but not that the `PSC` is not None, in which case the function crashes with
// an access violation. It's unclear whether holes in the `ActiveComponents`
// can legitimately happen or indicate another bug, so the None check here
// is an additional defensive measure.
```

perhaps with a `` `REDSCREEN `` that shows when we would otherwise attempt to call `EmitterPool.OnParticleSystemFinished` with a None argument.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
```diff
@@ -1148,6 +1148,15 @@ function OnParticleSystemFinished(ParticleSystemComponent PSystem)
 			CancelDelayedReturnToPoolPSC(Projectiles[i].ParticleEffectComponent);
 			// End Issue #720
 
+			// Start issue #1201
+			/// HL-Docs: ref:Bugfixes; issue:1201
+			/// mark the projectile bConstantComplete as true so that it will not be called again by EndConstantProjectileEffects and cause a crash.
```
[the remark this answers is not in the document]
Tedster59: makes sense,  Should I be documenting it as something else then, or just leaving out the HL-Docs reference all together?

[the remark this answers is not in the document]
Tedster59: Updated to remove the HL-Docs reference

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
```diff
@@ -1140,13 +1157,21 @@ function OnParticleSystemFinished(ParticleSystemComponent PSystem)
...
-			// Start Issue #720
-			/// HL-Docs: ref:ProjectilePerformanceDrain
-			// Allow the pool to reuse this Particle System's spot in the pool.
-			//WorldInfo.MyEmitterPool.OnParticleSystemFinished(Element.ParticleEffectComponent);
-			WorldInfo.MyEmitterPool.OnParticleSystemFinished(Projectiles[i].ParticleEffectComponent);
-			CancelDelayedReturnToPoolPSC(Projectiles[i].ParticleEffectComponent);
-			// End Issue #720
+			// Start issue #1201
+			// Mark the projectile bConstantComplete as true so that it will not be called again by EndConstantProjectileEffects and cause a crash.
+			if (Projectiles[i].ProjectileElement.UseProjectileType == eProjectileType_RangedConstant)
```
Iridar, line 1162: Shouldn't this `if (Projectiles[i].ProjectileElement.UseProjectileType == eProjectileType_RangedConstant)` statement encompass only the `Projectiles[i].bConstantComplete = true;` line? As it stands, adding this if() check will mean `OnParticleSystemFinished()` is never called for any non RangedConstant projectiles.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
```diff
@@ -1140,13 +1157,21 @@ function OnParticleSystemFinished(ParticleSystemComponent PSystem)
...
+			{
+				// Start Issue #720
+				/// HL-Docs: ref:ProjectilePerformanceDrain
+				// Allow the pool to reuse this Particle System's spot in the pool.
+				//WorldInfo.MyEmitterPool.OnParticleSystemFinished(Element.ParticleEffectComponent);
+				WorldInfo.MyEmitterPool.OnParticleSystemFinished(Projectiles[i].ParticleEffectComponent);
```
Iridar, line 1168: Since the crash seemed to be caused by calling `EmitterPool.OnParticleSystemFinished()` on a none-object, shouldn't we just guard every call of this function with none-checks just to be safe?

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
Tedster59: Maybe.  I borrowed that line from Robojumper as is as well as had confirmation from SwfDelicious.

image

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
Tedster59: Doesn't the `if (Projectiles[i].ParticleEffectComponent == PSystem)` on line 1158 catch this? It can't be `none` if it equals `PSystem`

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
Iridar: Definitely:

> There are probably a lot of ways to avoid the crash but I was thinking about something like this:
```
if (Projectiles[i].ProjectileElement.UseProjectileType == eProjectileType_RangedConstant) {
    Projectiles[i].bConstantComplete = true;
}
```
> in that function

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
Iridar: My first instinct was to say "What if `PSystem` is `none` too?"

But it seems this function is only ever called as a delegate which is assigned as:
```
Projectiles[Index].ParticleEffectComponent.OnSystemFinished = OnParticleSystemFinished;
```
So effectively `ParticleEffectComponent` is the PS itself being passed to the delegate, which means it physically can't be none. So I guess it's fine to not do the check. I'd do it anyway though. We lose a only a few CPU cycles on it.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
Iridar: And this probably needs to be done *before* calling OnParticleSystemFinished() on it.

Tedster59: Reordered the if check to only set the bConstantComplete variable

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2UnifiedProjectile.uc
Tedster59: Added the none check

Iridar (approved):

Issue #1325, opened 2024-04-19
Start of initiative interruptions behave too much like start of turn interruptions


TacGit1:
Initiative interruptions are special, occurring when it isn't the unit's turn. In X2TacticalGameRuleset, in the function: SetupUnitActionsForGroupTurnBegin. It calls SetupActionsForBeginTurn from XComGameState_Unit. In SetupActionsForBeginTurn, things like untouchable, bGotFreeFireAction, and, very critically, CleanupUnitValues(eCleanup_BeginTurn) is here. This means if a skirmisher interrupts a turn, but had untouchable, a hair trigger occurrence, or had some eCleanup_BeginTurn unit values on them, they all get cleared as if their real turn has started.

As another example, if a templar had Interrupt and Parry, their Parry gets cleared on interruption.

I have a fix set up that uses unit values, though this will be my first submission to the CHL, so gonna need some help/time doing this properly.

Iridar:
The key problem highlighted by this issue is that initiative interrupts tick unit values, but do not tick effects. This inconsistency means that interrupting initiative is kinda like a real turn from the standpoint of unit values, but not a real turn from the standpoint of effects. 

Description of Interrupt and Battlelord also have no indication that initiative interrupts are meant to be "real turns", and instead just say "take an action". 

As such, I believe unit values ticking during interrupts is a bug. Due to the deeply-rooted nature of initiative interrupts that go through Tactical Ruleset, I believe fixing this bug in Highlander would be appropriate, though in theory doing so might be breaking for existing mods that work tightly with interrupt mechanics. 

This would not only be a bugfix, but also an alteration of base game mechanics, so in this case I would be comfortable with gating a fix for this bug behind a config toggle, and disabled by default.

PR #1376, opened 2024-08-17
Marking Units that Interrupt Initiative

TacGit1:
Fixes Issue #1325 | 3rd time's the charm? Doing this from a non-master branch so I can actually work on other stoof

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
@@ -4461,6 +4461,11 @@ simulated function InterruptInitiativeTurn(XComGameState NewGameState, StateObje
 {
 	local XComGameState_BattleData NewBattleData;
 	local XComGameState_AIGroup InterruptedAIGroup, InterruptingAIGroup;
+	
+	//Start issue #1325 - Variable Declarations
+	local StateObjectReference MemberRef;
+	local XComGameState_Unit Member;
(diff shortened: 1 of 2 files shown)


BlackDog86:
This should reference issue #1325

Iridar:
This PR alters base game behavior, this is allowed only for fixing blatant bugs. What's your reasoning for suggesting that interrupting initiative shouldn't behave like a "real" turn? What are the differences?

TacGit1:
When toying around with things that interrupt initiative, I had a templar who used Parry on their main turn. When they interrupted, their parry was consumed because unit values are cleaned up on _any_ turn start. This means they never actually got to parry. I don't think this is intended because you don't have things like status effects proccing on turn interrupts, and I doubt skirmishers would like having untouchable removed prematurely on interrupt (that ability would have to be granted to them via mods).

Iridar:
So to be clear, effects don't tick on initiative interrupt, but unit values do?

TacGit1:
Yes, went and tested again to make sure. Poison does not tick and the templar I'm testing with loses parry after the interrupt despite having never used it.

Iridar:
Then yeah, I'd say it's fair to call that one a bug.

TacGit1:
Gotcha. Looking forward to your implementation!

TacGit1:
At this time, I will be going to sleep, so I won't be able to respond past this point for a while.

Iridar:
Closing due to lack of participation, tracking issue moved to ~~issue graveyard~~ future major improvements milestone.

PR #1385, opened 2024-09-05
Implement DLC Hook Manager to call DLC methods only from DLC Infos that implement them

### LINKED ISSUE #212: Improve performance on DLC Hooks (X2DownloadableContentInfo)

Iridar:
Fixes #212

Variant of #1247 with fixed merge conflicts.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHDLCHookManager.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/CHDLCHookManager.uc
@@ -1,0 +1,195 @@
+/// HL-Docs: feature:CHDLCHookManager; issue:212; tags:
+/// The CHDLCHookManager class is an internal component of CHL that allows us 
+/// to retrieve only DLCs (X2DownloadableContentInfo classes) that override a 
+/// specific function or event by method name.
+///
+/// It was introduced to optimize the number of calls made on X2DownloadableContentInfo
+/// classes. Instead of CHL calling X2DLCI methods on every known DLC class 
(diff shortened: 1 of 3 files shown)


Iridar:
robojumper reviewed and approved the original PR, I just fixed merge conflicts by dropping new files on top of CHL's current master branch.

PR #1387, opened 2024-09-08
Cache hit chance when sorting the list of potential targets in tactical and ensure the sorted order matches the order of TAB-cycling

### LINKED ISSUE #1233: UITacticalHUD_Enemies UpdateVisibleEnemies should use cached hit chances when sorting final targets array

Iridar:
Fixes #1233
Alternative implementation of #1234 

Differences: moved the sorting into a separate function, added some code to fix the "bug" where the order of TAB-cycling through targets does not always the match the order of displayed target icons for targets that have the same hit chance. 

I separated them into two commits so that the TAB-cycling functionality is easier to see during the review.

I added HL-Docs for this issue as if it's a bugfix, not sure how correct is that, but I figured there's nothing about this change that needs to be mod-dev facing, as it's mostly an optimization change.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Enemies.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UITacticalHUD_Enemies.uc
@@ -16,6 +16,7 @@ struct CH_CachedTargetHitChanceStruct
 	var StateObjectReference TargetRef;
 	var bool bIsDestructibleObject;
 	var int HitChance;
+	var int AvailableTargetIndex;
 };
 // End Issue #1233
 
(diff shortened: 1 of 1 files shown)


Iridar:
Tested in lab conditions, tutorial, and first mission afterwards. Seems to work fine.

Iridar:
This needs testing with various targetable interactive objects, such as explosive barrels and psi nodes on the Sabotage Psi Emitter mission. Make sure the order of displayed icons matches the order of clicking on the icon or using an ability to enter targeting.

BlackDog86:
I can have a go at checking this out

Iridar:
Would be appreciated

BlackDog86:
OK so this all looks good to me - I tested with the psionic transmitter nodes, an alien relay and a couple other destructables - each time the destructables were sorted to the end of the the targetting array & I didn't get any instances where clicking on a target took me to the wrong unit or anything weird like that. Even the hacking preview appears to keep the units in the correct hit chance order (obviously excluding units that are irrelevant for hacking) so I'm pretty confident this is working. I can't really think of many edge-case destructables that might not behave in the same way as the rest either but if you can think of any more I'm happy to give them a trial before we sign this one off.

Iridar:
Thank you, I think this is good to go.

PR #1388, opened 2024-09-09
Prevent Low Intel warnings if all regions are contacted

### LINKED ISSUE #1347: Persistent "Low Intel" warning after contacting all regions

Iridar:
Fixes #1347

Alternative implementation of #1346, which addressed specifically the Low Intel popup, but not Low Intel icon in the lower left corner of the Geoscape, and probably more UI elements that make use of `class'UIUtilities_Strategy'.static.GetMinimumContactCost()` function.

I tested and it works as intended at least for the post-mission popup and the Geoscape icon.


BlackDog86:
Let me know if you'd like this PR tested on my side as well

Iridar:
Yeah that would be appreciated, not thrilled about merging submissions without at least some kind of peer review.

Issue #1389, opened 2024-09-11
ChosenBrutal restores will


boundir:
Brutal uses `GetBaseStat()` instead of `GetCurrentStat()`. The result in most occasions is a will gain.

Affected methods:
`X2Effect_Brutal::OnEffectAdded`
`X2Effect_Brutal::AddX2ActionsForVisualization`

BlackDog86:
Will need to split the commit in two if adding new files - one adding the file and the other documenting the changes.

PR #1390, opened 2024-09-11
Fixes issue #1389

boundir:
Make use of `GetCurrentStat()` instead of `GetBaseStat()`

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Brutal.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Brutal.uc
@@ -1,0 +1,42 @@
+class X2Effect_Brutal extends X2Effect;
+
+var localized string WillChangeMessage;
+var int WillMod;
+
+simulated protected function OnEffectAdded(const out EffectAppliedData ApplyEffectParameters, XComGameState_BaseObject kNewTargetState, XComGameState NewGameState, XComGameState_Effect NewEffectState)
+{
+	local XComGameState_Unit TargetUnit;
+	local int CurrentWill;
+
+	TargetUnit = XComGameState_Unit(kNewTargetState);
+	if (TargetUnit != none)
+	{
+		CurrentWill = TargetUnit.GetCurrentStat(eStat_Will); // issue 1389
+		if (CurrentWill + WillMod <= 0)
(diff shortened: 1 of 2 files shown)


chrishayesmu:
I don't do HL PRs much, but I believe the standard is to add new files in one commit, then apply HL changes in a separate commit (in the same PR). This makes it much easier to tell what you've changed while reviewing.

boundir:
> I don't do HL PRs much, but I believe the standard is to add new files in one commit, then apply HL changes in a separate commit (in the same PR). This makes it much easier to tell what you've changed while reviewing.

Now that you mention it it's exactly what I should have done. I will redo the thing tomorrow then.

Iridar:
I'm afraid messing with this can cause the same issues as my old version of System Infiltration, where it was meant to temporarily buff unit's Hacking stat, but ended up permanently applying temporary hacking bonus to the unit, which is what the current version of the X2Effect_Brutal is seemingly meant to do - apply permanent changes to unit's Will.

I'm probably wrong and `SetCurrentStat()` doesn't cause permanent changes, but right now the effect does seem weird, though looking from code, it seems like it would immediately make the unit lose benefits from Will bonuses from other sources?

Seems like a typical Firaxis-style mess. If the effects from Brutal are not meant to be temporary, then the ability should be patched to use X2Effect_PersistentStatChange instead. 

If the effects are meant to be permanent, then I'm pretty sure the fix is changing `SetCurrentStat()` into `SetBaseMaxStat()`.

faanlez:
It was originally permanent loss of max Will, got changed in a patch.

https://store.steampowered.com/news/app/268500/view/2872689832680249387?l=english

Iridar:
I see. In that case I suggest dumping the original effect and making the ability use a stacking X2Effect_PersistentStatChange.

Mitzruti:
A persistent stat change would only modify max will, which could lead to the weird sort of situation where soldiers actually recover faster after fighting a brutal chosen. (due to will loss from seeing enemies/etc being capped by max will)

what brutal is supposed to be doing is basically damage to will, so they're more likely to wind up shaken afterwards.

Iridar:
Hmm, makes sense.

boundir:
Replaced by https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/pull/1392

PR #1392, opened 2024-09-12
Fix X2Effect_Brutal by using GetCurrentStat() instead of GetBaseStat()

### LINKED ISSUE #1389: ChosenBrutal restores will

boundir:
Fixes #1389

Fixes Brutal effect by using GetCurrentStat() instead of GetBaseStat().

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Brutal.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Brutal.uc
@@ -1,0 +1,42 @@
+class X2Effect_Brutal extends X2Effect;
+
+var localized string WillChangeMessage;
+var int WillMod;
+
+simulated protected function OnEffectAdded(const out EffectAppliedData ApplyEffectParameters, XComGameState_BaseObject kNewTargetState, XComGameState NewGameState, XComGameState_Effect NewEffectState)
+{
+	local XComGameState_Unit TargetUnit;
+	local int CurrentWill;
+
+	TargetUnit = XComGameState_Unit(kNewTargetState);
+	if (TargetUnit != none)
+	{
+		CurrentWill = TargetUnit.GetBaseStat(eStat_Will);
+		if (CurrentWill + WillMod <= 0)
+			TargetUnit.SetCurrentStat(eStat_Will, 1);
+		else
+			TargetUnit.SetCurrentStat(eStat_Will, CurrentWill + WillMod);
+	}
+}
+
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Brutal.uc
```diff
@@ -11,7 +11,7 @@ simulated protected function OnEffectAdded(const out EffectAppliedData ApplyEffe
 	TargetUnit = XComGameState_Unit(kNewTargetState);
 	if (TargetUnit != none)
 	{
-		CurrentWill = TargetUnit.GetBaseStat(eStat_Will);
+		CurrentWill = TargetUnit.GetCurrentStat(eStat_Will); // issue 1389
```
Iridar, line 14: All Highlander submissions must follow the same comment convention. 

For single line change, you would use `// Single line for Issue #1389`, and for anything more than one line, you would use `// Start Issue #1389` and `// End Issue #1389`. Then there are Highlander inline docs, which get compiled automatically into this webpage: https://x2communitycore.github.io/X2WOTCCommunityHighlander/

They use a single `/// HL-Docs: ` entry, and there are two types of those, one is used to document _features_, so they can be easier used by mod authors, and other uses to document _bugfixes_, and these are meant for mod users, and all get compiled into a single page: https://x2communitycore.github.io/X2WOTCCommunityHighlander/Bugfixes/

In case of single line bugfixes, it is acceptable to use a HL-Docs entry instead of "single line for issue". 

Only one HL-Docs entry per issue is needed normally.

In this case comments would look like:

```cpp
/// HL-Docs: ref:Bugfixes; issue:1389
/// Make `X2Effect_Brutal` use unit's current Will instead of base Will so it can properly reduce it.
// Single line for Issue #1389 - use GetCurrentStat() instead of GetBaseStat().
CurrentWill = TargetUnit.GetCurrentStat(eStat_Will);
```
and
```cpp
// Single line for Issue #1389 - use GetCurrentStat() instead of GetBaseStat().
WillChange = NewUnit.GetCurrentStat(eStat_Will) - OldUnit.GetCurrentStat(eStat_Will);
```

Iridar (changes requested):

Iridar:
Perfect, though a bit annoyingly I have to ask you to squish the last two commits together, so there's still one commit with adding new files, and one commit where the issue gets actually resolved. This can be done via `$git rebase -i HEAD~2` git command.

PR #1401, opened 2024-10-29
Allow forcing 24h clock and leading zeroes on month and date strings

### LINKED ISSUE #1400: Allow mods to force displaying the 24h clock and allow appending zeroes to months and days for display consistency.
X2StrategyGameRulesetDataStructures contains some unusual logic for when the game should display the 24h clock (GetTimeStringSeparated function) - if the language is INT or ESN, then the 12h clock is preferred for some strange reason. In one of my mods I currently over-ride X2StrategyGameRulesetDataStructures to fix this minor issue but it would be better if mods could just set this via a config variable.

Additionally, it has always bugged me that dates are displayed like this:

9/6/2024
13/6/2024 

(Different string lengths)

Rather than something like:
09/06/2024
or better, in ISO standard:
2024-06-09

By creating a couple of leading-zero-appended strings in GetDateString, mods can adjust the display of these dates just by using localisation without having to interfere with any base-game functionality and/or MCO this relatively important class.

BlackDog86:
Fixes #1400

### DIFF
--- a/X2WOTCCommunityHighlander/Config/XComGameData.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGameData.ini
@@ -6,3 +6,8 @@
 ; uncomment for Beta Strike
 ;+DEBUG_SecondWaveOptions="BetaStrike"
 ;;;
+
+[XComGame.X2StrategyGameRulesetDataStructures]
+; Issue #1400 - Uncomment to force 24h clock and/or addition of a leading zero to the hours (e.g. to display 03:45 instead of 3:45 on the geoscape)
+;bForce24hClock = true
+;bForce24hClockLeadingZero = true
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
@@ -1084,6 +1084,9 @@ var config int START_DAY;
 var config int START_MONTH;
 var config int START_YEAR;
 
+// Single Line for Issue #1400
+var config bool bForce24hClock;
+var config bool bForce24hclockLeadingZero;
 // Intro movie narrative moment
 var config string IntroMovie;
 
@@ -1567,9 +1570,12 @@ static function GetTimeStringSeparated(TDateTime kDateTime, out string Hours, ou
 	Lang = GetLanguage();
 	iHour = GetHour(kDateTime);
 
+	/// HL-Docs: feature:DateAndTimeFormattingChanges; issue:1400; tags:
+	/// Allow forcing the 24h clock independently of locale and add an option to display leading zeroes for military-style time - e.g. 03:41 instead of 3:41
+
 	// INT and ESN use the 12 hour clock for events, checked with Loc 12/15/2015. -bsteiner 
-	if( Lang == "INT" || Lang == "ESN" )
-	{
+	if( !default.bForce24hclock && (Lang == "INT" || Lang == "ESN"))
+	{		
 		// AM
 		if( iHour < 12 )
 		{
@@ -1602,7 +1608,15 @@ static function GetTimeStringSeparated(TDateTime kDateTime, out string Hours, ou
 		Minutes = string(GetMinute(kDateTime));
 	}
 
-	Hours = string(iHour);
+	if( default.bForce24hClockLeadingZero && GetHour(kDateTime) < 10 )
+	{
+		Hours = "0"$GetHour(kDateTime);
+	}
+	else
+	{
+		Hours = string(iHour);
+	}
+	// End Issue #1400
 }
 
 static function string GetDateString(TDateTime kDateTime, optional bool bShortFormat = false)
@@ -1620,6 +1634,26 @@ static function string GetDateString(TDateTime kDateTime, optional bool bShortFo
 		kTag.IntValue0 = kDateTime.m_iDay;
 		kTag.IntValue1 = kDateTime.m_iYear;
 		kTag.IntValue2 = kDateTime.m_iMonth;
+
+		// Issue #1400 - Add additional localization tags to allow mods to display days and months with leading zeroes
+		if(kDateTime.m_iDay < 10)
+		{
+			kTag.StrValue2 = 0 $ kDateTime.m_iDay;
+		}
+		else
+		{
+			kTag.StrValue2 = string(kDateTime.m_iDay);
+		}
+		
+		if(kDateTime.m_iMonth < 10)
+		{
+			kTag.StrValue3 = 0 $ kDateTime.m_iMonth;
+		}
+		else
+		{
+			kTag.StrValue3 = string(kDateTime.m_iMonth);
+		} 
+		// End issue #1400
 
 		Lang = GetLanguage();
 


X2WOTCCommunityHighlander/Config/XComGameData.ini
```diff
@@ -6,3 +6,9 @@
 ; uncomment for Beta Strike
 ;+DEBUG_SecondWaveOptions="BetaStrike"
 ;;;
+
+[XComGame.X2StrategyGameRulesetDataStructures]
+; Begin Issue #1400
+;;; HL-Docs: ref:DateAndTimeFormattingChanges
+; uncomment to force 24h clock
+;bForce24hclock = true
```
Iridar, line 14: Missing an "end Issue" comment tag, could probably just replace it with a single tag with issue number and following commend, something like:

; Issue #1400 - remove the semicolon at the start of the line to force 24H clock.

Also missing a new line at the end of the file.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
```diff
@@ -1621,6 +1629,26 @@ static function string GetDateString(TDateTime kDateTime, optional bool bShortFo
 		kTag.IntValue1 = kDateTime.m_iYear;
 		kTag.IntValue2 = kDateTime.m_iMonth;
 
+		// Issue #1400 - Add additional localization tags to allow mods to display days and months with leading zeroes
+		if(kDateTime.m_iDay < 10)
+		{
+		kTag.StrValue2 = 0 $ kDateTime.m_iDay;
```
Iridar, line 1635: Please use proper indentation.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
```diff
@@ -1563,13 +1565,19 @@ static function GetTimeStringSeparated(TDateTime kDateTime, out string Hours, ou
 {
 	local int iHour;
 	local string Lang;
+	
+	/// HL-Docs: feature:DateAndTimeFormattingChanges; issue:1400; tags:
+	/// Add code to allow forcing the 24h clock independently of locale
+	local bool use24hclock;
```
Iridar, line 1571: Is this supposed to be `use12hclock`? Does this code even compile? I'm not sure why the additional variable is even necessary.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
```diff
@@ -1621,6 +1629,26 @@ static function string GetDateString(TDateTime kDateTime, optional bool bShortFo
 		kTag.IntValue1 = kDateTime.m_iYear;
 		kTag.IntValue2 = kDateTime.m_iMonth;
 
+		// Issue #1400 - Add additional localization tags to allow mods to display days and months with leading zeroes
```
Iridar, line 1632: This seems like an opinionated change, it should be at least gated behind a separate config var.

Iridar (changes requested):

X2WOTCCommunityHighlander/Config/XComGameData.ini
BlackDog86: Yep OK - done

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
BlackDog86: Fixed

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
BlackDog86: Yeah, that was a mistake, apologies - I've removed the var altogether now and just used !default.bForce24hClock directly in the expression.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
BlackDog86: this isn't really a change at all - the default localization file uses this for the display of dates:
`m_strMonthDayYearShort="<XGParam:IntValue2/><XGParam:StrValue1/!SeparatorString/><XGParam:IntValue0/><XGParam:StrValue1/!SeparatorString/><XGParam:IntValue1/>"` This would not change at all with the code adjustment. A mod would have to specifically remove the localization line and replace it with <XGParam:StrValue2> or <XGParam:StrValue3> explicitly in order for this to adjust base-game behaviour at all so I don't think a config gate is necessary.

Iridar: Does this have to be a Highlander change then? It this does nothing without mod-supplied localization lines, what's stopping the mod that is adding those lines from adding its own localized tags and using the tag expand handler to fill them with whatever date format it wants?

BlackDog86: Because the mod would need to MCO X2StrategyGameRulesetDataStrcutures just to create a couple of tags which is quite an invasive MCO for a very small change (plus there are already some mods existing MCOing other parts of this, creating conflicts. I thought this sort of thing is more or less exactly what CHL is for?)

Iridar: I don't understand why would MCO be necessary. You can create arbitrary tags and fill them with arbitrary content via the DLC Hook, no MCO required. Mods can always ship their own adjusted version of `GetTimeStringSeparated()` to use in their tags.

BlackDog86:
After a bit of further investigation & for the purposes of clarity, the only place that the 'short' date is actually called for in the base game, is in the memorial screen. However, other mods (such as beat's mission history) also call for it. See below some examples of the display options:
Memorial before:
image
Memorial after, swapping the short date tags:
image

Clock example, base game:
image
Clock example with bforce24hclock and bforce24hclockleadingzero on:
image
Clock example with bforce24hclock and bforce24hclockleadingzero on and using tag swapping to swap the long dates as well:
image

I should say that the save and load game screens do not use this function and have their own logic for parsing the save file headers to display the dates of the save games. This logic isn't great either in my view and in one of my own mods I MCO a few of the classes responsible in order to display those dates differently (https://steamcommunity.com/sharedfiles/filedetails/?id=3268722967). This is the output of that mod, for example:
image. The base-game code is a bit janky in that area as well but making that better is out of scope for this, which is just making a few small changes for the sake of flexibility.

BlackDog86:
Hmm something weird happened in testing. Don't merge this yet please - need to check something!

Iridar:
For context to the above discussion, I didn't understand that the following pipeline was in place:

1. Different parts of the code call the `GetDateString()` function
2. Which takes one of the two existing strings that exist solely for the purpose of arranging date
3. Expands its tags
4. Returns the formatted date to whatever code caled the function

So I didn't understand why changes to `GetDateString()` were necessary.

Issue #359, opened 2018-12-08
Extend enumerated type Photobooth_ParticleEffectType to allow us to attach more particle systems to weapons


E3245:
Right now the enumerated type only contains 3 variables ( + 1 none variable):
```enum Photobooth_ParticleEffectType
{
    ePPET_None,
    ePPET_TemplarBladeRight,
    ePPET_TemplarBladeLeft,
    ePPET_TemplarShield
};
```

Would it be possible to add 5 - 20 more variables to this so we can use more Particle Systems attached to weapons and other accessories?

PR #1403, opened 2024-11-02
Docs - Photobooth Enums

BlackDog86:
Adds docs for issue #359

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Photobooth.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Photobooth.uc
@@ -65,12 +65,14 @@ enum Photobooth_ParticleEffectType
 	ePPET_TemplarShield,
 	// Start Issue #359
 	/// HL-Docs: feature:PhotoboothParticleEffectEnums; issue:359; tags:strategy
-	/// This feature was intended to provide additional enum values for Photobooth Particle effects.
-	/// It is currently unused, with the original idea being that modders would submit a PR with updated 
-	/// enum names first, and these could then be used to when defining particle systems for the photobooth in 
-	/// config later. Typically this would be done in XComContent.ini by specifying ParticleEffectTypes=ePPET_NAME
-	/// within an m_arrAnimationPoses entry and ParticleEffectType = ePPET_NAME within a ParticleEffects entry
-	/// in the corresponding arrays.
+	/// This feature provides additional enum values for Photobooth particle effects. To use it,
+	/// modmakers are required to submit a Pull Request where they will add new enum names, 
+	/// which can then be used when defining particle effects for the Photobooth in `XComContent.ini`
+	/// by specifying `ParticleEffectTypes = ePPET_NAME` within an `m_arrAnimationPoses` entry,
+	/// and `ParticleEffectType = ePPET_NAME` within a `ParticleEffects` entry.
+	/// The enum value will then be used by the game to connect the two entries.
+	/// Note that it's also possible to add particle effects to Photobooth poses by adding them
+	/// directly to AnimSequences via PlayParticleEffect AnimNotifies, like TLP poses do.
 	ePPET_Buffer04,
 	ePPET_Buffer05,
 	ePPET_Buffer06,
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Photobooth.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Photobooth.uc
@@ -64,9 +64,13 @@ enum Photobooth_ParticleEffectType
 	ePPET_TemplarBladeLeft,
 	ePPET_TemplarShield,
 	// Start Issue #359
-	// Similar to above, but we don't need any sort of special marker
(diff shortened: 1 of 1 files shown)


Iridar (commented):
Appreciate the efforts, but at the moment docs don't make it clear how to make use of this feature. I assume it expects mod authors to submit their own PRs that would add more members to the enum so then their mods can take advantage of the new values? TBH I have a hard understanding why this feature is necessary at all.

Iridar:
So based on some digging these enums are used in the ParticleEffects config array to add particle effects to photobooth poses via config, but like why do that when poses can be added to pose animations directly via animnotifies like TLE poses do?

BlackDog86:
OK so - as I understand it, m_arrAnimationPoses is a config array of AnimationPoses structs with the following stuff inside:
struct native AnimationPoses
```
{
	var localized string							AnimationDisplayName;
	var name									AnimationName;
	var float									AnimationOffset;
	var Photobooth_AnimationFilterType				AnimType;
	var array<Photobooth_ParticleEffectType>			ParticleEffectTypes; <<<----- this is the array of enums
	var bool									bExcludeFromGroupShots;
	var bool									bExcludeFromTemplar;
};
```

By convention, most people add new poses by adding new lines to XComContent.ini like this:
`+m_arrAnimationPoses=(AnimationName="AS_Cal_Pinup_16",AnimationOffset=0.25,AnimType=ePAFT_None)` (this one from Kexx calendar pack).

Presumably new poses can also specify particle effects, like this:
`+m_arrAnimationPoses=(AnimationName="Photobooth_TemplarPoses",AnimationOffset=1.25,AnimType=ePAFT_Templar,ParticleEffectTypes=(ePPET_TemplarBladeRight,ePPET_TemplarBladeLeft))`

Meanwhile, there is a config array of ParticleEffects which can be specified like:
`+ParticleEffects=(PSTemplateName="TLE_FX_Photobooth.P_Deflect_Shield_Looping",SocketName="CIN_Root",ParticleEffectType=ePPET_TemplarShield)`

From what I can tell, the code in SetParticleEffects seems to goes through each ParticleEffectTypes in the array of AnimationPoses and checks those against each element of the seperate ParticleEffects array before playing them, with the link between the two being this ePPET_<STUFF> enum which has to be specified in both arrays. So presumably, if one wants to add an animpose with a custom particle effect just using config, it would have to use one of the other not-already-defined enums otherwise the SetParticleEfffects code would prevent the effect from playing, maybe?

Iridar:
Sounds like if mod authors want to add new poses using this feature, they have to first submit a CHL PR to add new enums. Ugh, this is awful, how did this even get approved. I suppose at this point we just have to support it, but in the presence of a much simpler method (adding PFX directly to pose animsequences like TLP poses do), I don't see anybody ever using it.

BlackDog86:
Yeah I don't know of any mods actively using the feature to be fair. It does seem like a very awkward way of doing it - the code should probably be refactored so that a single animpose config array including all the particle effects can be specified but yeah, maybe I should never have dug this up in the first place!

Iridar:
> OK so - as I understand it, m_arrAnimationPoses is a config array of AnimationPoses structs with the following stuff inside: struct native AnimationPoses
> 
> ```
> {
> 	var localized string							AnimationDisplayName;
> 	var name									AnimationName;
> 	var float									AnimationOffset;
> 	var Photobooth_AnimationFilterType				AnimType;
> 	var array<Photobooth_ParticleEffectType>			ParticleEffectTypes; <<<----- this is the array of enums
> 	var bool									bExcludeFromGroupShots;
> 	var bool									bExcludeFromTemplar;
> };
> ```
> 
> By convention, most people add new poses by adding new lines to XComContent.ini like this: `+m_arrAnimationPoses=(AnimationName="AS_Cal_Pinup_16",AnimationOffset=0.25,AnimType=ePAFT_None)` (this one from Kexx calendar pack).
> 
> Presumably new poses can also specify particle effects, like this: `+m_arrAnimationPoses=(AnimationName="Photobooth_TemplarPoses",AnimationOffset=1.25,AnimType=ePAFT_Templar,ParticleEffectTypes=(ePPET_TemplarBladeRight,ePPET_TemplarBladeLeft))`
> 
> Meanwhile, there is a config array of ParticleEffects which can be specified like: `+ParticleEffects=(PSTemplateName="TLE_FX_Photobooth.P_Deflect_Shield_Looping",SocketName="CIN_Root",ParticleEffectType=ePPET_TemplarShield)`
> 
> From what I can tell, the code in SetParticleEffects seems to goes through each ParticleEffectTypes in the array of AnimationPoses and checks those against each element of the seperate ParticleEffects array before playing them, with the link between the two being this ePPET_ enum which has to be specified in both arrays. So presumably, if one wants to add an animpose with a custom particle effect just using config, it would have to use one of the other not-already-defined enums otherwise the SetParticleEfffects code would prevent the effect from playing, maybe?

Well, since you already started, would you mind laying out some of that in the docs? The short version.

Iridar:
I've reworded the docs a bit, please take a look and if you don't see anything objectionable, this is good to go.

BlackDog86:
Fine by me :)

Issue #1404, opened 2024-11-03
X2Effect_Sustain does not set the unit's LowestHP when triggered


Tedster59:
In `PreDeathCheck`, X2Effect_Sustain sets the current HP to 1, but does not set the lowest HP to 1 as well.

![image](https://github.com/user-attachments/assets/e46a7a8a-7ad7-43ab-8d9a-4299e24e2b9d)

If the unit sustains, it short circuits the `TakeDamage` function in XCGS_Unit (If statement at line 6682 here) which normally handles updating that value (If statement at line 6708).

![image](https://github.com/user-attachments/assets/7ee811cd-e11a-4554-bb78-feed81f05382)

Thus, if a unit is hit to sustain without taking previous damage, you get after battle such as this, since the unit's "LowestHP" value is still set to the max HP, and the game doesn't even register the unit as wounded and you can get a flawless rating despite the unit being hit into Sustain and taking a wound.

![image](https://github.com/user-attachments/assets/1eaa4b61-b1c6-469b-8360-bf0809942198)

This affects both the Sustain ability and the Sustaining Sphere item.

PR #1405, opened 2024-11-03
Set a unit's lowest HP value to 1 if they activate sustain

Tedster59:
Fixes issue #1404.

LW also needed to do this for its Emergency Life Support PCS in it's own PreDeathCheck function.

image


Iridar:
Looks good to me. "since it wasn't being set properly before." part of the comment isn't really necessary but not gonna ask to redo the PR just because of that.

Issue #1406, opened 2024-11-12
Add a new Event Trigger after Mission Intro kismet has completed


furudee:
I suggest adding a new event trigger after Mission Intro kismet has run it's course, just before players enters their first XCOM turn

Reasoning: 
missions such as Supply Extraction spawns objective crates and submits new GameStates when the OnTacticalMissionStartBlocking kismet SeqEvent is triggered,
which happens after OnBeginTacticalPlay event trigger, so you don't have a handle for anything that happens in the kismet event, not before your first turn.

I believe it's better to place it after OnTacticalMissionStartBlocking event, not OnTacticalMissionStartNonBlocking,
because the former is supposedly latent thus making sure everything has been initialized correctly, no?

The following code is in X2TacticalGameRuleset, in PostCreateTacticalGame state:
```
Begin:
	`SETLOC("Start of Begin Block");

	//Posts a game state updating all state objects that they are in play
	BeginTacticalPlay();

	//Let kismet know everything is ready
	`XCOMGRI.DoRemoteEvent('XGBattle_Running_BeginBlockFinished');

	`CURSOR.ResetPlotValueCache( );

	CachedHistory.CheckNoPendingGameStates();

	// immediately before running kismet, allow sitreps to modify values on the kismet canvas
	PostCreateTacticalModifyKismetVariables( );
	class'X2SitRepEffect_ModifyKismetVariable'.static.ModifyKismetVariables();
	SecondWaveModifyKismetVariables( );

	// kick off the mission intro kismet, and wait for it to complete all latent actions
	WorldInfo.TriggerGlobalEventClass(class'SeqEvent_OnTacticalMissionStartBlocking', WorldInfo);


	while (WaitingForVisualizer())
	{
		sleep(0.0);
	}
         ////////////////////////////////////////////////////
         SUGGEST PUTTING EVENT TRIGGER HERE
         ////////////////////////////////////////////////////

	// set up start of match special conditions, including squad concealment
	ApplyStartOfMatchConditions();

         ////////////////////////////////////////////////////
         OR HERE
         ////////////////////////////////////////////////////

	// kick off the gameplay start kismet, do not wait for it to complete latent actions
	WorldInfo.TriggerGlobalEventClass(class'SeqEvent_OnTacticalMissionStartNonBlocking', WorldInfo);

	StartStateCheckAndBattleDataCleanup();

	// give the player a few seconds to get adjusted to the level
	Sleep(0.5f);

[...]

PR #1407, opened 2024-11-12
Add a new event trigger after Mission Intro kismet has completed 

furudee:
Fixes issue #1406

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
@@ -3893,6 +3893,24 @@ Begin:
 		sleep(0.0);
 	}
 
+	// Start Issue #1406
+	/// HL-Docs: feature:PostMissionIntroKismet; issue:1406; tags:tactical
+	/// This event triggers after mission intro kismet has been completed.
+	/// For example, a mission such as Supply Extraction spawns their objective crates
+	/// when `'SeqEvent_OnTacticalMissionStartBlocking'` has been triggered,
+	/// which happens after `'OnBeginTacticalPlay'` event has already been triggered.
+	/// As such, you don't have a handle for anything the mission kismet causes before your first XCOM turn.
+	/// This event can be considered a later alternative to `'OnBeginTacticalPlay'` event.
+	///
+	/// ```event
+	/// EventID: PostMissionIntroKismet,
+	/// EventData: none,
+	/// EventSource: none,
+	/// NewGameState: none
+	/// ```
+	`XEVENTMGR.TriggerEvent('PostMissionIntroKismet',,,);
+	// End Issue #1406
+
 	// set up start of match special conditions, including squad concealment
 	ApplyStartOfMatchConditions();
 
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
```diff
@@ -4245,6 +4259,11 @@ Begin:
 		sleep(0.0);
 	}
 
+	// Start Issue #1406
+	/// HL-Docs: ref:PostMissionIntroKismet
```
Iridar, line 4263: Don't need a HL-Docs entry here, just the standard issue tags are enough. Could even do `// Single Line for Issue #1406`.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
```diff
@@ -3887,12 +3887,26 @@ Begin:
...
+	// Start Issue #1406
+	/// HL-Docs: feature:PostMissionIntroKismet; issue:1406; tags:tactical
+	/// This event triggers after mission intro kismet has been completed.
+	/// For example, a mission such as Supply Extraction spawns their objective crates
+	/// when `'SeqEvent_OnTacticalMissionStartBlocking'` has been triggered,
+	/// which happens after `'OnBeginTacticalPlay'` event has already been triggered.
+	/// As such, you don't have a handle for anything the mission kismet causes before your first XCOM turn.
+	/// This event can be considered a later alternative to `'OnBeginTacticalPlay'` event.
+	///
+	/// ```unrealscript
```
Iridar, line 3904: HL-Docs for events have a different style, here's an example:

```
/// ```event
/// EventID: AllowOnCovertActionCompleteAnalytics,
/// EventData: [
///   inout bool bAllow,
///   in Object OriginalEventData,
///   in Object OriginalEventSource,
///   in XComGameState OriginalGameState,
///   in name OriginalEvent,
///   in Object OriginalCallbackData
/// ],
/// EventSource: none,
/// NewGameState: none
```

Also, are you sure there's no useful stuff you could pass along with the event trigger?

Iridar (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
furudee: I figured that the reference would be useful, since there are two different places where the trigger is placed and both will show the location on the documentation

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
furudee: I'm not sure there is anything special to pass with the trigger, since there is no pending gamestate and you can access the ruleset normally when the event trigger is called

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TacticalGameRuleset.uc
Iridar: That's fair. Good work.

Iridar (approved):

Iridar:
This is ready to go. I'll merge it when the work on 1.30 starts in earnest. Thanks for your contribution.

Iridar:
Btw it wasn't mandatory to squash the commits, when you're not adding new files, all commits can be squashed automatically during the merging process, we require exactly two commits only when new files are added to CHL.

Issue #1408, opened 2024-11-14
X2TargetingMethod_ArcWave doesn't validate its AdditionalTargets it gets as parameter


furudee:
The game assigns any additional targets an ability might hit, I believe in class X2AbilityMultiTargetStyle, native function GetMultiTargetOptions.

The issue is that it seems to incorrectly assign these additional targets when using an ability that uses class X2TargetingMethod_ArcWave as its Targeting Method. 
In a function GetAdditionalTargets, X2TargetingMethod_ArcWave only adds more targets and validates them, but doesn't validate the existing ones in the function parameter it's given. 
Other targeting methods that make use of the function seem to validate the targets.

I suggest adding a validation loop in ArcWave targeting method, that checks all targets being hit are also on the tiles being hit.
I attempted to use an existing native function that other targeting methods use for validation, but it returned incorrect results.

I'm not aware of any other bug/ability that makes an ability hit more/less targets than intended, so I think just adding a validation is enough in this case

![Untitled5](https://github.com/user-attachments/assets/7736a555-c635-41e1-a015-de3fd4dd8a69)
Example: ability has been assigned 2 additional targets instead of the intended amount of 1. This seems to be dependent on the position where the unit initiates the ability. Initiating from left or right and doing a 90 degree attack gives wrong amount of targets, but a straight on attack gives correct amount.

Tedster59:
This was fixed by #1409 but this wasn't closed, so closing.

PR #1409, opened 2024-11-14
Add validation for X2TargetingMethod_ArcWave:GetAdditionalTargets

furudee:
Issue #1408

Needs testing with other mod abilities to make sure nothing breaks

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TargetingMethod_ArcWave.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TargetingMethod_ArcWave.uc
@@ -1,0 +1,163 @@
+class X2TargetingMethod_ArcWave extends X2TargetingMethod_MeleePath;
+
+var X2AbilityTemplate ArcWaveTemplate;
+var TTile LastDestination;
+
+function Init(AvailableAction InAction, int NewTargetIndex)
+{
+	ArcWaveTemplate = class'X2AbilityTemplateManager'.static.GetAbilityTemplateManager().FindAbilityTemplate('ArcWave');
+
+	super.Init(InAction, NewTargetIndex);	
+}
+
+function DirectSetTarget(int TargetIndex)
+{
+	local Actor TargetedActor;
+	local array<TTile> Tiles;
+	local TTile TargetedActorTile;
+	local XGUnit TargetedPawn;
+	local vector TargetedLocation;
+	local XComWorldData World;
+
+	super.DirectSetTarget(TargetIndex);
+
+	TargetedActor = GetTargetedActor();
+	TargetedPawn = XGUnit(TargetedActor);
+	if (TargetedPawn != none)
+	{
+		World = `XWORLD;
+		TargetedLocation = TargetedPawn.GetFootLocation();
+		TargetedActorTile = World.GetTileCoordinatesFromPosition(TargetedLocation);
+		TargetedLocation = World.GetPositionFromTileCoordinates(TargetedActorTile);
+		LastDestination = TargetedActorTile;
+
+		ArcWaveTemplate.AbilityMultiTargetStyle.GetValidTilesForLocation(Ability, TargetedLocation, Tiles);
+
+		if (Tiles.Length > 1)
+		{
+			DrawAOETiles(Tiles);
+		}
+	}
+}
+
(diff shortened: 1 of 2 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TargetingMethod_ArcWave.uc
```diff
@@ -137,22 +138,47 @@ function bool GetAdditionalTargets(out AvailableTarget AdditionalTargets)
...
+						}
+
+						ValidationTargets.AddItem(StateObject.GetReference());
 					}
 				}
 			}
+			// Start Issue #1408
+			/// HL-Docs ref:Bugfixes; issue:1408
+			/// Validates targets that were given as function parameter
+			for (i = AdditionalTargets.AdditionalTargets.Length - 1; i >= 0; --i)
```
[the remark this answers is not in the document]
furudee: Yes, that would be smarter I believe. I made it a loop due to the "validation" concept.

Now that I looked at how to implement this again, I figured that emptying the additional target array of the function parameter at the start might be the easiest solution, since the function will fill out the array with correct targets anyway.
Unless I'm missing something that can have an ability have any other additional targets or secondary effects that are not related to the tiles being hit by the ability

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TargetingMethod_ArcWave.uc
```diff
@@ -116,10 +116,17 @@ function bool GetAdditionalTargets(out AvailableTarget AdditionalTargets)
...
 	local XComGameStateHistory History;
 	local XComGameState_BaseObject StateObject;
-	local array<StateObjectReference> ValidationTargets;
 
 	TargetedActor = GetTargetedActor();
 	TargetedPawn = XGUnit(TargetedActor);
+
+	// Start Issue #1408
+	/// HL-Docs ref:Bugfixes; issue:1408
+	/// Empty the additional target array, since it can already have incorrectly assigned targets by native code
```
Iridar, line 125: The Bugfixes docs must be like a changelog entry, they need to explain the nature of the bug and how it is getting fixed.  Something like: "Fix bug with Arc Wave targeting where it could hit enemies that were not targeted by the player" or something along those lines.

Iridar (changes requested):
Another unfortunate issue is that Highlander guidelines require that PRs that add new files to source must have the file addition in one commit, and everything else associated with the Issue in the second commit. This PR now has three commits, the two last commits must be squashed together before I can merge this PR.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TargetingMethod_ArcWave.uc
furudee: I've squashed the two commits to one and rephrased the doc entry

Iridar: Looks good to me. I think this is ready for merge, but just to confirm, did you run some tests with the new implementation to check it's fixing the issue?

furudee:
Diagnosing another issue related to arcwaves and came across this:

In UITacticalHUD_AbilityContainer::ConfirmAbility where the GetAdditionalTargets function is called, it checks to see if the function returns true before overwriting the array:

		if( TargetingMethod.GetAdditionalTargets(AdditionalTarget) )
		{
			//	overwrite target to include data from GetAdditionalTargets
			AvailableActionInfo.AvailableTargets[TargetIndex] = AdditionalTarget;
		}

Should the GetAdditionalTargets function be changed to always return true, since we would like to overwrite the array if it has no valid additional targets?

In the edge-case where TargetedPawn != none or Tiles.Length > 1, it would return false and thus not overwrite the array

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2TargetingMethod_ArcWave.uc
furudee: Yes, it appears to be working

pic

Iridar:
Alright, this is good to go then. I'll look at the other issue.

Issue #326, opened 2018-11-06
X2Action_ApplyWeaponDamageToUnit ignores X2Effect_ApplyWeaponDamage's damage type


robojumper:
Example of real code impacted:

`X2Ability_Cyberus`:

    RiftDamageEffect = new class'X2Effect_ApplyWeaponDamage';
    RiftDamageEffect.EffectDamageValue.DamageType = 'Psi';
    RiftDamageEffect.EffectDamageValue = class'X2Item_DefaultWeapons'.default.CYBERUS_PSI_BOMB_BASEDAMAGE;
    RiftDamageEffect.bIgnoreArmor = true;
    Template.AddMultiTargetEffect(RiftDamageEffect);

`X2Action_ApplyWeaponDamageToUnit`:

    //Set up a damage type
    if (WeaponTemplate != none)
    {
        DamageTypeName = WeaponTemplate.BaseDamage.DamageType;
        if (DamageTypeName == '')
        {
            DamageTypeName = WeaponTemplate.DamageTypeTemplateName;
        }
    }
    else if (TickContext != none || WorldEffectsContext != none)
    {
        //...
    }
    else
    {
        DamageTypeName = class'X2Item_DefaultDamageTypes'.default.DefaultDamageType;
    }

The Psi Rift doesn't use Psi damage flags.

Labelling this as a bug (a really minor one though!)

Notes: `DamageTypeName` in `X2Action_ApplyWeaponDamageToUnit` is only used for the world message type (red damage -- psi damage) and for the Unit pawn hit effects in some edge cases. We can afford to make the check somewhat overzealous without really breaking anything.

PR #1411, opened 2024-11-23
Allow Effects to over-ride damage types set by weapon templates for the purposes of damage flyovers

### LINKED ISSUE #1274: X2Effects which modify a weapon's base damage type do not visualise correctly

BlackDog86:
Fixes #1274


BlackDog86:
Pretty happy with this implementation. Commit is simple and clean, docs are done, doesn't interfere with any base-game logic and it's fairly easy to use (modders just have to set bOverRideWeaponTemplateForFlyover on the ability template and it'll use the X2Effect's DamageTypes array for the flyover instead of the one specified by the weapon template). I've tested it with sword slice (got conventional flyover) and a custom ability with the over-ride set and psionic damage in the damagetypes array on the effect (get psionic flyover).
Conv:

Over-ride ability:

Iridar:
Unfortunately this PR is a no-go, since X2Effect is native and you can't add new variables to it.

BlackDog86:
I just realised, we don't actualy 'need' a flag on the X2Effect / ability template at all. If we just check the damageresults array on the unit, iterate through the X2Effects to find what effect caused the damage, then check the damagetypes array for instances where the effect damage is Psi and set the damagetypename there, it will change the flyover if the X2Effect is psionic but does nothing otherwise. Then it will fall back on the default behaviour of getting it from the weapon template if it's not set on the effect. I've tested it and it seems to work correctly with base-game abilities and modded ones where Psi is added as a damagetype on an X2Effect applied on conventional weapons - it shows psi flyovers for the modded damage and normal damage when just shooting with the weapon.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
```diff
@@ -157,8 +159,31 @@ function Init()
 			WeaponTemplate = X2WeaponTemplate(SourceItemGameState.GetMyTemplate());
 	}
 
-	//Set up a damage type
-	if (WeaponTemplate != none)
+	// Start Issue #1274 - Check the DamageResults array on the Unit and find the relevant X2Effect that caused the damage.
+	// Check the damage types array of the effect and if 'psi' is found on any of the elements, set the DamageTypeName for the flyover to Psi.
+	if (UnitState != none && UnitState.DamageResults.Length != 0)
+	{
+		for (DmgIndex = 0; DmgIndex < UnitState.DamageResults.Length; ++DmgIndex)
```
Iridar, line 166: It would be more optimal to iterate from the end of the array, since that's where our context would be.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
```diff
@@ -157,8 +159,31 @@ function Init()
...
+		{
+			if (UnitState.DamageResults[DmgIndex].Context == StateChangeContext)
+			{
+				LookupEffect = UnitState.DamageResults[DmgIndex].SourceEffect.EffectRef;
+				SourceEffect = class'X2Effect'.static.GetX2Effect(LookupEffect);
+				
+				for(DamageTypeIndex = 0; DamageTypeIndex < SourceEffect.DamageTypes.Length; ++DamageTypeIndex)
```
Iridar, line 173: Since you're explicitly checking for `'Psi'`, you could use a `.Find()` instead, it's faster.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
```diff
@@ -157,8 +159,31 @@ function Init()
 			WeaponTemplate = X2WeaponTemplate(SourceItemGameState.GetMyTemplate());
 	}
 
-	//Set up a damage type
-	if (WeaponTemplate != none)
+	// Start Issue #1274 - Check the DamageResults array on the Unit and find the relevant X2Effect that caused the damage.
```
Iridar, line 162: This is technically a bugfix, is it not? I think bugfixes docs would be appropriate.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
```diff
@@ -157,8 +159,31 @@ function Init()
...
+							DamageTypeName = SourceEffect.DamageTypes[DamageTypeIndex];	
+							break;				
+						}			
+					}					
+			}
+		}
+	}
+	// If a psi DamageTypeName is not found on the X2Effect, fallback to the weapon template
+	// End Issue #1274
+	if (WeaponTemplate != none && DamageTypeName == '')
```
Iridar, line 186: If I understand the issue correctly, the game can display two types of damage flyovers, regular and psionic one. If the ability is not bound to a weapon, the game will automatically determine the damage type by looking at the damage types in the damaging effect, but if the ability is bound to a weapon, the game will always display the flyover for the damage type of the weapon, even if the ability does not apply base damage.

Is this correct? 

If so, I would suggest modifying the logic so it displays the damage type of the effect even if the ability is attached to the weapon, as long as the effect has `bIgnoreBaseDamage = true`. Then you it would be a one line fix, and it wouldn't be necessary to duplicate the code of finding damage types in the effect.

Something like:

instead of:
```cs
if (WeaponTemplate != none)
```
have this:
```cs
DamageEffect = X2Effect_ApplyWeaponDamage(OriginatingEffect);
if (WeaponTemplate != none && DamageEffect  != none && !DamageEffect.bIgnoreBaseDamage)
```
Alternatively, we could rearrange the original logic, so it always prioritizes the damage type on the effect, and if none specified, only then it defaults to the weapon's damage type. 

I'm sorry to be prickly, but I'm really not hot about a solution that would involve mostly copypasting vanilla code in order to bypass vanilla code, that just doesn't seem right.

Iridar (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
```diff
@@ -156,9 +157,24 @@ function Init()
 		if (SourceItemGameState != none)
 			WeaponTemplate = X2WeaponTemplate(SourceItemGameState.GetMyTemplate());
 	}
+	// Start Issue #1274
+	/// HL-Docs: feature:OverRideWeaponDamageForFlyovers; issue:1274; tags:tactical
```
Iridar, line 161: This seems mostly good to go, though I still have some nitpicks. Now that this doesn't actually have any modmaker-facing functionality, I'd suggest making it a ref:Bugfixes, cuz mostly it's just about making in-game visuals match expectations.

It also seems max line length for GitHub is around 114 symbols, so I suggest keeping docs line lengths under 110 symbols. 

Finally, while Yellow flyovers are technically are a type of flyover that can be displayed, I don't think this issue has any interaction with them, exactly, especially that - correct me if I'm wrong - crits will override psionic flyover anyway. 

Finally, I think it would be prudent to be explicit that we take the damage type only from the damage effect (X2Effect_ApplyWeaponDamage or subclass), then we take it from the weapon, and if there's no weapon, we take it from the effect even if it's not X2Effect_ApplyWeaponDamage. Which is tbh a bit of a weird way to do things, perhaps just taking damage type from any effect first, and then default to weapon would be better.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
```diff
@@ -156,9 +157,24 @@ function Init()
...
+	/// This feature adjusts the base-game logic for determining whether to display the standard Red (Normal Damage), Yellow (Crit Damage)
+	/// or Purple (Psionic) flyover when damage is inflicted on a unit. In the base-game, the damage type is first taken from the weapon template,
+	/// then if nothing is found, it checks for DOTs / environmental damage. Issue #326 resolved the issue where psionic effects not bound to
+	/// weapons were displaying the 'normal' damage flyovers - However, some modded abilities may want to display the psionic flyover even
+	/// when bound to a weapon with a 'normal' base damage type. This feature re-works the logic so the damagetype is taken from the originating effect
+	/// first, then falls back to the weapon's base damage type instead.
+
```
Iridar, line 168: Stray empty line

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
```diff
@@ -156,9 +157,24 @@ function Init()
...
+	/// HL-Docs: feature:OverRideWeaponDamageForFlyovers; issue:1274; tags:tactical
+	/// This feature adjusts the base-game logic for determining whether to display the standard Red (Normal Damage), Yellow (Crit Damage)
+	/// or Purple (Psionic) flyover when damage is inflicted on a unit. In the base-game, the damage type is first taken from the weapon template,
+	/// then if nothing is found, it checks for DOTs / environmental damage. Issue #326 resolved the issue where psionic effects not bound to
+	/// weapons were displaying the 'normal' damage flyovers - However, some modded abilities may want to display the psionic flyover even
+	/// when bound to a weapon with a 'normal' base damage type. This feature re-works the logic so the damagetype is taken from the originating effect
+	/// first, then falls back to the weapon's base damage type instead.
+	
+	EffectDamageType = X2Effect_ApplyWeaponDamage(OriginatingEffect);
+
```
Iridar, line 170: Stray empty line.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
```diff
@@ -187,19 +203,21 @@ function Init()
 	/// HL-Docs: ref:Bugfixes; issue:326
```
Iridar, line 203: Docs for Issue #326 seems to have gotten cut in half.

Iridar (changes requested):

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
BlackDog86: This isn't adjusted from the current codebase. I've adjusted the final line to be part of the original docs description though, if that's preferred?

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
```diff
@@ -186,21 +204,22 @@ function Init()
...
-	if (DamageTypeName == '')
-	{
-		if (X2Effect_ApplyWeaponDamage(OriginatingEffect) != none)
-		{
-			DamageTypeName = X2Effect_ApplyWeaponDamage(OriginatingEffect).EffectDamageValue.DamageType;
-		}
-
+	/// (e.g. Effect Damage from X2Effect_ApplyWeaponDamage, PsiBombStage2 & modded abilities)
+	
+	// Start Issue #1274 - This code block is not needed as the behaviour is now checked earlier in the function
```
Iridar, line 209: Technically, the behavior is now different. Previously we would use any damage type specified in the effect, now we use only if it is 'Psi'. However, since this logic is used to determine exclusively whether we use red or purple flyover, it boils down to the same thing in the end anyway.

Iridar (approved):

Iridar:
The purist in me hates the logic that doesn't support all damage types, but since the damage type in this code context is used purely to separate psi flyover from non-psi, it doesn't matter.

This is good to go.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_ApplyWeaponDamageToUnit.uc
BlackDog86: Yes that's correct - basically, my justification for the adjustment was that without it, things like Rend would end up having the wrong flyover and would need to be patched (In the case of Rend, the base damage type on the effect is 'melee' but the damagetype on the weapon template is 'psi' so without the check, it would take 'melee' preferentially - clearly that's an example of an ability where it was intended to display as psionic). There may be other abilities affected too so this is a 'softer / more tolerant' approach, which doesn't rely on having to adjust any effect or weapon damage types down the line.

PR #1416, opened 2024-11-28
Make teleport abilities not offensive

### LINKED ISSUE #1391: Codex Teleport ability is offensive
`X2AbilityTemplate` has `Hostility = eHostility_Offensive` by default, and Codex' Teleport doesn't set anything else to it, which can cause some unexpected ability interactions. 

I believe this to be an oversight, because as a general rule, abilities shouldn't be offensive unless they apply damage or negative effects.

Iridar:
Fixes #1391


Tedster59 (approved):
Looks good to me

Issue #1417, opened 2024-11-29
Actions that grab the UI on the geoscape sometimes do not properly check if something else is happening


Tedster59:
For example, Covert Actions only check if the strategy map is present, the UI is not in flight mode, and that a UIAlert screen isn't present. 

![image](https://github.com/user-attachments/assets/b2508528-58f8-407c-8c32-4d08efcfd54d)

However, this screen check doesn't allow for other types of screens to block the return, including Squad Select, Black Market, etc. If two things want to show an alert or another custom mod added screen on the same geoscape tick, you can get collisions such as the following clip:

https://www.twitch.tv/xwynns/clip/ObliqueGrotesqueHerdGivePLZ-GSVBaVbonwQzarHc

In this case, it's LWOTC stepping on this rake, trying to show its "must launch or abort" mission screen when the black market loads, but this can happen with the base game too: https://youtu.be/rNSZXlFeSmU?t=11639
In this case, the Covert Op check above doesn't know the game is currently in the middle of the UFO defense mission, so completes the covert op, allowing you to skip the mission entirely and be chased around by a UFO the rest of the game.

Fix: Update the checks to check that the current screen is the strategy map instead of checking that it isn't an alert, like what @Xymanek does in CI:

![image](https://github.com/user-attachments/assets/bcc211fa-05c2-40b6-8019-0c7c45e2038a)

Tedster59:
Found a few more in the base game that I need to update, so re-opening this issue to get them fixed. This search is against a SrcOrig folder with CHL overwriting the existing ones, so these should be most that were some form of the original issue.

![Image](https://github.com/user-attachments/assets/86876f64-dd20-4384-8ab2-9a5bc76bc84e)

XComHeadquartersInput uses a slightly different check for a few functions it has and I'm not sure if it needs updating as well:

![Image](https://github.com/user-attachments/assets/c43ca240-c749-43a3-a98a-265527a83407)

PR #1418, opened 2024-11-30
Swap "UI is busy" checks to a more robust function

Tedster59:
Fixes issue #1417 by updating all uses of 

```
(StrategyMap != none && StrategyMap.m_eUIState != eSMS_Flight && !`HQPRES.ScreenStack.IsCurrentClass(class'UIAlert')
```

to a helper function written by @Xymanek for Covert Infiltration that checks if the current screen is the strategy map instead of just checking that an alert is present.

```
static function bool GeoscapeReadyForUpdate()
{
	local UIStrategyMap StrategyMap;

	StrategyMap = `HQPRES.StrategyMap2D;

	return
		StrategyMap != none &&
		StrategyMap.m_eUIState != eSMS_Flight &&
		StrategyMap.Movie.Pres.ScreenStack.GetCurrentScreen() == StrategyMap;
}
```

image

This will solve edge cases where two geoscape entities try to throw UI popups at the same time, such as a Covert Action returning during the UFO chase and allowing an Avenger Defense mission to be skipped.

### DIFF
--- a/Components/DLC2CommunityHighlander/DLC2CommunityHighlander/DLC2CommunityHighlander.x2proj
+++ b/Components/DLC2CommunityHighlander/DLC2CommunityHighlander/DLC2CommunityHighlander.x2proj
@@ -42,6 +42,9 @@
     <Content Include="Src\DLC_2\Classes\X2Ability_DLC_Day60ViperKing.uc">
       <SubType>Content</SubType>
     </Content>
+    <Content Include="Src\DLC_2\Classes\X2DownloadableContentInfo_DLC_Day60.uc">
+      <SubType>Content</SubType>
+    </Content>
     <Content Include="Src\DLC_2\Classes\X2Item_DLC_Day60Armors.uc">
       <SubType>Content</SubType>
     </Content>
--- a/Components/DLC2CommunityHighlander/DLC2CommunityHighlander/Src/DLC_2/Classes/X2DownloadableContentInfo_DLC_Day60.uc
+++ b/Components/DLC2CommunityHighlander/DLC2CommunityHighlander/Src/DLC_2/Classes/X2DownloadableContentInfo_DLC_Day60.uc
@@ -1,0 +1,881 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2DownloadableContentInfo_DLC_Day60.uc
+//  AUTHOR:  Ryan McFall
+//           
+//	Installs Day 60 DLC into new campaigns and loaded ones. New armors, weapons, etc
+//
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+
+class X2DownloadableContentInfo_DLC_Day60 extends X2DownloadableContentInfo;
+
+var config array<name> DLCTechs;
+var config array<name> DLCPointsOfInterest;
+var config array<name> DLCObjectives;
+
+var config name DLCAlwaysStartObjective;
+var config name DLCOptionalNarrativeObjective;
+var config name DLCNoNarrativeObjective;
+
+var config name ArmoryUpgradeName;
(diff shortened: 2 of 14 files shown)


Components/DLC2CommunityHighlander/DLC2CommunityHighlander/DLC2CommunityHighlander.x2proj
```diff
@@ -42,6 +42,9 @@
     <Content Include="Src\DLC_2\Classes\X2Ability_DLC_Day60ViperKing.uc">
       <SubType>Content</SubType>
     </Content>
+    <Content Include="Src\DLC_2\Classes\X2DownloadableContentInfo_DLC_Day60.uc">
```
Iridar, line 45: This seems to be in the wrong commit.

Components/DLC2CommunityHighlander/DLC2CommunityHighlander/Src/DLC_2/Classes/X2DownloadableContentInfo_DLC_Day60.uc
```diff
@@ -345,20 +345,24 @@ static event UpdateDLC()
...
+		//StrategyMap = `HQPRES.StrategyMap2D;
 		RulerMgr = XComGameState_AlienRulerManager(History.GetSingleGameStateObjectForClass(class'XComGameState_AlienRulerManager'));
 		
 		NewGameState = class'XComGameStateContext_ChangeContainer'.static.CreateChangeState("Update Alien Hunters DLC");
 		RulerMgr = XComGameState_AlienRulerManager(NewGameState.ModifyStateObject(class'XComGameState_AlienRulerManager', RulerMgr.ObjectID));
 		
 		// Don't trigger updates while the Avenger or Skyranger are flying, or if another popup is already being presented
-		if (StrategyMap != none && StrategyMap.m_eUIState != eSMS_Flight && !`HQPRES.ScreenStack.IsCurrentClass(class'UIAlert'))
+		/// HL-Docs: ref:Bugfixes; issue:1417
+		/// Use a more robust check to determine if the game should trigger the update.
```
Iridar, line 364: The docs for bugfixes should be user-facing and akin to a changelog, I.e. "This is fixing bug X that can occur when Y". And just one HL-Docs entry is enough, others can be just regular Issue tags. The appropriate place for HL-Docs would be above the new helper function.

X2WOTCCommunityHighlander/X2WOTCCommunityHighlander.x2proj
```diff
@@ -811,6 +811,9 @@
     <Content Include="Src\XComGame\Classes\XComGameState_AIUnitData.uc">
       <SubType>Content</SubType>
     </Content>
```
Iridar, line 813: Also in the wrong commit.

Iridar (changes requested):

Components/DLC2CommunityHighlander/DLC2CommunityHighlander/DLC2CommunityHighlander.x2proj
Tedster59: Fixed

X2WOTCCommunityHighlander/X2WOTCCommunityHighlander.x2proj
Tedster59: Fixed

Tedster59:
Ok, so this causes a mod to break the game, because said mod swaps the geoscape for a subclass using a UISL.
https://steamcommunity.com/sharedfiles/filedetails/?id=1161736731

```
`HQPRES.StrategyMap2D = `HQPRES.Spawn(class'Override_UIStrategyMap', `HQPRES);
```

I expect this is causing the following check to fail:

```
StrategyMap.Movie.Pres.ScreenStack.GetCurrentScreen() == StrategyMap;
```

Issue #1419, opened 2024-12-01
Highlander causing reaper voicelines to not trigger


Iridar:
Source: https://www.reddit.com/r/xcom2mods/comments/1h3eitf/x2wotccommunityhighlander_causing_reaper/

> Wanted to return to modded xcom after a long break, had problems with reapers staying silent when doing blue move, certain abilities and on killing enemies.

> After some testing I found that highlander was the cause. Tried verifying files and resetting config but it only fixed the issue for blue move, reapers are still silent when killing aliens regardless of whether they had, had not or lost their shadow concealment after the killing shot.

furudee:
Seems to happen with all hero classes
Debugger tells me it wants to play a personality line of whatever action, like Moving_INTENSE yet nothing happens after it's passed into PlayAkEvent in `XComCharacterVoice::PlaySoundForEvent`

Running console command like `MakeCharacterSpeak Moving_INTENSE` on a hero class produces nothing both in highlander and vanilla game, yet according to vanilla script file in `XGUnit::MaybeAddPersonalityToSpeech` it should be picking these same personality lines in vanilla game too, but it plays the lines normally

I made hero classes with base-game personalities so that it would pick a proper personality line according to vanilla script

I can't get DevSound log header to work even with `UnSuppress DevSound` so in vanilla game I don't know if it actually is trying to pass a personality voiceline into PlayAkEvent or not

BlackDog86:
Yeah, I came to roughly the same conclusions. There are a few confounding factors though:

1. If the squad is concealed then it will use the MaybyUseHushedSpeechInstead function, which takes precedance over the personality lines so you need to make sure the squad is unconcealed before testing
2. If there is narrative playing, that also suppresses the voicelines. 

That said, your assessment seems more or less on point to me. Console commanding Moving_<PERSONALITY> doesn't work (for either normal or hero soldiers, base or CHL) so we can't test it that way.

However, I was able to reproduce the issue in the base game as well - When using a hero unit greater than veteran rank and unconcealed, it also tries to play one of the Moving_<PERSONALITY> lines and fails so I don't think this is purely a CHL issue, it's just relatively conditional.

[...]

PR #1422, opened 2024-12-08
Add a check if a WWise voiceline exists to prevent attempts to play missing voicelines

furudee:
Fixes issue #1419 

The implementation can be expanded to check if SoundCues exist, if so is wished, yet it's redundant since in `XComCharacterVoice::PlaySoundForEvent` it makes such check before playing a soundcue voiceline

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnit.uc
@@ -1452,7 +1452,7 @@ function UnitSpeak(Name nCharSpeech, bool bDeadUnitSound = false)
 	nHushedSpeech = MaybeUseHushedSpeechInstead(nCharSpeech);
 	nPersonalitySpeech = MaybeAddPersonalityToSpeech(nCharSpeech);
 
-	// Start Issue #1419 - Add an additional condition
+	// Start Issue #1419 - Perform an AkSpeechExists() check before calling m_kPawn.UnitSpeak()
 	if (nHushedSpeech != '' && AkSpeechExists(nHushedSpeech))
 	{
 		m_kPawn.UnitSpeak(nHushedSpeech);
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnit.uc
@@ -1452,15 +1452,53 @@ function UnitSpeak(Name nCharSpeech, bool bDeadUnitSound = false)
 	nHushedSpeech = MaybeUseHushedSpeechInstead(nCharSpeech);
 	nPersonalitySpeech = MaybeAddPersonalityToSpeech(nCharSpeech);
 
-	if (nHushedSpeech != '')
+	// Start Issue #1419 - Add an additional condition
+	if (nHushedSpeech != '' && AkSpeechExists(nHushedSpeech))
+	{
 		m_kPawn.UnitSpeak(nHushedSpeech);
-	else if (nPersonalitySpeech != '')
+	}
+	else if (nPersonalitySpeech != '' && AkSpeechExists(nPersonalitySpeech))
+	{
 		m_kPawn.UnitSpeak(nPersonalitySpeech);
-	else
+	}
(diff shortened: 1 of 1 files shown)


BlackDog86:
I tested this earlier today and it does indeed have the intended effect (fixing missing voicelines of hero units with personalities / attitudes by falling back to the default voiceline if the attitude voiceline doesn't exist). So I'm OK with it, can you confirm in testing from your side as well?

furudee:
Yes, it seemed like it works, exactly as you said. I also tested with a random voicepack soldier and soundcues seem to pass through the function normally

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnit.uc
```diff
@@ -1452,16 +1452,52 @@ function UnitSpeak(Name nCharSpeech, bool bDeadUnitSound = false)
...
 		m_kPawn.UnitSpeak(nCharSpeech);
+	}
 
 	m_fTimeSinceLastUnitSpeak = 0.0f;
 }
 
+// Start Issue #1419
+/// HL-Docs: ref:Bugfixes; issue:1419
+/// Characters that use WWise to play voicelines no longer attempt to play missing voicelines
+function bool CheckAkSpeechExists(Name Speech)
```
Iridar, line 1474: This is internal CHL function and should be marked as `private`.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnit.uc
```diff
@@ -1452,16 +1452,52 @@ function UnitSpeak(Name nCharSpeech, bool bDeadUnitSound = false)
...
+{
+	local string AkEvent;
+	local int SoundIndex;
+
+	// check if unit voice has a WWise SoundBank
+	if(XComHumanPawn(m_kPawn).Voice != None && XComHumanPawn(m_kPawn).Voice.AkBankName != "")
```
Iridar, line 1480: CHL code convention is to include a space after if, i.e. do `if (Condition)`, not `if(Condition)`.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnit.uc
```diff
@@ -1452,16 +1452,52 @@ function UnitSpeak(Name nCharSpeech, bool bDeadUnitSound = false)
 	nHushedSpeech = MaybeUseHushedSpeechInstead(nCharSpeech);
 	nPersonalitySpeech = MaybeAddPersonalityToSpeech(nCharSpeech);
 
-	if (nHushedSpeech != '')
+	if(nHushedSpeech != '' && CheckAkSpeechExists(nHushedSpeech))
```
Iridar, line 1455: Changes must be marked with Start Issue / End Issue tags.

Iridar (changes requested):
This is mostly good to go, just needs some docs / code styling changes. Good work.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnit.uc
Iridar: Initially I meant something like this:

```cpp
if (nHushedSpeech != '' && AkSpeechExists(nHushedSpeech)) // Issue #1419 - add an AkSpeechExists() condition
{
	m_kPawn.UnitSpeak(nHushedSpeech);
}
else if (nPersonalitySpeech != '' && AkSpeechExists(nPersonalitySpeech)) // Issue #1419 - add an AkSpeechExists() condition
{
	m_kPawn.UnitSpeak(nPersonalitySpeech);
}
else if (AkSpeechExists(nCharSpeech)) // Issue #1419 - add an AkSpeechExists() condition
{
	m_kPawn.UnitSpeak(nCharSpeech);
}
```
However, now that I've looked at the original code, it's a poorly formatted mess, so they way you've ended up doing it is fine, with just one pair of Start - End Issue tags. 

However, I would be more explicit with the the comment then. Something like: 

`// Start Issue #1419 - Perform an AkSpeechExists() check before calling m_kPawn.UnitSpeak()`

I'll do it myself via web interface so you don't have to.

Iridar:
This is good to go. Thank you.

PR #1423, opened 2024-12-08
Fix for loading screen hang when bSkipCampaignIntroMovies = true in CHHelpers

### LINKED ISSUE #1393: bSkipCampaignIntroMovies=true causes infinite loading screen on gatecrasher
I'm using beta 1.29.0 with alien hunters DLC

BlackDog86:
Fixes #1393


Iridar:
Good docs, but I kinda hate the placement, we don't normally do comments between if-else brackets, let me adjust that.

Iridar:
This should be good to go, though I'm curious why does this issue not happen when using debug start? Does it take a different codepath?

Issue #1217, opened 2023-07-01
Add more Unit Value cleanups


Iridar:
Such as `eCleanup_EndTurn` and maybe `eCleanup_EndTactical`. Other suggestions are welcome. 

It's worth noting it should be possible for any mod to pass any arbitrary integer value along with SetUnitFloatValue() and CleanupUnitValues() and have it just work, but it would still be useful to handle the common additional cases in Highlander.

Iridar:
Leaving this around until there's an actual use case for it.

Issue #559, opened 2019-06-09
Direct transfers shouldn't clean up some UnitValues


rcd47:
The introduction of direct mission transfers created a hole in the cleanup logic for UnitValues. Values that are created with eCleanup_BeginTactical get cleaned up before every part of a mission, by XCGS_Unit OnBeginTacticalPlay. But this creates some bugs:

- Free reloads magically get replenished at the beginning of each part of the mission.
- If a unit earns a promotion in an early part of the mission, the promotion notification will be shown the first time they get a kill in every subsequent part of the mission.
- Sustain could trigger once in each part of the mission instead of once across all parts of the mission.
- Possibly others. There's a LOT of stuff that uses eCleanup_BeginTactical just in the base game code, never mind what mods might be doing with it.

I was going to send a PR to fix this, but I'm not sure what the right solution is. The easiest and most obvious choice is to alter XCGS_Unit OnBeginTacticalPlay so that it does not clean up the unit values if doing a direct transfer and the unit has appeared in some earlier part. But that alters the behavior of eCleanup_BeginTactical, which might be problematic, and doesn't offer any solution for situations where (for whatever reason) you actually want a UnitValue to be cleaned up before each part of a mission.

So then I considered adding a new option to EUnitValueCleanup. It could be called something like eCleanup_BeginTacticalChain. That way the behavior of eCleanup_BeginTactical stays the same, and mods can get either behavior as needed. But then I noticed that all of the UnitValue functions are native functions. So adding a new enum value is potentially dangerous given that there's no way to know what those functions are doing or how the UnitValue is actually stored.

Then I considered replacing the native UnitValue functions with non-native equivalents. But I don't know if it's safe to change a function definition like that, even if the rest of the signature remains identical. Also it's possible native code is reading and/or writing UnitValues, so replacing the functions effectively creates two separate UnitValue systems and that might not be safe.

So the safest option I can think of (but also the one that involves the most work, naturally) would be to create an "enhanced" UnitValues system that exists alongside the current one. This enhanced system would not be interoperable with the current system and would have its own functions/enums/structs/etc. The current system gets left alone, and the code that creates the bugs listed up top is changed to use the enhanced system. This minimizes risk, and offers the potential to make UnitValues more flexible. For example, enhanced UnitValues could support primitive types other than float.

Thoughts?

robojumper:
CC #44, which adresses a slightly different problem related to direct transfers.

[...]

PR #1424, opened 2024-12-09
Issue #559 - Add eCleanup_BeginTacticalChain

rcd47:
As per #559 (and also #1217), add eCleanup_BeginTacticalChain and apply it to the handful of base-game cases where it matters.

I thought I had submitted this patch years ago, but evidently I did not. I just played a whole campaign with this change in place and did not notice any problems.


Iridar (changes requested):
This is like super nitpicky, but traditionally issue tags are capitalized, i.e. do `Issue #111`, not `issue #111`. Other than that, this is good to go.

rcd47:
Fixed. I've nitpicked things like that in reviews too. 😄

Issue #1420, opened 2024-12-04
Right-click melee attack with Arc Wave is broken - incorrect amount of additional targets


furudee:
Related to #1408

Native function that assigns additional targets that an ability might have can make abilities with X2TargetingMethod_ArcWave targeting method produce incorrect additional targets that the ability cannot hit.

Issue is that in XComTacticalInput::ClickToPath function, which is called when the player right-clicks an enemy to make a melee attack, it doesn't validate the additional targets an ability might have. Thus in the case of X2TargetingMethod_ArcWave, the ability can hit seemingly random amount of additional targets.

A solution could be adding a segment, that calls GetAdditionalTargets and in the case of #1408, would validate the targets

I've made a mockup that seems to work:

```
function bool ClickToPath()
{
	local XComGameStateHistory History;
	local XComGameState_Unit ActiveUnitState;
	local XComGameState_Ability AbilityState;
	local GameRulesCache_Unit UnitCache;
	local XComPathingPawn PathingPawn;
	local array<TTile> WaypointTiles;
	local int ActionIndex;
	local int TargetIndex;
	local string ConfirmSound;

	//////////////////////////////////////////// NEW ////////////////////////////////////////////////
	local X2TargetingMethod_ArcWave TargetingMethod;
	local AvailableTarget AdditionalTargets;
	//////////////////////////////////////////// NEW END ////////////////////////////////////////////////

	if(`XCOMVISUALIZATIONMGR.VisualizerBlockingAbilityActivation())
	{
		return false;
	}

	PathingPawn = XComTacticalController(Outer).m_kPathingPawn;

	// try to do a melee attack
	History = `XCOMHISTORY;

	ActiveUnitState = XComGameState_Unit(History.GetGameStateForObjectID(GetActiveUnit().ObjectID));
	AbilityState = class'X2AbilityTrigger_EndOfMove'.static.GetAvailableEndOfMoveAbilityForUnit_CH(ActiveUnitState, PathingPawn.LastTargetObject); // Issue #1138
	if(AbilityState != none 
		&& PathingPawn.LastTargetObject != none
		&& `TACTICALRULES.GetGameRulesCache_Unit(ActiveUnitState.GetReference(), UnitCache))
	{

		// find the melee ability's location in the action array
		ActionIndex = UnitCache.AvailableActions.Find('AbilityObjectRef', AbilityState.GetReference());
[...]

PR #1427, opened 2024-12-11
Add target validation when making a right-click pathing attack with ability that has X2TargetingMethod_ArcWave

furudee:
Fixes issue #1420 - Depends on #1409 

Implementation has a few side-effects that I hopefully catched and resolved without anything else breaking:
`ArcWaveTargetingMethod.Init` will create a camera centered on target enemy, pull up the AbilityHUD and marks units and tiles ability will hit.
```
`CAMERASTACK.RemoveCamera(ArcWaveTargetingMethod.LookAtCamera); // won't center at target enemy
`PRES.GetTacticalHUD().CancelTargetingAction(); // pulls down the abilityhud
ArcWaveTargetingMethod.Canceled(); // unmarks units and tiles
```

I believe targeting method can be initialized with `InitFromState(AbilityState)` without side-effects, but it will be have variables undefined, which are either protected or private. For example, `var protected int LastTarget` in `X2TargetingMethod_MeleePath` and `var privatewrite AvailableAction Action` in `X2TargetingMethod` which are needed for the class functions to work properly

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComTacticalInput.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComTacticalInput.uc
@@ -3101,6 +3101,9 @@ function bool ClickToPath()
 	local int ActionIndex;
 	local int TargetIndex;
 	local string ConfirmSound;
+	// Variables for Issue #1420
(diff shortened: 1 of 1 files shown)


Iridar:
Looks good to me, as much as I can follow the issue / implementation.

PR #1433, opened 2024-12-27
Fix for weapon stacking behaviour

### LINKED ISSUE #1352: PCS without stats do not stack
Playing LWoTC, there are PCS without stats that add other abilities to soldiers, these PCS do not stack
XCom2_K6dEZAyW7E
XCOM2_Launcher_LXoqx9LEZ5
See attached image of the PCS selection screen and modlist
[Relevant save file here](https://drive.google.com/file/d/1Np92JYy1g6daPlNDA0ylsIvDfBk-268r/view?usp=sharing)

### LINKED ISSUE #1429: Weapons which have forced attachments lose their attachments when stacked
Some weapons (like the Tactical Legacy Pack weapons) do not set a Template.NumUpgradeSlots on the weapon template, and instead, have their attachments added via a custom OnAcquired function, meaning that by design, the attachments cannot be changed or removed. If mods end up creating more than one of these weapons, they become stacked in the inventory.

When this happens, there is a bug wherein the attachments may be removed when switching loadouts. 

Here are the exact steps to reproduce the issue:

1. Console command: giveitem TLE_AssaultRifle_MG 2 
2. Go to soldier loadout and equip one of the stacked TLE rifles

Attchments will be removed from the weapon when equipped (GetItemFromInventory creates a new item state when the item is equipped but this item does not get its upgrades from the onacquired function when the new item state is created). 

From there, it is then possible to permanently remove the attachments by swapping between the stacked weapons and normal ones - If HasUnmodifiedItem in XCGS_HeadquartersXCom reports that the item is unmodified (i.e. in the case of the weapon with the stripped attachments), it will then just increase the quantity of the stripped weapon instead of putting the old one back in the inventory, with its attachments.

Speculation on a fix would be to run the items onacquired function (if it exists) when equipping but I'm unsure whether this would have any adverse effects on other weapons which may have onacquired functions which do weird & wonderful things that you wouldn't want to run every time you equipped the item.

BlackDog86:
Fixes #1429 

This PR removes a check on number of upgrade slots from HasBeenModified. This was erroneously returning false for weapons where attachments are added by an onAcquired function andwith 0 upgrade slots on the template (e.g. TLE / chosen weapons), which would cause duplicates of those items to be stacked. This could cause downstream issues with attachments when equipping / removing them as the inventory functions expect all weapons with attachments to not be stacked.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
@@ -792,10 +792,14 @@ simulated function bool HasBeenModified()
 		return true;
 	//end issue #104
 		
-	WeaponTemplate = X2WeaponTemplate( m_ItemTemplate );
-
-	// Single line for Issues #93 and #306
-	if ((WeaponTemplate != none) && (GetNumUpgradeSlots() > 0) && (GetMyWeaponUpgradeCount() > 0))
+	WeaponTemplate = X2WeaponTemplate( m_ItemTemplate );	
+	/// HL-Docs: ref:Bugfixes; issue:1429
+	/// Fix bug that caused weapons that have weapon upgrades installed, despite not having any weapon upgrade slots, 
+	/// to stack in the HQ inventory, causing them to lose their installed upgrades when equipped. The fix removes the 
+	/// check for the number of weapon upgrade slots on the weapon from the logic that determines whether the item has 
+	/// been modified or not, and leaves just the check for number of installed weapon upgrades.
+	// Single line for Issues #93, #306 and #1429
+	if ((WeaponTemplate != none) /* && (GetNumUpgradeSlots() > 0)*/ && (GetMyWeaponUpgradeCount() > 0))
 		return true;
 
 	return false;


Larryturbo:
"items with identical attachments now stack together in the inventory"
As long as it can still see the weapon's colour and pattern as a difference. When multiple soldiers unequip their weapon for wherever reason, using the colour is how I can tell who dropped what.

BlackDog86:
Will have to test that - I imagine that customising the weapon sets the 'force unique' flag but I'll be away from the PC for a few days so not sure if I'll be able to test / address this at least until the new year.

faanlez:
`HasBeenModified()` does not check `WeaponAppearance` at all, it only cares about nickname and weapon upgrades. So with these changes identically upgraded weapons will stack and they will all get the same appearance. But, keep in mind that soldiers actually have their own weapon customization saved to them that they will use over the weapon's own customization *unless* `HasBeenModified()` returns `true`.

Larryturbo:
Then upgraded and customized weapons will lose their appearance and stack. That should never be the case.

BlackDog86:
Fair enough - It should be easy enough to revise to take account of customisation to be fair. If you force the behaviour to work 'correctly' & no-stack everything it makes a royal mess of the strategy debug start inventory and I do think having non unique items with the same attachments stacking makes sense (the original code in that function is downright awful) so I'd prefer to revise it that way than just blanket no-stack everything which isn't infinite but yeah - fixing the other two issues without playing with the item stacking behaviour is also perfectly possible.

BlackDog86:
Updated PR now checks weapon appearances when putting items back into the inventory and ensures that upgraded weapons with custom appearances do not stack (non-upgraded weapons / infinite items continue to set the default weapon appearance on the soldier which is used if no custom appearance is already set on the weapon).

This means:
Two yellow magnetic rifles with a scope, = stack
1 x Blue and 1 x Yeloow magnetic rifle with a scope = no stack
Two yellow magnetic rifles, one with a scope and one with a stock = no stack

Does that seem reasonable? I've also added the chosen weapons to the 'OnEquipFix' array so their attachments don't get stripped (so if, for whatever reason) mods decided to add multiple copies of the chosen weapons they wouldn't get stacked and have their attachments removed just like the TLE weapons.

BlackDog86:
Well the alternative is basically patching all of the weapons in X2Item_TLE_Weapons so that they fire the onAcquired function when you equip them, which is also a perfectly reasonable solution but I'm not sure if we're even allowed to modify that file since we don't have a 'TLE Highlander' as far as I know.

If I can't create a manual array of templates and I can't modify the templates, then I fail to see how the issue can be fixed.

Iridar:
Why is simply not stacking weapons with *any* weapon upgrades attached is not an option?

BlackDog86:
Of course, not stacking weapons with any upgrades is an option. Simply removing the check for number of slots in HasBeenModified() is sufficient to do that & implementing it that way also has the benefit of not needing to run the onacquired function when equipping the weapon (and thus no need for the config array), since the bit that strips the attachments in GetItemFromInventory can't happen if the weapons aren't stacked in the first place.

That said, when you just do the simple fix, what happens is HasBeenModified then *correctly* returns true for all items with attachments and it unstacks all the items created by strategy debug start (which is basically 10 x copies of every non-infinite item) so you get this:


I couldn't find the code that's responsible for doing this but if we can find it, perhaps we can use the event in HasBeenModified to over-ride the stacking behavior for those items.

Honestly, unstacking items in strat debug isn't a big deal for me & I'm perfectly happy to sacrifice a debug annoyance for better in-game control of the inventory so I can adjust the PR to just deal with the two explicit issues / bugs and not do anything regarding improving the item stacking behaviour, which I guess is probably the preferred approach.

Iridar:
Yeah, that's better. I'll do a proper review later. To be clear, my main issue wasn't with configurable array itself, but more with filling it in Highlander. If you want, you can still add a configurable array of stackable items with upgrades, but it must be empty by default.

Iridar:
And I'd probably prefer it done in a separate PR.

BlackDog86:
Makes sense - it would be a lot easier, now I understand how these functions work, to make a configurable feature in a new PR, allowing items with identical attachments & appearance to stack. I agree it's better that way.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersXCom.uc
```diff
@@ -4137,6 +4141,7 @@ function bool HasUnModifiedItem(XComGameState AddToGameState, X2ItemTemplate Ite
 		for(idx = 0; idx < Inventory.Length; idx++)
 		{
 			ItemState = XComGameState_Item(`XCOMHISTORY.GetGameStateForObjectID(Inventory[idx].ObjectID));
+			hasIdenticalStatBoosts = false;
```
Iridar, line 4144: What is the purpose of this line?

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersXCom.uc
```diff
@@ -4101,6 +4101,10 @@ function bool HasItemInInventoryOrLoadout(X2ItemTemplate ItemTemplate, optional
 function bool HasUnModifiedItem(XComGameState AddToGameState, X2ItemTemplate ItemTemplate, out XComGameState_Item ItemState, optional bool bLoot = false, optional XComGameState_Item CombatSimTest)
 {
 	local int idx;
+	// Start Issue #1352 - New variables for combat sim stacking behavior
+	local int idxStatBoost;
+	local bool hasIdenticalStatBoosts;
```
Iridar, line 4106: Variable naming conventions would have these as `StatBoostIdx` and `bHasIdenticalStatBoosts`.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersXCom.uc
```diff
@@ -4149,10 +4154,27 @@ function bool HasUnModifiedItem(XComGameState AddToGameState, X2ItemTemplate Ite
 				{
 					if(ItemState.GetMyTemplate().ItemCat == 'combatsim')
 					{
-						if(ItemState.StatBoosts.Length > 0 && CombatSimTest.StatBoosts.Length > 0 && ItemState.StatBoosts[0].Boost == CombatSimTest.StatBoosts[0].Boost && ItemState.StatBoosts[0].StatType == CombatSimTest.StatBoosts[0].StatType)
+						/// HL-Docs: ref:Bugfixes; issue:1352
+						/// Mod-added PCS items which did not provide exactly 1 stat boost were not being properly checked by HasUnModifiedItem
+						/// New checks added here mean that PCS items with no stat boosts, or more than one stat boost, are properly compared
+						// Begin Issue #1352
```
Iridar, line 4160: Begin Issue tag should be above the HL-Docs entry. I would also reword the docs:

> Fix bug that caused PCS items that did not provide exactly one stat boost to not stack properly in the UI by making `HasUnModifiedItem()` properly compare items with multiple stat boosts or no stat boosts at all.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_HeadquartersXCom.uc
```diff
@@ -4149,10 +4154,27 @@ function bool HasUnModifiedItem(XComGameState AddToGameState, X2ItemTemplate Ite
...
+						hasIdenticalStatBoosts = true;
+						if(CombatSimTest.StatBoosts.Length != 0)
```
Iridar, line 4162: Missing empty space after `if` and `for` in this entire code block.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
```diff
@@ -792,10 +792,14 @@ simulated function bool HasBeenModified()
 		return true;
 	//end issue #104
 		
-	WeaponTemplate = X2WeaponTemplate( m_ItemTemplate );
```
Iridar, line 795: Generally it's preferable to not fix multiple issues in one PR, so that later they can be reverted separately if necessary. So I'm gonna have to ask you to separate this into two PRs.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Item.uc
```diff
@@ -792,10 +792,14 @@ simulated function bool HasBeenModified()
...
-
-	// Single line for Issues #93 and #306
-	if ((WeaponTemplate != none) && (GetNumUpgradeSlots() > 0) && (GetMyWeaponUpgradeCount() > 0))
+	WeaponTemplate = X2WeaponTemplate( m_ItemTemplate );	
+	/// HL-Docs: ref:Bugfixes; issue:1429
+	/// When checking weapons for HasBeenModified, checking that the number of upgrade slots > 0 gives an incorrect response
+	/// for weapons which have attachments added by OnAcquired and therefore no slots on the weapon template (e.g. TLE / chosen 
+	/// weapons). Mods which add additional copies of these weapons then cause the items to stack, which interferes with downstream
+	/// logic in GetItemFromInventory and PutItemInInventory, causing attachments to be stripped when the stacked item is equipped.
```
Iridar, line 800: I'd reword docs:

> Fix bug that caused weapons that have weapon upgrades installed, despite not having any weapon upgrade slots, to stack in the HQ inventory, causing them to lose their installed upgrades. The fix removes the check for the number of weapon upgrade slots on the weapon from the logic that determines whether the item has been modified or not, and leaving just the check for installed weapon upgrades.

Though feel free to leave the mention of the downstream logic if you want.

Iridar (changes requested):
Good work, but needs hair-licking.

PR #1437, opened 2024-12-29
Add bunch of knockback fixes

### LINKED ISSUE #1431: X2Effect_Knockback tile trace intersects with ground
In ```X2Effect_Knockback : GetTilesEnteredArray```, an if-condition fails to take ground tiles in account and then later prematurely exits the loop and causes KnockbackToLocation to not be where intended

From my limited testing case, a trace info from WorldData.GetAllActorsTrace returns two actors with following object archetypes:
```
BLDGKit_Floor_03x06_A
ARC_TWN_GroundPlaneAx16
```
```if( TraceHitInfo.HitActor == FloorTileActor )```
doesnt take into account if the actor is a ground plane.
```FloorTileActor = WorldData.GetFloorTileActor(TempTile)``` returns same XComLevelActor for both object archetypes,
but only the first one is actually exactly the floor tile actor.

So it continues to the if-condition where it was not supposed to go, 
and will fail the ```XComFracLevelActor(TraceHitInfo.HitActor) == none``` condition
Thus prematurely ending the for loop and setting the KnockbackToLocation at the ground plane.

For a fix, I suggest adding another condition to check if the tile is a ground tile, as so:
```
	if( WorldData.GetAllActorsTrace(StartLocation, KnockbackToLocation, Hits, Extents) )
	{
		foreach Hits(TraceHitInfo)
		{
			TempTile = WorldData.GetTileCoordinatesFromPosition(TraceHitInfo.HitLocation);
			FloorTileActor = WorldData.GetFloorTileActor(TempTile);

			if( TraceHitInfo.HitActor == FloorTileActor -----NEW-----> || WorldData.IsGroundTile(TempTile) <-----NEW------ )
			{
				continue;
			}

			if ((!CanBeDestroyed(XComInteractiveLevelActor(TraceHitInfo.HitActor), DamageAmount) && XComFracLevelActor(TraceHitInfo.HitActor) == none) || !bKnockbackDestroysNonFragile)
			{
				//We hit an indestructible object
				KnockbackToLocation = TraceHitInfo.HitLocation + (-OutAttackDirection * 16.0f); //Scoot the hit back a bit and use that as the knockback location
				break;
			}
		}
	}
```

Following picture before and after adding the new condition
test

### PR DESCRIPTION
**Pull request by furudee**
Fixes #1431
This request does a few things:

1. Makes `WorldData.GetAllActorsTrace` a line trace by commenting out Extents, since it often got stuck on ground, curbs and other floor tiles. I assume the logic was that units cannot be knocked down from buildings and other high places
2. Units may now take fall damage from falling from such high places, and units landed on take damage. ~~After whatever gamestate the knockback is in, I create a delegate to make a new falling gamestate to deal with the falling action.~~
I have a new function `HandleFalling` that is a modified version of `XComGameStateContext_Falling::ContextBuildGameState` function which deals with dealing damage to falling units ect. Visualization actions for that are added in `AddX2ActionsForVisualization`
3. Units may fly over other units and land on them, if they're high enough to take fall damage. Before it ended and set the knockback location before the unit you're flying over.
4. ~~Adds custom visualization tree merging logic when fall is triggered by a knockback. This removes the `X2Action_Falling` node from the build tree, since otherwise the unit will "fall down" twice. A normal falling context caused by destruction of floor beneath the unit shouldn't end up at the newly added segment, but I'm not certain~~ 
Not relevant after amending the PR

And since it's a quite large fix, I would like it to be tested more

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Knockback.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Knockback.uc
@@ -1,0 +1,429 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2Effect_Knockback.uc
+//  AUTHOR:  Ryan McFall  --  5/5/2015
+//  PURPOSE: This effect implements a game mechanic that moves units based on incoming 
+//			 damage or attacks. This is done with a rag doll similar to what was done with
+//			 kinetic strike in EW.
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+class X2Effect_Knockback extends X2Effect config(GameCore);
+
+struct KnockbackDistanceOverride
+{
+	var name OverrideReason;
+	var int NewKnockbackDistance_Meters;
+};
+
+/** Distance that the target will be thrown backwards, in meters */
+var() int KnockbackDistance;
+
+/** Used to step the knockback forward along the movement vector until either knock back distance is reached, or there are no more valid tiles*/
+var() private float IncrementalStepSize;
+
+/** If true, the knocked back unit will cause non fragile destruction ( like kinetic strike ) */
+var() bool bKnockbackDestroysNonFragile;
+
+/** Distance that the target will be thrown backwards, in meters */
+var() float OverrideRagdollFinishTimerSec;
+
+/** Knockback effects can happen on every attack or only killing attacks */
+var() bool OnlyOnDeath;
+
+var config	float DefaultDamage;
+var config	float DefaultRadius;
+var config  array<KnockbackDistanceOverride> KnockbackDistanceOverrides;
+
+function name WasTargetPreviouslyDead(const out EffectAppliedData ApplyEffectParameters, XComGameState_BaseObject kNewTargetState, XComGameState NewGameState)
+{
+	// A unit that was dead before this game state should not get a knockback, they are already a corpse
+	local name AvailableCode;
+	local XComGameState_Unit TestUnitState;
+	local XComGameStateHistory History;
+
+	AvailableCode = 'AA_Success';
+
+	History = `XCOMHISTORY;
+
+	TestUnitState = XComGameState_Unit(History.GetGameStateForObjectID(kNewTargetState.ObjectID));
+	if( (TestUnitState != none) && TestUnitState.IsDead() )
+	{
+		return 'AA_UnitIsDead';
+	}
+
+	if( OnlyOnDeath )
+	{
+		TestUnitState = XComGameState_Unit(kNewTargetState);
+		if( TestUnitState != None && (TestUnitState.IsAlive() || TestUnitState.IsIncapacitated()) )
+		{
+			return 'AA_UnitIsAlive';
+		}
+	}
+
+	return AvailableCode;
+}
+
+private function bool CanBeDestroyed(XComInteractiveLevelActor InteractiveActor, float DamageAmount)
+{
+	//make sure the knockback damage can destroy this actor.
+	//check the number of interaction points to prevent larger objects from being destroyed.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Iridar** [comment]
Looks very intriguing, I can't really follow the code, so we'll just have to do some extensive testing.

**Comment by BlackDog86** [comment]
Fixes #1431

**Comment by Tedster59** [comment]
Looks good to me, want to see proper testing

**Comment by furudee** [comment]
A while ago I released this fix as [a standalone mod ](https://steamcommunity.com/sharedfiles/filedetails/?id=3474547616) and I haven't had people complain that it doesn't work: 
one user said that it crashed their game after finishing a mission when they installed it mid-campaign, but I think it's because it MCO'ed `XComGameStateContext_Falling`, and I wasn't aware that any MCO that overrides state classes or context classes will crash your game if installed/removed mid-campaign (I think)

File: X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Knockback.uc [hunk 1]
```diff
@@ -0,0 +1,520 @@
...
+
+/** Knockback effects can happen on every attack or only killing attacks */
+var() bool OnlyOnDeath;
+
+var config	float DefaultDamage;
+var config	float DefaultRadius;
+var config  array<KnockbackDistanceOverride> KnockbackDistanceOverrides;
+
+// Variables for Issue #1431
+var XComGameState_Unit FallingUnit;
```
Tedster59, line 39: This is very bad.  Causes garbage collection errors on mission end.  need to rewrite this and the lines below it, anything that is runtime should be kept inside the functions.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Knockback.uc
```diff
@@ -227,7 +235,16 @@ private function GetTilesEnteredArray(XComGameStateContext_Ability AbilityContex
 				{
 					TileUnits = WorldData.GetUnitsOnTile(TempTile);
 					if (TileUnits.Length > 0)
-						break;
+					{
+						/// HL-Docs: ref:Bugfixes; issue:1431
+						/// Units knocked back may fly over units and land on them if they are high enough to take fall damage
+						// Start Issue #1431
+						if ((StartTile.Z - TempTile.Z) < 4)
```
Tedster59, line 242: Should this be opened to config? or is 4 the "standard" height.

Tedster59 (changes requested):
For now, marking Request Changes, because this needs rewriting to not store runtime variables at the template level.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Knockback.uc
furudee: It's `iStoryHeightInTiles` local variable from `XComGameStateContext_Falling::ContextBuildGameState`, so a hardcoded value

furudee:
I have amended the PR to do all the operations inside functions so avoiding putting any new global variables that caused a garbage collection crash, now the PR only has one added new file `X2Effect_Knockback` and the previous file `XComGameStateContext_Falling` has been removed

Tedster59 (approved):
Changes look good.  Have you tested the updated version?

BlackDog86:
Noting that this issue is closed, it creates a warning when compiling CHL - 

Instead of doing this:
```
local X2Action_ApplyWeaponDamageToUnit DamageAction;
DamageAction = X2Action_ApplyWeaponDamageToUnit(class'X2Action_ApplyWeaponDamageToUnit'.static.AddToVisualizationTree(ActionTrack, VisualizeGameState.GetContext(), false, KnockbackAction));
```

You can just call the function without assigning it to a variable and then not manipulating the variable :)
`class'X2Action_ApplyWeaponDamageToUnit'.static.AddToVisualizationTree(ActionTrack, VisualizeGameState.GetContext(), false, KnockbackAction));` should be sufficient.

PR #1438, opened 2025-01-01
Effects which can be removed by a medikit are no longer removed by revival protocol

### LINKED ISSUE #1435: Revival Protocol cures the same statuses that Medikit does, if able
According to localization, Revival Protocol only removes mental impairments. What it doesn't tell you is that it also cures those statuses that a Medikit would. However, being afflicted by those statuses doesn't make a unit an eligible target for Revival Protocol. Therefore, there is a three-way conflict between the localization, targeting conditions, and removed effects.

Suggested fix is to align with localization, since that's the best guess of Firaxis' intent.

Related, if not directly applicable code from LWOTC here: https://github.com/long-war-2/lwotc/pull/1728

BlackDog86:
Fixes #1435


Iridar:
As much as this fix makes sense, backwards compatibility policy prevents us from altering the existing mod-facing functions. I suggest making a new `RemoveAdditionalEffectsForRevivalProtocolAndRestorativeMist_CH()` function and calling it instead where relevant. This would fix Revival Protocol without affecting any other abilities that might be using the original `RemoveAdditionalEffectsForRevivalProtocolAndRestorativeMist()`, whether that change would be beneficial or not.

BlackDog86:
Yeah, I had considered making it work that way initially. I think the original function is only called twice in the specialist ability set anyway so it's a straightforward alteration.

BlackDog86:
Adjusted the commit with a new function for BC. Actually works out better because restoration can continue to use the old function which does heal those things, and we can just adjust this for revival protocol.

PR #1440, opened 2025-01-01
Restoration Targeting Fix

### LINKED ISSUE #1436: Restoration can't be activated if the squad is not missing HP
Restoration states the following:
> The GREMLIN flies to each squad member, healing or reviving them as needed.

However, when computing whether the ability can be activated, the ability iterates through squad members. If it finds that all of them are in full health, the ability is unusable. In effect, if no one is hurt, Restoration can't be used as a comically oversized Revival Protocol.

Best also check what the interaction with any of the damaging statuses on squad members is and fix that if needed.

Originally posted to LWOTC: https://github.com/long-war-2/lwotc/issues/1727

BlackDog86:
Fixes #1436

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_Restoration.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_Restoration.uc
@@ -1,0 +1,45 @@
+class X2Condition_Restoration extends X2Condition;
+
+// Begin Issue #1436 
+// This function adjusts the targetting conditions for the restoration ability. If the unit either:
+// 1. Requires Healing
+// 2. Has a mental status effect
+// 3. Has an effect which can normally be removed by a medikit
+// They are now valid targets for the ability.
+event name CallMeetsCondition(XComGameState_BaseObject kTarget)
+{
+	local XComGameState_Unit TargetUnit;
+	local name HealType;
+
+	TargetUnit = XComGameState_Unit(kTarget);
+	if (TargetUnit == none)
+	{
+		return 'AA_NotAUnit';
+	}
+	
+	if (TargetUnit.IsBeingCarried() || TargetUnit.IsDead() || TargetUnit.IsBleedingOut())
+	{
+		return 'AA_UnitIsImmune';
+	}
+	
+	if (TargetUnit.GetCurrentStat(eStat_HP) < TargetUnit.GetMaxStat(eStat_HP))
+	{
+		return 'AA_Success';
+	}
+	
+	if (TargetUnit.IsPanicked() || TargetUnit.IsUnconscious() || TargetUnit.IsDisoriented() || TargetUnit.IsDazed() || TargetUnit.IsStunned())
+	{
+		return 'AA_Success';
+	}
+	
+	foreach class'X2Ability_DefaultAbilitySet'.default.MedikitHealEffectTypes(HealType)
+	{
+		if (TargetUnit.IsUnitAffectedByDamageType(HealType))
+		{
+			return 'AA_Success';
+		}
+	}
+
+	return 'AA_UnitIsImmune';
+}
+// End Issue #1436
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_SpecialistAbilitySet.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_SpecialistAbilitySet.uc
@@ -1179,8 +1179,10 @@ static function X2AbilityTemplate RestorativeMist()
 	local X2AbilityTemplate                 Template;
 	local X2AbilityCost_ActionPoints        ActionPointCost;
 	local X2Condition_UnitProperty          HealTargetCondition;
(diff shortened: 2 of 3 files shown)


BlackDog86:
Needs further stress testing but seemed to work fine in my preliminary round

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_Restoration.uc
```diff
@@ -0,0 +1,58 @@
+class X2Condition_Restoration extends X2Condition;
+
```
Iridar, line 2: Could use some comments with issue tags here to explain why this exists and where it is used.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_Restoration.uc
```diff
@@ -0,0 +1,58 @@
...
+	{
+		return 'AA_UnitIsImmune';
+	}
+	
+	if (TargetUnit.GetCurrentStat(eStat_HP) < TargetUnit.GetMaxStat(eStat_HP))
+	{
+		return 'AA_Success';
+	}
+	
+	if(TargetUnit.IsPanicked() || TargetUnit.IsUnconscious() || TargetUnit.IsDisoriented() || TargetUnit.IsDazed() || TargetUnit.IsStunned())
```
Iridar, line 24: missing empty space after `if`

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_Restoration.uc
```diff
@@ -0,0 +1,58 @@
...
+	local XComGameState_Unit TargetUnit;
+	local name HealType;
+
+	TargetUnit = XComGameState_Unit(kTarget);
+	if (TargetUnit == none)
+	{
+		return 'AA_NotAUnit';
+	}
+	
+	if (!TargetUnit.GetMyTemplate().bCanBeRevived || TargetUnit.IsBeingCarried() || TargetUnit.IsDead() || TargetUnit.IsBleedingOut())
```
Iridar, line 14: This will prevent non-revivable units from getting healed. The `bCanBeRevived` check should be elsewhere. 

I'm also not a fan of checking a property on the return of a getter function. If the getter fails to return the template object for whatever reason - granted, if it's happening there are worse issues than Revival not working - this will always evaluate to false.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_Restoration.uc
```diff
@@ -0,0 +1,58 @@
...
+	}
+	
+	if(TargetUnit.IsPanicked() || TargetUnit.IsUnconscious() || TargetUnit.IsDisoriented() || TargetUnit.IsDazed() || TargetUnit.IsStunned())
+	{
+		return 'AA_Success';
+	}
+	
+	foreach class'X2Ability_DefaultAbilitySet'.default.MedikitHealEffectTypes(HealType)
+	{
+		if(TargetUnit.IsUnitAffectedByDamageType(HealType))
```
Iridar, line 31: missing empty space after `if`

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_Restoration.uc
```diff
@@ -0,0 +1,58 @@
...
+		{
+			return 'AA_Success';
+		}
+	}
+
+	return 'AA_UnitIsImmune';
+}
+
+event name CallMeetsConditionWithSource(XComGameState_BaseObject kTarget, XComGameState_BaseObject kSource)
```
Iridar, line 40: This should be unnecessary - `RequireSquadmates` should already be handling this.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_Restoration.uc
```diff
@@ -0,0 +1,58 @@
...
+{
+	local XComGameState_Unit SourceUnit, TargetUnit;
+
+	SourceUnit = XComGameState_Unit(kSource);
```
Iridar, line 44: Anal-level optimization, but it would be more optimal to get and none-check each unit state separately.

Iridar (changes requested):
Needs fixing.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Condition_Restoration.uc
BlackDog86: I think this check can be removed - I started out using the basic targetting conditions from Revival Protocol, of which this is one. The base-game ability template prevents this being used on robotic units anyway, which probably accounts for 95% of cases where bCanBeRevived is set on the character template.

Iridar:
Looks good, though I wish it was possible to fix this without replacing X2Conditions on the ability. I'd like to get some testing done on this before merging, but otherwise this is good to go.

Tedster59:
Tested and confirmed working

PR #1442, opened 2025-01-02
Add shooter effect exclusions to Andromedon Acid Blob

### LINKED ISSUE #1434: Andromedon's Acid Blob doesn't have shooter effect exclusions
Andromedon's Acid Blob is strangely missing shooter effect exclusions, so it can be used e.g. while disoriented. Seems like a Firaxis oversight since they do have the primary to fire if the Blob is unavailable.

Should be a single line fix that can be yoinked from here: https://github.com/long-war-2/lwotc/pull/1669

BlackDog86:
Fixes #1434


Iridar:
This could do with Bugfixes HLDocs, otherwise good to go.

BlackDog86:
Docs added

PR #1443, opened 2025-01-02
Add check for Tick context to eHit_Crit and eHit-Graze

### LINKED ISSUE #1396: Damage over time effects applied by an attack that grazed are also halved in damage
When using Dragon Rounds or other applied on hit effects their damage gets also halved by a dodge for all ticks of the effects because they remember the hit result being `eHit_Graze`. This can end up with them doing no damage at all.

`X2Effect_ApplyWeaponDamage::CalculateDamageAmount()`
```cpp
else if (ApplyEffectParameters.AbilityResultContext.HitResult == eHit_Graze)
{
	WeaponDamage *= GRAZE_DMG_MULT;
	`log("GRAZE! Adjusted damage:" @ WeaponDamage, true, 'XCom_HitRolls');
}
```

BlackDog86:
Fixes #1396 

Still needs a lot of testing before I'd be fully confident to merge.


Iridar:
Looks good to me, if testing shows this works, this is okay to merge.

BlackDog86:
I've tested this with dragon rounds and a couple of units with setStatonclosestunit eStat_dodge 100 to force a graze & the damage is definitely no longer adjusted by the graze but I've not tried it in a modded / campaign setting yet.

Iridar:
Ideally waiting on some more testing, but generally this is good to go.

PR #1444, opened 2025-01-04
Add Ignore Base Damage check to Whiplash

### LINKED ISSUE #1386: Whiplash can do more damage than intended if assigned to a weapon
Skirmisher Whiplash has a damage value specified in code but not bIgnoreBaseDamage. This is not a problem in base game because it is not assigned to a weapon slot in Skirmisher class data. But mods often assign it to a weapon slot when they assume it's a ripjack ability and this causes it to additionally do the weapon's damage in both damage effects on it.

```cpp
WeaponDamageEffect = new class'X2Effect_ApplyWeaponDamage';
WeaponDamageEffect.EffectDamageValue = default.WHIPLASH_BASEDAMAGE;
Template.AddTargetEffect(WeaponDamageEffect);

// Jwats: Double damage for robotic
WeaponDamageEffect = new class'X2Effect_ApplyWeaponDamage';
WeaponDamageEffect.bIgnoreArmor = true;		//	armor was applied the first time, don't let it be applied the second time as well -jbouscher
WeaponDamageEffect.EffectDamageValue = default.WHIPLASH_BASEDAMAGE;
UnitPropertyCondition = new class'X2Condition_UnitProperty';
UnitPropertyCondition.ExcludeRobotic = false;
UnitPropertyCondition.ExcludeOrganic = true;
WeaponDamageEffect.TargetConditions.AddItem(UnitPropertyCondition);
Template.AddTargetEffect(WeaponDamageEffect);
```

Suggested fix: add `bIgnoreBaseDamage = true` to both of these damage effects.

BlackDog86:
Fixes #1386

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_SkirmisherAbilitySet.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_SkirmisherAbilitySet.uc
@@ -828,6 +828,13 @@ static function X2AbilityTemplate Whiplash()
 	Template.AbilityToHitCalc = StandardAim;
 
 	WeaponDamageEffect = new class'X2Effect_ApplyWeaponDamage';
+	/// HL-Docs: ref:Bugfixes; issue:1386
+	/// Whiplash is not, by default, attached to a weapon slot. However, many modded abilities assume that it is supposed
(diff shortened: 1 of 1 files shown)


Iridar:
I'm sorry for the wasted effort, but I strongly feel this would violate BC policy, and generally it's not in Highlander's MO to add guard rails against some mods shooting themselves in the foot, which in itself has a risk of shooting other mods in the foot. Sorry, but I am compelled by the BC policy to decline this fix.

PR #1445, opened 2025-01-04
Check effect damage is non-zero before applying rupture to the damage preview

### LINKED ISSUE #1394: Damage preview shows rupture damage on all abilities
Since all abilities without their own `DamagePreviewFn` fall back to `XComGameState_Ability::NormalDamagePreview()` it ends up showing rupture damage on abilities that are incapable of dealing damage. This includes Reload and such self targeted abilities on a ruptured unit.

BlackDog86:
Fixes #1394

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Ability.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Ability.uc
@@ -1230,13 +1230,14 @@ function NormalDamagePreview(StateObjectReference TargetRef, out WeaponDamageVal
 			MaxDamagePreview.Shred  += MaxDamagePreview.Shred * BurstFire.NumExtraShots;
 		}
 	}
+
+	// Begin Issue #1394
 	/// HL-Docs: ref:Bugfixes; issue:1394
 	/// Abilities which do not specify a custom damage preview function will show rupture damage on the damage
 	/// preview, even if the ability is not capable of doing any damage (e.g. self target abilities like reload). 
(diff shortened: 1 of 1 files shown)


BlackDog86:
Needs further testing - one potential issue would be with abilities which deal 0 damage but are somehow 'intended' to cause the rupture damage to trigger. I can't think of anything off the top of my head which does this but it ought to be considered & obviously willing to accept revisions to the methodology to check other things as well.

faanlez:
Rupture isn't applied anyway if there is no damage being dealt.
```cpp
//	Start Issue #1299
/// HL-Docs: ref:Bugfixes; issue:1299
/// Remove the cap from the amount of bonus damage that can be added to an attack by rupture, and do not add rupture added by this attack to the attack's damage.
//RuptureAmount = min(kTarget.GetRupturedValue() + NewRupture, RuptureCap);
RuptureAmount = kTarget.GetRupturedValue();

// While Rupture Cap is removed, we still want to add bonus damage from rupture only if the attack deals damage,
// as that was part of of the original functionality of the Rupture Cap.
if (WeaponDamage > 0)
{
	if (RuptureAmount != 0)
	{
		WeaponDamage += RuptureAmount;
		`log("Target is ruptured, increases damage by" @ RuptureAmount $", new damage:" @ WeaponDamage, true, 'XCom_HitRolls');
	}
}
// End Issue #1299
```

But, it would be more correct to check both min and max separately instead of adding them together for the check. So it would need to be each one checked and displayed separately because that would be how it actually works as well, if it is possible to roll zero damage then there is zero rupture applied.

Iridar:
Please put the Begin Issue tag before the HL-Docs entry in the future, as such is the convention.

PR #1446, opened 2025-01-05
Fix PCS stacking behaviour

### LINKED ISSUE #1352: PCS without stats do not stack

BlackDog86:
Fixes #1352


Iridar:
Looks good, I assume this was tested? A screenshot before/after would be ideal.

BlackDog86:
Will do, yeah I tested it a while ago but that was before refactoring the code into 2 PRs so I'll do it again to be 100% certain and post here before merge.

BlackDog86:
Here's an example of spawning in a boatload of copies (in groups of 3) of a PexM PCS with two stat boosts - notice that items with the same two stats stack correctly, but if either of the two stats are different they don't:


Here is a screenshot of the behaviour with some PCSes from the LW perk pack with no stat boosts (Hyper reactive pupils and smart macrophages):

BlackDog86:
I don't have a 'before' because my modded save will probably break horrendously but basically the PexM ones would stack and the LW perk pack ones wouldn't, both of which are equally wrong!

Iridar:
Great job. This been actually bothering me a lot in LWOTC.

PR #1448, opened 2025-01-06
Fix for erroneous burst-fire effect stacking

### LINKED ISSUE #1380: Burst Fire abilities that apply effects that can stack erroneouly multiply stacks by burst shots.
That is, if you have a three shot burst that's supposed to apply 1 stack of an effect per hit, it will instead apply 3 stacks per hit for a total of 9 stacks.

The good news is that this should be easily fixed: just remove the totally unnecessary code that does it from OnRefresh is XCGS_Effect.

BlackDog86:
Fixes #1380 

Tested with Knife rounds from MZ Perk pack (which applies 1 stack of bleeding @ 1 damage per instance) combined with Fan Fire and Cyclic Fire from the LW2 Gunner class pack. Correct amount of stacks seem to be applied - further testing required but looks good so far.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Effect.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Effect.uc
@@ -266,19 +266,28 @@ function PostCreateInit(EffectAppliedData InApplyEffectParameters, GameRuleState
 	if (EffectTemplate.bStackOnRefresh)
 	{
 		iStacks = 1;
-		AbilityContext = XComGameStateContext_Ability(NewGameState.GetContext());
-		if (AbilityContext != none && AbilityContext.InputContext.PrimaryTarget.ObjectID == ApplyEffectParameters.TargetStateObjectRef.ObjectID)
-		{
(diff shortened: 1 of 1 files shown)


Iridar:
Looks fine, but please use per-line comments to comment out code, the `//` rather than `/*`, it's less messy. It would also be useful to mention that fix specifically is being accomplished just by commenting out the game's original code that does unnecessary special handling.

X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_Effect.uc
```diff
@@ -393,19 +402,23 @@ function OnRefresh(EffectAppliedData NewApplyEffectParameters, XComGameState New
...
-			if (AbilityTemplate != none)
-			{
-				BurstFire = X2AbilityMultiTarget_BurstFire(AbilityTemplate.AbilityMultiTargetStyle);
-				if (BurstFire != none)
-				{
-					iStacks += BurstFire.NumExtraShots;
-				}
-			}
-		}
+		// Start Issue #1380 - Since onRefresh is be called for each burst-fire shot, we don't need to adjust the
```
Iridar, line 405: Thar be a typo / skipped words in the docs. Two, actually.

Iridar (commented):
Other than that, good to go.

Tedster59:
Has this been tested yet?

Tedster59 (approved):
Looks good to me

PR #1449, opened 2025-01-06
Fix for Enemy Protocol granting additional hacking bonuses equal to temporary hacking stat boosts

### LINKED ISSUE #1350: Enemy Protocol Hack Reward also grants a permanent hack bonus equal to current temporary Hack stat modifiers
This appears to be caused by the reward calculating what to set the soldier's base stat to by using GetMaxStat.

presumably it could be fixed by changing GetMaxStat to GetBaseStat, but i haven't tested that to see if it results in any other wierdness.

vanilla modifiers (gremlin, skullmining) work around this by being built into the aim calculation rather than being actual stat mods, so this is primarily an issue with modded bonuses.

BlackDog86:
Fixes #1350


X2WOTCCommunityHighlander/Src/XComGame/Classes/X2HackReward.uc
```diff
@@ -297,8 +297,11 @@ function ApplyEnemyProtocol(XComGameState_Unit Hacker, XComGameState_BaseObject
 	local XComGameState_Unit HackerState;
 
 	HackerState = XComGameState_Unit(NewGameState.ModifyStateObject(class'XComGameState_Unit', Hacker.ObjectID));
-
-	HackerState.SetBaseMaxStat(eStat_Hacking, HackerState.GetMaxStat(eStat_Hacking) + ENEMY_PROTOCOL_HACKING_BONUS);
+	// Single Line for Issue #1443 - Use GetBaseStat instead of GetMaxStat
```
Iridar, line 300: In cases where you provide HL-Docs, it's better to use Start / End Issue tags, so it goes like this:

// Start Issue X
/// HL-Docs: blah
/// blah blah
<actual code>
// End Issue X

Iridar (commented):
Other than this looks good, needs testing in modded environment.

PR #1451, opened 2025-01-07
Make the AI Condition bTestTargetEffectsApply work properly and allow it to be used with grenades

### LINKED ISSUE #1369: Allow mods to adjust AI behaviour of AoE profiles to specifically exclude lost units

BlackDog86:
Fixes #1369 

This is an alternative approach to the original issue of allowing mods to specifically exclude the lost from AOE targetting profiles and centres around some findings from testing, wherein XGAIBehavior::GetAllAoETargets was incorrecly adding a bunch of units to its out<targetstates> array, which should have been filtered out when bTestTargetEffectsApply was set on the AOE profile in XCOMAI.ini. Furthermore, the same function was also not capable of checking the area of effect profiles of grenades, as it was only set up to look directly at multitargeteffects on the ability templates (which for launchgrenade and throwgrenade, don't exist).

It could be argued that this is a bugfix - strictly speaking, but basically the only abilities in the base game which use this part of the function are Harbor Wave and the Purifier Flamethrower and the only base-game effect the oversight would've had, is that purifiers would probably try to fire at things even if they have fire immunity (the assasin's harbor wave has a unique damage type which I don't think anything is immune to anyway).

With the fix in place, enemy & AI modders can simply specify bTestTargetEffectsApply on their AOE Profiles in the AI templates and that should be sufficient to prevent those abilities being used against units which are immune to the effects (whether those are created via multi-target effects, or by making additional grenades / rockets).

### DIFF
--- a/Components/DLC2CommunityHighlander/DLC2CommunityHighlander/Src/DLC_2/Classes/CHDLC2Version.uc
+++ b/Components/DLC2CommunityHighlander/DLC2CommunityHighlander/Src/DLC_2/Classes/CHDLC2Version.uc
@@ -10,7 +10,7 @@ var string Commit;
 defaultproperties
 {
 	MajorVersion = 1;
-	MinorVersion = 29;
+	MinorVersion = 30;
 	PatchVersion = 0;
 	Commit = "";
 }
--- a/VERSION.ps1
+++ b/VERSION.ps1
@@ -5,7 +5,7 @@ $version_block = @'
 defaultproperties
 {
 	MajorVersion = 1;
-	MinorVersion = 29;
+	MinorVersion = 30;
 	PatchVersion = 0;
 	Commit = "%COMMIT%";
 }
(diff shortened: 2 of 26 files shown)


BlackDog86:
Also for posterity:
image

PR #1454, opened 2025-01-14
Config gate creation of auto-generated photos (#1453)

### LINKED ISSUE #1453: Allow mods to easily disable automatic photobooth photos
Several places in the base-game code create automatic photobooth posters. Many players prefer to take their own photobooth shots and would prefer that the game doesn't automatically create these under different circumstances. Also the auto-photos seem to be mainly of horrible quality compared with user-taken shots, though I've not traced the codepath, perhaps the default resolution for user-generated photos is different but anyway, mostly they look like garbage. 

Preliminary investigation reveals the following culprits:

AUTO-DEAD:
UIAfterAction::LoadPhotoboothAutoGenDeadSoldiers 
(Called from InitScreen - Takes automatic photos of dead soldiers above rank 3)
AUTO-PROMO:
XComHQPresentationLayer::UIArmory_Promotion
(Takes automatic photos of soldiers when promoting to Rank 3 or Rank 6)
AUTO-BONDED:
UISoldierBondConfirmScreen::KickOffAutoGen
(Weirdly, automatically generates a photo and then asks the user if they want to create their own)
AUTO-MISSION:
UIMissionSummary::CloseScreenTakePhoto
(Automatically takes a mission photo with the full squad in it, if the user didn't take one themselves)

May I suggest gating all of these checks behind a config flag so they can be turned off easily without having to MCO each of these classes (In particularly XComHQPresentationLayer is probably not a great one to need MCOing).

An alternative approach may be to adjust X2Photobooth_StrategyAutoGen and X2Photobooth_TacticalAutoGen which are ultimately called by these functions but it's probably better to gate the calls at source rather than adjut the functions themselves which has a higher chance of messing with other base-game behaviour that may call / create these classes (they are used for soldier headshots and quite possibly various other modded things)

BlackDog86:
Fixes #1453 - Preliminary tests are convincing. 

Further testing ongoing - will comment again once completed and confident the fix covers all possible scenarios (In particular, need to test soldier captured events, as I didn't adjust anything related to that, assuming that the auto-gen code used for captures is mainly there for creating the poster in the UI alert which is displayed when a capture occurs and not for actually saving the photo to the campaign photobooth folder).

### DIFF
--- a/X2WOTCCommunityHighlander/Config/XComGame.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGame.ini
@@ -297,3 +297,10 @@ iMixedCharacterPoolChance = 50
 
 ;;; Issue #1398 - Change to adjust the delay of the 'Squad member dead' voiceline after a unit is killed
 fSquadMemberDeadVoicelineDelay = 3.0f
+
+;;; HL-Docs: ref:DisableAutoPhotos
+; Uncomment to disable automatic photobooth poster generation when certain events happen
+;bDisableAutomaticMissionPhoto = true
+;bDisableAutomaticMemorialPhoto = true
+;bDisableAutomaticPromotionPhoto = true
+;bDisableAutomaticBondPhoto = true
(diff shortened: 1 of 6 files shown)


Tedster59 (approved):
This looks good to me

PR #1462, opened 2025-05-17
Fix powered ragdoll not working after previous ragdoll

### LINKED ISSUE #1461: Units don't play animations that use a powered ragdoll after being ragdolled once
When you want to play an animation that eventually would ragdoll, like a death animation, it's actually ragdolling along the animation.
`XComUnitPawn` has a variable `fPhysicsMotorForce` which is used to power that ragdoll, and that variable will be eventually ticked down to 0 when the unit is ragdolling.

The issue is that `fPhysicsMotorForce` is never set to the default value, thus if the unit happens to end their ragdoll and stand up again, like if they were knocked down or the floor was blown up under them, the next time it would use a powered ragdoll such as a result of death, it will get weak on the knees, flop to the ground and not play their intended animation.

So to fix this, in `EndRagdoll` function of `XComUnitPawn` I suggest setting `fPhysicsMotorForce` to the default value.

furudee:
Fixes #1461

To replicate the issue:
Spawn a trooper on a roof -> blow it up with a grenade -> shoot and kill it = no death animation and instant ragdoll
With this fix:
Spawn -> explode the roof -> kill it -> plays death animation

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComUnitPawn.uc
@@ -1058,6 +1058,13 @@ simulated function EndRagDoll()
 	Mesh.bUpdateJointsFromAnimation = true;
 	Mesh.PhysicsWeight = 0.0f;
 
+	/// HL-Docs: ref:Bugfixes; issue:1461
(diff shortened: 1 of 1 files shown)


Tedster59 (approved):
Tested and confirmed working

Issue #488, opened 2019-04-27
Extend X2Action_Death to allow overriding death animations


Musashi1584:
This should allow to override death animations globally without the need of subclassing X2Action_Death which only one mod can do at a time.

PR #1464, opened 2025-05-23
Add functionality for animations to be played for knockbacks

### LINKED ISSUE #1463: Feature request: allow animations to be played for knockbacks
As far as I know, currently it's impossible to have an unit play an animation when they're knocked back; they ragdoll instead.

So I suggest adding such functionality:
When an unit is knocked back, look up an appropriate animation name to play
If the unit has that animation in their animation set and can play it, use a powered ragdoll to play that animation
If not, ragdoll the unit instead like currently

In vanilla animation sets, such animations don't exist yet, so without any mods that add those animations, units will behave and ragdoll just like before. This feature just adds a possibility for mod makers to create knockback animations and have them played appropriately

furudee:
Fixes #1463 

Feature looks for an appropriate animation to play and plays it if it exists when an alive unit is knocked back
Animation lookup function `ComputeAnimationToPlay` is mostly copied from `X2Action_Death` since I figured it worked for this case too

Animation names it looks for:
`HL_MeleeKnockback`
`HL_Knockback`
`HL_KnockbackRight`
`HL_KnockbackLeft`
and variations of last three, depending on a persistant effect an unit is affected by
`HL_"$OverrideAnimEffectString$"Knockback`
`HL_"$OverrideAnimEffectString$"KnockbackRight`
`HL_"$OverrideAnimEffectString$"KnockbackLeft`
if an unit is mind controlled for example, it could look for `HL_MindControlKnockback`
 but it will default to `HL_Knockback` if such special animation doesn't exist
Also added a new variable `CustomKnockbackAnimationName` from #488 to override the knockback animation

This will do nothing in an unmodded game, since animations that the function looks for don't exist in any animsets.
I'm kinda bad at writing documentation so I hope it's sufficient

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_Knockback.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_Knockback.uc
@@ -25,6 +25,9 @@ var private Actor							DamageDealer;
 var			bool OnlyRecover;
 var private bool StayInPlace;
 
+// Variable for Issue #1463
+var name CustomKnockbackAnimationName;
+
 function Init()
 {
 	local XComGameStateHistory History;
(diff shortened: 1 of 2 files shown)


Tedster59 (approved):
Looks good to me, also got a double check from Reshi.

Issue #1465, opened 2025-06-17
Resource loot item states bloating the history


furudee:
As per https://gist.github.com/rcd47/9ef5a855d78f0388ae6416ec2e0bfb0c#bloat-problem-1-loot
the game creates tens of item states when counting the amount to be rewarded, yet never deleting the useless states after counting, thus bloating the history and savefile sizes

An easy solution to this as noted in the analysis, is to remove the indicated useless item states from the pending gamestate, so it will not be part of the history, and so will be garbage collected away (to my understanding)

See this function in `XComGameState_HeadquartersXCOM` responsible for counting the loot, note where I have added the suggested solution:
```
function bool PutItemInInventory(XComGameState AddToGameState, XComGameState_Item ItemState, optional bool bLoot = false)
{
	local bool HQModified;
	local XComGameState_Item InventoryItemState, NewInventoryItemState;
	local X2ItemTemplate ItemTemplate;

	ItemTemplate = ItemState.GetMyTemplate();

	if( ItemState.HasBeenModified() || ItemTemplate.bAlwaysUnique )
	{
		HQModified = true;

		if(bLoot)
		{
			LootRecovered.AddItem(ItemState.GetReference());
		}
		else
		{
			AddItemToHQInventory(ItemState);
		}
	}
	else
	{
		if(!ItemState.GetMyTemplate().bInfiniteItem)
		{
			if( HasUnModifiedItem(AddToGameState, ItemTemplate, InventoryItemState, bLoot, ItemState) )
			{
				HQModified = false;
				
				if(InventoryItemState.ObjectID != ItemState.ObjectID)
				{
					NewInventoryItemState = XComGameState_Item(AddToGameState.ModifyStateObject(class'XComGameState_Item', InventoryItemState.ObjectID));
					NewInventoryItemState.Quantity += ItemState.Quantity;

					///////////////// ADD THIS HERE /////////////////////////////////////
					// AddToGameState.PurgeGameStateForObjectID(ItemState.ObjectID);   //
					/////////////////////////////////////////////////////////////////////
				}
			}
			else
			{
				HQModified = true;

				if(bLoot)
				{
					LootRecovered.AddItem(ItemState.GetReference());
				}
				else
				{
					AddItemToHQInventory(ItemState);
				}
			}
		}
		else
		{
			HQModified = false;
			AddToGameState.RemoveStateObject(ItemState.ObjectID);
		}
	}

[...]

PR #1469, opened 2025-06-29
Issue #1465 - Delete item after merging quantity into HQ inventory item

rcd47:
Tested by clearing a room on the Avenger. X2DE confirms that the supply items were all removed. Played through a few missions and didn't notice any problems.


Tedster59 (approved):
Looks good to me

Issue #1466, opened 2025-06-27
XComHQ::arrGeneratedMissionData bloat part 2


Tedster59:
While #1188 Talks about part of the issue, another cause of the bloat is that "stale" missions are never cleaned up anywhere after they are no longer needed.  This causes the array to bloat with "valid" data for long gone missions, or in the case of Open Squad Select At Any Time, lots of additional "dummy missions" every time the squad select is opened.  The array can be cleaned up with a mod similar to how Tactical State Cleaner works, but ideally systems that add to this array should clean up their entries when they no longer need them.

rcd47:
After some more investigation, it turns out that delaying removal of XCGS_MissionSite isn't so easy. It happens as part of the X2MissionSourceTemplate's OnSuccess/Failure/Expire functions. So now I'm thinking we do need to populate arrGeneratedMissionData, but only temporarily. We could copy the mission data into the array in XCGSC_StrategyGameRule.ProcessMissionResults(), right before invoking the success/failure functions. Then, in XComHQPresentationLayer.ExitPostMissionSequence(), we could clear the array at the same time that XComHQ.MissionRef is cleared.

Tedster59:
I haven't done anything directly for CHL yet.  My testing involved an equivalent functionality mod to Tactical State Cleaner.  Ideally we'd have a more robust solution for CHL.

Tedster59:
You'd think, but they are missing some still since I was able to find
savings on DerAva's lategame LW saves.  I think my logging showed ~75 extra
missions or so that were missed.  CI has its issues with Squad Select at
Any Time not cleaning up its dummy missions.

PR #1470, opened 2025-06-29
Issue #1466 - Stop duplicating mission data in arrGeneratedMissionData

rcd47:
Started a new campaign with this change in place and played through a few missions. Did not notice any problems. Tested SSAAT separately and did not notice any problems with that either. I have not tested LWOTC or CI, but this approach should not cause any problems for them either.


Tedster59 (approved):
Tested with LWOTC and confirmed it works. LW experimental handling was not finding anything to remove after several test missions.

PR #1471, opened 2025-06-30
Geoscape ready checking round 2

Tedster59:
Fixes issue #1417 further with additional checks updated to the GeoscapeReadyForUpdate helper.

(This also cleans up some compiler warnings about unused variables added by some previous PRs)


Tedster59:
Good to know, I will revisit and test further

Tedster59:
Ok, found what I think might be causing the POI flickering (missed a variable), am testing that now, will dig into the end of month stuff + avatar project too

Tedster59:
Ok, I just did some further testing and the issues you identified have been resolved.  Can you test again?

rcd47:
Played through a few more missions with these changes and didn't notice any problems.

Tedster59:
Thanks for double checking!

PR #1473, opened 2025-07-04
None check unit states during bond resets

### LINKED ISSUE #1457: Dismissing soldiers doesn't remove their bond data resulting in redscreens
```
[1689.00] Warning: Redscreen: ModifyStateObject called with a specified ObjectID of '6763' for class 'XComGameState_Unit'- this is an invalid argument! 
Stack:
Script call stack:
    (X2StrategyGameRulesetDataStructures) X2StrategyGameRulesetDataStructures::ResetNotBondedSoldierCohesion
    (UISoldierBondConfirmScreen_0) UISoldierBondConfirmScreen::ConfirmClicked
    (UIButton_1049) UIButton::Click
    (UIButton_1049) UIButton::OnMouseEvent
    (UIMovie_2D_8) UIMovie::FlashRaiseMouseEvent
```

This is a redscreen from the moment 2 soldiers are bonded. ResetNotBondedSoldierCohesion tries to reset the bond data for all soldier but the 2 that are being bonded. Unit with Object ID 6763 is likely a dismissed unit whose bond data has lingered despite the unit itself not existing, producing a redscreen.

Tedster59:
Fixes a redescreen for trying to edit a UnitState that doesn't exist by adding a none-check, Fixes #1457.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StrategyGameRulesetDataStructures.uc
@@ -2986,17 +2986,33 @@ static function ResetNotBondedSoldierCohesion(XComGameState NewGameState, XComGa
 		if (BondData.Bondmate.ObjectID != BondmateB.ObjectID)
 		{
 			// Grab the Unit whose bond data needs to be reset for the bonded pair
-			UnitState = XComGameState_Unit(NewGameState.ModifyStateObject(class'XComGameState_Unit', BondData.Bondmate.ObjectID));
-			ResetAIndex = UnitState.AllSoldierBonds.Find('Bondmate', BondmateA.GetReference());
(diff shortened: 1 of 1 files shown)


furudee:
I suppose it works, only note is that why this happens is because game state only has the two soldiers being bonded and nobody else, so I am left wondering how/if the cohesion between a bonded soldier and a random grunt was ever reset

faanlez:
I guess the real bug is dead units not being removed from those arrays. Bonds are reset for captured or dismissed units but not dead ones.

Tedster59:
yeah, that's probably a better solution

faanlez:
And actually `ResetAllBonds` doesn't even remove the bond data, it just resets cohesions. So dismissing needs the same kind of fix whereas captures are fine since they still exist but just not in current Crew.

Tedster59:
> I guess the real bug is dead units not being removed from those arrays. Bonds are reset for captured or dismissed units but not dead ones.

going to close this one and redo it later to address this instead.

PR #1475, opened 2025-07-12
Issue #1325: New XCGS_Unit feature and opt in interruption bug fix

### LINKED ISSUE #1325: Start of initiative interruptions behave too much like start of turn interruptions

TacGit2:
#1325 
This is a bit of a 2 part pull request. 1. Adds a new function to XComGameState_Unit called CHL_SetupActionsForBeginTurn. The function comes with additional parameters for greater control compared to its vanilla counterpart. Useful in its own right. 2. Also adds an optional rework in X2TacticalGameRuleset.TurnPhase_UnitActions.SetupUnitActionsForGroupTurnBegin that utilizes the previous feature. By default, there shouldn't be any change to logic until the associated config values are uncommented/changed.

### DIFF
--- a/X2WOTCCommunityHighlander/Config/XComGameCore.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGameCore.ini
@@ -25,7 +25,7 @@
 ;InterruptionsCleanupBeginTurnUnitValues = false
 ;InterruptionsUpdateTurnStartLocation = false
 ;InterruptionsResetPanicTestsPerformedThisTurn = false
-;InterruptionsTriggerGroupTurnBegunEvent = false
+;InterruptionsTriggerGroupTurnBegunEvent = true
 ;End Issue #1325
 
 [XComGame.X2AbilityToHitCalc_StandardAim]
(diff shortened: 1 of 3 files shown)


TacGit2:
Latest commit alters the optional interruption fix to trigger group turn begun events.

Here's why:

PLAYER turns: Many effects tick, many values reset/get updated, etc.

GROUP turns (vanilla): Extremely niche. Only progresses stuns and groups are closely related to interruptions. Groups starting their turn will get some things reset, like untouchable and unit values.

GROUP turns (optional fix): Extremely niche. Only progresses stuns and groups are closely related to interruptions.

PR #1484, opened 2025-08-07
Pillar Cover Bug - Finally Fixed!

### LINKED ISSUE #1288: Templars' Pillar doesn't give cover when spawned directly adjacent to units

BlackDog86:
Fixes #1288 

So goodness knows how many hours have been spent on this, but finally it's done - this bug has been getting on my nerves for months so many thanks to Furudee and Tedster for helping to come up with a workable solution!

This probably needs a bit of more thorough gameplay testing, since building additional gamestates & visualizations which trigger on a deferred listener after the visualization block is complete is not conventional and I worry there may be some unusual interactions with other in-game abilities but it's a big improvement from where we were. Having an ability where its entire intended purpose is to generate cover, not generate cover properly, is pretty egregious.

I think the overall visualization of the ability is satisfactory - I decided to remove the 'step out of cover' animation since it looks quite strange & given the psionic nature of the ability you'd think that standing in front of a bin probably isn't much of a hindrance to a templar summoning a pillar of psionic force with the power of his mind!

There may be things we can do to improve the 'snap' of the pawn between the end of the ability and the update of the cover situation but honestly I think it's barely noticeable - see video on Discord for more info :)

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_TemplarAbilitySet.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Ability_TemplarAbilitySet.uc
@@ -950,7 +950,9 @@ static function X2AbilityTemplate Pillar()
 
 	Template.AbilityShooterConditions.AddItem(default.LivingShooterProperty);
 	Template.AddShooterEffectExclusions();
-
+	// Single line for Issue #1288 - Remove step out animation as it looks silly
+	Template.bSkipExitCoverWhenFiring = true;
+	
 	PillarEffect = new class'X2Effect_Pillar';
 	PillarEffect.BuildPersistentEffect(1, false, true, false, eGameRule_PlayerTurnBegin);	
 	PillarEffect.DestructibleArchetype = "FX_Templar_Pillar.Pillar_Destructible";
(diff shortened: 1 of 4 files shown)


Iridar:
If OSS works, it would be a good choice.

Tedster59:
@robojumper - do you have thoughts on this? Iridar going "IIRC this is bad" is enough for me to want additional guidance, especially since this is a longstanding bug that can be very frustrating to encounter.

robojumper:
Visualization deferrals are pretty bad because they can and will cause race conditions with latent submissions (also they don't work for AI players since AI performs many moves before visualizing them)

Tedster59:
Thanks for confirming. @BlackDog86 - we'll want to convert this to OSS somehow with some other delay mechanism to properly trigger the update.

Tedster59 (changes requested):
Given robojumper's comments on using the VisualizationBlockComplete deferral, we'll need to redo this with a different method

Tedster59:
Ok, so after a bunch of discussion and experimenting, the following were tried in various combinations:

- OnStateSubmitted deferral
- PostActivationEvents
- Flushing the visibility cache and tile data in the WorldInfo
- PostBuildVisiblity functions
- custom BuildVisiblityFns
- MergeVisibility functions
- Visualization Fences

The latest solution simply flushes and resets the visualization manager's caches using functions normally only called during multiplayer or when loading a tactical game, it does not submit any gamestates so can use the visualization deferral, and it doesn't seem to have any noticeable performance impact.

Tedster59 (approved):

PR #1487, opened 2025-08-12
X2AllowSelectAll Fixes

### LINKED ISSUE #1486: Debugging Annoyances
1. When selecting Alien units with X2AllowSelectall = true from the console, the camera doesn't pan across to the selected unit
2. When granting action points to an Alien unit, for testing purposes, the action points are granted to the unit, but the tactical controller switches back to the XCOM team's last available unit
3. KillAllAIs / KillClosestEnemyToCursor just remove the unit's state objects and not their visualizers / pawns. XPand included some updated commands in his console mod to handle the onDeath situation but it would be nice if this functionality was the default for debugging purposes.

There are probably more to come!

BlackDog86:
Fixes #1486 

Improves X2AllowSelectAllBehavior by:

1. Allowing the camera to pan to Non-XCom units being controlled by the player
2. Preventing the re-selection of the player's units when issuing commands to the non-player units
3. Preventing a softlock caused by ending the turn while X2AllowSelectAll is active
4. Forcing visibility between all units to improve the control / pathing of non XCom Units

Exclusion of civilians from the command and improved behaviour which breaks xcom concealment + scampers and alerts all aliens (breaking the patrol groups and allowing them to move normally) are also included as optional config flags.

Preliminary testing looks OK - more testing & comments on the code are appreciated.


BlackDog86:
There is still a small visualizer hang when using a unit controlled by X2AllowSelectAll fires on an XCom Unit - I couldn't find the source of the issue but if anyone can figure out where it comes from it would be great!

Tedster59:
Closing this since it got ruled as out of scope for CHL.

PR #1494, opened 2025-08-22
1453 poster config 2

BlackDog86:
Adds an additional config variable & check to disable the mission poster which is displayed on the dropship, if automatic photos are disabled and the user didn't take a manual one on the last mission.


Tedster59 (approved):
Relevant bugfix for a feature added with the beta CHL

PR #1500, opened 2025-08-25
Allow unused 'Alien Sighted' narratives to be played properly again

### LINKED ISSUE #1492: VO For Upgraded ADVENT Units Doesn't Play
XComGameState_HeadQuartersXCom::AddSeenCharacterTemplate and HasSeenCharacterTemplate functions add the character **group** to the array of seen template names, not the individual templates. 

This means that for various advent units (trooperm2/m3, lancerm2/m3 and various others), where a specific narrative moment is described but the character group has already been seen, the narrative moment which is supposed to play when these units are first sighted, doesn't.

This can be confirmed by testing in TQL/Skirmish+ and dropping one of the affected units into play. Using SetFirstSeenVODisabled false may help if using any other mods which disables narrative stuff.
```
function AddSeenCharacterTemplate(X2CharacterTemplate CharacterTemplate)
{
	SeenCharacterTemplates.AddItem(CharacterTemplate.CharacterGroupName);
}

function bool HasSeenCharacterTemplate(X2CharacterTemplate CharacterTemplate)
{
	return (SeenCharacterTemplates.Find(CharacterTemplate.CharacterGroupName) != INDEX_NONE);
}
```

Suggested fix would be to somehow gather an array of 'played narrative moments', switch these functions over to using the character templates directly and simply exclude narrative moments that have already been played.

BlackDog86:
Fixes #1492

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_RevealAIBegin.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Action_RevealAIBegin.uc
@@ -1,0 +1,792 @@
+class X2Action_RevealAIBegin extends X2Action_PlayMatinee
+	config(Camera);
+
+// How long to pause after kicking off the first sighted NM before continuing with the reveal
+var const config float FirstSightedDelay;
+
+var XComGameStateContext_RevealAI RevealContext;
+var float RevealFOWRadius;  //A radius in units for how much FOW should be revealed
+
+var private XGUnit MatineeFocusUnitVisualizer;              //Visualizer of the target of the cinematic
+var private XComGameState_Unit MatineeFocusUnitState;
+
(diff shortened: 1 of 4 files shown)


BlackDog86:
This will require some reasonably extensive testing to ensure there are no unintended consequences before merge

BlackDog86:
Confident this is good to go now :)

BlackDog86:
To summarise:

Base game:

Tedster59:
Yeah, I'm not sure exactly what to do here, since one might consider this "breaking" for other mods. @BlackDog86 on point 1 of the adjusted behavior would a "new character template" play the narrative regardless of whether it has been played before? That seems to conflict with point 4.

BlackDog86:
I'm the adjusted code, each 'first sighted' narrative moment can only be played once. So if a cut content thing happens to use that narrative moment before a modded enemy that is set up to use the same one, the modded one won't play. Obviously if the modded one shows up first, it'll play, and the cut content one won't.

PR #1504, opened 2025-08-25
Prevent Squadmate Dead Voicelines from playing when non-soldiers are killed

### LINKED ISSUE #1501: Soldiers play 'squadmate dead' voicelines when a mimic beacon is killed
Squadmate dead voicelines are handled in XGUnit::OnDeath - 

There is no check in the dead unit visualizer (in this case 'self') that the squad member is someone whose death we should care about. As a minimum, we should introduce a check something like if(self.GetVisualizedGameState.GetMyTemplateName() != 'MimicBeacon' etc. before doing SurvivingUnit.UnitSpeak.

Apparently there are also other possible scenarios where this can happen - it would be good to understand what these are so we can properly guard them.

```
private function DelaySpeechSquadMemberDead()
{
	local XGUnit	SurvivingUnit;	

	SurvivingUnit = GetSquad().GetNextGoodMember();
	
	if (SurvivingUnit != none && !IsRobotic() && !IsAlien_CheckByCharType())
	{
		SurvivingUnit.UnitSpeak( 'SquadMemberDead' );
	}
}
```

BlackDog86:
Fixes #1501 - Currently untested, will report back once testing is finished.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnit.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGUnit.uc
@@ -1858,7 +1858,10 @@ function OnDeath( class<DamageType> DamageType, XGUnit kDamageDealer )
 	local XGPlayer PlayerToNotify;	
 //	Issue #1398 - Variable no longer required here - moved to DelaySpeechSquadMemberDead() function
 //	local bool kIsRobotic;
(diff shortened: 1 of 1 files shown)


Tedster59 (commented):
@BlackDog86 - have you tested this now? you mention you were going to in the first post.

Tedster59 (approved):

Issue #1508, opened 2025-09-06
Units with non-human pawns cannot use voicepacks


Tedster59:
In `XGUnit.UnitSpeak`, the code checks if the unit is an alien based on character type. This prevents some mod-added units such as playable aliens, Allies Unknown, etc from being able to allow voicepacks since many of these units are marked as Alien on the character template to avoid issues such as Spectre Shadowbind attempting to target a non-human enemy.

<img width="1505" height="132" alt="Image" src="https://github.com/user-attachments/assets/a524a83c-4fa9-46f4-843c-0f4a14a37f94" />

This check simply checks the unit's character template via `UnitState.IsAlien`

<img width="861" height="357" alt="Image" src="https://github.com/user-attachments/assets/a2f01b84-d044-4788-940b-a8de78744302" />

Solution: add some way to override this check.  A basic example I tested was updating the return statement to `return UnitState.IsAlien() && !UnitState.IsSoldier();` so that units that are marked as soldiers can speak but not allow other units to speak. This fixed the specific case I tested but may not be as comprehensive.

PR #1512, opened 2025-09-14
Support alien character templates for XCOM soldier voicepacks

Tedster59:
Fixes issue #1508 by adding a "is not a soldier" check before blocking alien pawns from speaking.

Mods still need to handle setting, loading, and updating the voicepacks themselves (I do have a functioning prototype which works with this change). So far, it's being tested by one streamer who is playing 12 hours a day with no reported issues.


Tedster59:
Discussed with Blackdog for review, merging.

PR #1530, opened 2025-11-01
Call `TriggerOverrideEligibleStartingRegion` with self

### LINKED ISSUE #1529: `XComGameState_WorldRegion.CanBeStartingRegion` calls event with wrong region
https://discord.com/channels/165245941664710656/273238884433788928/1433919428017524859

The current iteration of `XComGameState_WorldRegion.CanBeStartingRegion` reads
```
function bool CanBeStartingRegion(XComGameState StartState)
{
    local XComGameState_WorldRegion RegionState;
    local int idx, Count;

    // Variable for issue #1303
    local bool bEligible;

    Count = 0;

    for(idx = 0; idx < LinkedRegions.Length; idx++)
    {
        RegionState = XComGameState_WorldRegion(StartState.GetGameStateForObjectID(LinkedRegions[idx].ObjectID));
        if(RegionState != none &&
            !class'XComGameState_RegionLink'.static.TriggerOverrideAllowStartingRegionLink(StartState, self, RegionState) /* Issue #774 */)
        {
            return false;
        }
        else
        {
            Count++;
        }
    }

    // Start Issue #1303
    bEligible = (Count > 1);

    return class'XComGameState_RegionLink'.static.TriggerOverrideEligibleStartingRegion(RegionState, bEligible);
    // End Issue #1303
}
```
We should be `return`ing whether the object calling the method (`self`) can be a starting region. Instead, we're triggering the event with `RegionState`, which is a leftover from the for-loop, one of `self.LinkedRegions`. The effect is that any setups where this event is being fired is flagging wrong regions eligible for starting.

Fix is simple, replace the `return` line with
```
    return class'XComGameState_RegionLink'.static.TriggerOverrideEligibleStartingRegion(self, bEligible);
```

copyrite:
Fixes #1529

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_WorldRegion.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameState_WorldRegion.uc
@@ -203,7 +203,7 @@ function bool CanBeStartingRegion(XComGameState StartState)
 	// Start Issue #1303
 	bEligible = (Count > 1);
 
-	return class'XComGameState_RegionLink'.static.TriggerOverrideEligibleStartingRegion(RegionState, bEligible);
+	return class'XComGameState_RegionLink'.static.TriggerOverrideEligibleStartingRegion(self, bEligible);
 	// End Issue #1303
 }
 


copyrite:
Ran a test where I had the following in a for-loop inside `XComGameState_HeadquartersXCom.SetUpHeadquarters`:
```
class'XComGameState_RegionLink'.static.CreateAllLinks(StartState);
class'XComGameState_RegionLink'.static.RandomizeLinks(StartState);

//Pick which region the HQ will start in
foreach StartState.IterateByClassType(class'XComGameState_WorldRegion', IterateRegion)
{
	if (default.PossibleStartingRegions.Find(IterateRegion.GetMyTemplateName()) != INDEX_NONE)
	{
		AllStartingRegions.AddItem(IterateRegion);
	}

	// Try to find an optimal starting region
	if(IterateRegion.CanBeStartingRegion(StartState))
	{
		BestStartingRegions.AddItem(IterateRegion);
	}
}
if(BestStartingRegions.Length > 0)
{
	RandomIndex = `SYNC_RAND_STATIC(BestStartingRegions.Length);
	BaseRegion = BestStartingRegions[RandomIndex];
}
else
{
	RandomIndex = `SYNC_RAND_STATIC(AllStartingRegions.Length);
	BaseRegion = AllStartingRegions[RandomIndex];
}
`Log("BaseRegion is" @ string(BaseRegion.m_TemplateName));
```
Additionally, an event listener that returned `true` only for Eastern Europe, `false` for others. Also an event listener that allowed non-continent links, but I'm not sure if that was necessary.

Without the fix, `BaseRegion` was always Eastern Europe's neighboring region; with the fix, 100% of the time Eastern Europe. I think that's more than enough confirmation that the fix is effective.

Tedster59 (approved):

Issue #1539, opened 2025-12-16
Customize DoT tick schedules


DaloLorn:
### Problem Overview

Generally, the persistent effects for DoTs such as bleeding, burning, or poison are built with a `_WatchRule` of `eGameRule_PlayerTurnBegin`, causing them to tick at the start of the unit owner's turn. This is usually okay; the only exception I've heard of so far has been Requiem Corrosion, which would be pretty painful if it ticked before you had a chance to heal it.

However, with the use of Highlander features such as #89 and #629, and mods such as TeslaRage's Ablative Vests, or Iridar's Vest & Plating and/or Armor Overhaul, these effects have the potential to guarantee an inordinate amount of medbay time for anyone affected. 

### Proposed Solution

Add new configs, similar to those in the above-linked features, to override the default tick schedule. As I assume `eGameRule_PlayerTurnEnd` is the only other legitimate option for a DoT's `_WatchRule`, Boolean flags seem sufficient:

```ini
[XComGame.X2Effect_Burning]
BURNED_TICKS_AFTER_TURN=true ; Make burn and acid DOTs tick on turn end

[XComGame.X2StatusEffects]
POISONED_TICKS_AFTER_TURN=true ; Make poison DOT tick on turn end
BLEEDING_TICKS_AFTER_TURN=true ; Make bleeding DOT tick on turn end
```

That said, this solution lacks granularity: `X2Effect_Burning` is used for almost every custom DoT in the community, and it may make sense for different damage types to have different behaviors. Simply recycling #89's design seems... inadequate, and more refinement is warranted. If one is going to the trouble of making such sweeping changes to the game's rules, OPTCing against half the modding community will be an inelegant and needlessly laborious process when a few extra lines of code could solve it at the source.

To resolve this, a list of damage types could be provided to cause the inverse behavior, e.g. radiation burns ticking on turn start when everything else ticks on turn end (or vice versa, as is more convenient to a given modder):

```ini
[XComGame.X2Effect_Burning]
BURNED_TICKS_AFTER_TURN=true             ; Make most burn and acid DOTs tick on turn end
+BURNED_TICKS_REVERSE_SCHEDULE=Radiation ; But make radiation DOT tick on turn start!

; Or if the default is to tick on turn start...

[XComGame.X2Effect_Burning]
BURNED_TICKS_AFTER_TURN=false            ; Make most burn and acid DOTs tick on turn start
[...]

Issue #1540, opened 2025-12-16
Add events required for cover DR and point-blank bonuses


DaloLorn:
### Problem Overview

This used to be a very different issue before I realized I could just add a few event hooks instead of baking the intended feature directly into the Highlander. Several events will be needed, but for now, I'll write it as a single issue and break it up later if needed.

I want to change the `ArmorMitigation` value of an `X2Effect_ApplyWeaponDamage` on the fly, with access to *at least* the following information:

- Original mitigation value
- Unmitigated damage
- Armor piercing
- Source and target units
- Ability template
- Ability input context

I also want to change a unit's cover level against another unit, as seen by `X2AbilityToHitCalc_StandardAim`, with access to *at least* the following information:

- Original cover level
- Source and target units

Finally, I probably want to add extra flyovers to `X2Action_ApplyWeaponDamageToUnit` (or at least create the illusion of doing so). This might not require any Highlander changes, but I might need some advice to pull it off without using MCOs. Either way, the flyover must have ways to access:

- Cover level before the attack was executed. (My current, heavyweight draft doesn't know how to label the DR flyover if the cover was damaged/destroyed by the attack's environmental damage.)
- The results of the abovementioned `ArmorMitigation` event handler. (Its inputs are fine too, since they can be used to recalculate!)

<details>
<summary>Original Issue</summary>

Lately, I've been playing around with Anisotropic's Cover Damage Reduction mod, configuring it to something resembling Long War Rebalance's cover ruleset. However, where LWR's cover DR is (seemingly?) cleanly integrated into the game, Anisotropic's mod is... noticeably less so. Bluntly put, it's a heavy-handed mess with a tendency to reduce damage where it shouldn't, and neglects to factor the DR into damage previews in any way.
</details>

### Proposed Solution

Events seem to be a valid approach to at least two of the three functions I require. I'm open to suggestions as to what the payloads should look like, though.

<details>
<summary>Original Solution</summary>

The core principle seems sound, but needs to properly integrate with Highlander's damage calculations. I will quickly summarize the intended behavior:

- All features are controllable by configs, and disabled by default.

- For most intents and purposes, cover DR is a dynamic source of extra armor. Armor-ignoring effects ignore cover, armor-piercing effects reduce its effects after the DR has first been calculated. (Something to consider, however: In my experience, damage previews don't account for armor when calculating the final value. Should there be another feature to make them do that? 🤔)
[...]

PR #1541, opened 2025-12-17
#1539 - Add configs to force most DoTs to tick on turn end

DaloLorn:
Implements #1539.

### Testing

I didn't think I'd be able to include testing screenshots for this one, but it occurred to me that YAF1 might base its tooltips off an effect's `WatchRule`. I was right!

✅ When `POISONED_TICKS_AFTER_TURN` is true, poison ticks on turn end:
<details>
image
</details>

✅ When `BLEEDING_TICKS_AFTER_TURN` is true, bleeding ticks on turn end:
<details>
image
</details>

✅ When `"Fire"` is added to `BURN_TYPES_TICKING_AFTER_TURN`, fire ticks on turn end:
<details>
image
</details>

✅ While `"Acid"` is absent from `BURN_TYPES_TICKING_AFTER_TURN`, acid still uses the vanilla behavior of ticking on turn start:
<details>
image
</details>

✅ When `POISONED_TICKS_AFTER_TURN` is false, poison uses the vanilla behavior of ticking on turn start:
<details>image
</details>

✅ When `BLEEDING_TICKS_AFTER_TURN` is false, bleeding uses the vanilla behavior of ticking on turn start:
<details>image
</details>

💡 These configs can be changed (to some extent?) during tactical combat! I wasn't sure if that was going to work... but a keen eye will notice both sets of poison/bleed tests were on the same map.

❓ Effects ticking on turn end will still force the camera to look at the victim on turn start? This is in addition to looking at them when they're damaged... I don't remember that being a thing before, so might've done something wrong. Something with the visualizer, maybe?

### DIFF
--- a/X2WOTCCommunityHighlander/Config/XComGameCore.ini
+++ b/X2WOTCCommunityHighlander/Config/XComGameCore.ini
@@ -1,9 +1,16 @@
 [XComGame.X2Effect_Burning]
 ;BURNED_IGNORES_SHIELDS=true ; Make burn and acid DOT ignore shields
+; Issue #1539
+; Add damage types whose burn DOTs should tick on turn end.
+; These must be the same `DamageType` passed to SetBurnDamage() for a given DOT.
+;+BURN_TYPES_TICKING_AFTER_TURN="Fire" ; Make burn DOT tick on turn end!
 
 [XComGame.X2StatusEffects]
 ;POISONED_IGNORES_SHIELDS=true ; Make poison DOT ignore shields
 ;BLEEDING_IGNORES_SHIELDS=true ; Make bleeding DOT ignore shields
+; Issue #1539
+;POISONED_TICKS_AFTER_TURN=true ; Make poison DOT tick on turn end
+;BLEEDING_TICKS_AFTER_TURN=true ; Make bleeding DOT tick on turn end
 
 [XComGame.X2TacticalGameRuleset]
 ;PlayerTurnOrder=eTeam_XCom
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Burning.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Burning.uc
@@ -14,6 +14,7 @@ class X2Effect_Burning extends X2Effect_Persistent
 
 var privatewrite name BurningEffectAddedEventName;
 var config bool BURNED_IGNORES_SHIELDS; //Issue #89
+var config array<name> BURN_TYPES_TICKING_AFTER_TURN; //Issue #1539
 
 function bool IsThisEffectBetterThanExistingEffect(const out XComGameState_Effect ExistingEffect)
 {
@@ -52,7 +53,17 @@ simulated function SetBurnDamage(int Damage, int Spread, name DamageType)
 
 	ApplyOnTick.AddItem(BurnDamage);
 	`assert( ApplyOnTick.Length == 1 );
+
+	//Begin Issue #1539
+	// This isn't how I expected to do this, but I made one critical oversight
+	// in my initial research: SetBurnDamage isn't what calls BuildPersistentEffect. :o
+	if (BURN_TYPES_TICKING_AFTER_TURN.Find(DamageType) != INDEX_NONE)
+	{
+		self.WatchRule = eGameRule_PlayerTurnEnd;
+	}
+	//End Issue #1539
 }
+
 
 simulated function X2Effect_ApplyWeaponDamage GetBurnDamage()
 {
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StatusEffects.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StatusEffects.uc
@@ -99,6 +99,7 @@ var config int POISONED_DAMAGE;
 var config int POISONED_INFECT_DISTANCE;
 var config int POISONED_INFECT_PERCENTAGE;
 var config bool POISONED_IGNORES_SHIELDS; // Issue #89
+var config bool POISONED_TICKS_AFTER_TURN; // Issue #1539
 var localized string PoisonedFriendlyName;
 var localized string PoisonedFriendlyDesc;
 var localized string PoisonedEffectAcquiredString;
@@ -217,6 +218,7 @@ var localized string BleedingEffectAcquiredString;
 var localized string BleedingEffectTickedString;
 var localized string BleedingEffectLostString;
 var config bool BLEEDING_IGNORES_SHIELDS; // Single variable for Issue #629
+var config bool BLEEDING_TICKS_AFTER_TURN; // Issue #1539
 
 var config int ULTRASONICLURE_TURNS;
 var name UltrasonicLureName;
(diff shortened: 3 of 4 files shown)


Tedster59:
@DaloLorn - Can you pull the VSCode Tasks stuff into a separate PR? I'd like to keep the game-related stuff and infrastructure-related stuff separate.

chrishayesmu:
I would second Tedster's suggestion to test what's going on with the camera using a clean slate approach. Set up the config with the game closed and don't touch it while running. That's what players/mods will be doing, and we should match it in our testing.

Terminology nitpick throughout this PR: the correct terminology as used by the game rule is "player turn end", but the configs use the phrasing "ticks after turn".  I would prefer consistency, e.g. `POISONED_TICKS_AT_TURN_END`. This is more clear about the exact triggering event of the tick, rather than nebulously being some time after the turn.

X2WOTCCommunityHighlander/Config/XComGameCore.ini
```diff
@@ -1,9 +1,16 @@
 [XComGame.X2Effect_Burning]
 ;BURNED_IGNORES_SHIELDS=true ; Make burn and acid DOT ignore shields
+; Issue #1539
+; Add damage types whose burn DOTs should tick on turn end.
+; These must be the same `DamageType` passed to SetBurnDamage() for a given DOT.
+;+BURN_TYPES_TICKING_AFTER_TURN="Fire" ; Make burn DOT tick on turn end!
```
chrishayesmu: I don't like treating burn separately from other damage types here, unless there's a mod author who has already indicated that, if their DOT effect were moved to end-of-turn, they would add its damage type to the start-of-turn list to explicitly counter it. If such fine-grained control is required, let it be a separate issue raised by whoever wants to use it. We can implement it when (and if) it comes up without any backwards compatibility issues.

X2WOTCCommunityHighlander/Config/XComGameCore.ini
```diff
@@ -1,9 +1,16 @@
...
 
 [XComGame.X2StatusEffects]
 ;POISONED_IGNORES_SHIELDS=true ; Make poison DOT ignore shields
 ;BLEEDING_IGNORES_SHIELDS=true ; Make bleeding DOT ignore shields
+; Issue #1539
+;POISONED_TICKS_AFTER_TURN=true ; Make poison DOT tick on turn end
+;BLEEDING_TICKS_AFTER_TURN=true ; Make bleeding DOT tick on turn end
```
chrishayesmu: I would add "instead of turn start" to these. With the lines above the inverse case is clear (ignores/does not ignore shields). For tick timing, it's not as obvious (after turn/before turn/during turn/some other trigger).

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StatusEffects.uc
```diff
@@ -2285,10 +2329,23 @@ static function X2Effect_Persistent CreateBleedingStatusEffect(int NumTurns, int
 	local X2Effect_ApplyWeaponDamage DamageEffect;
 	local X2Condition_UnitProperty UnitPropCondition;
 
+	//Begin Issue #1539
```
chrishayesmu: I don't know HL convention here. I'd personally leave a comment pointing to the HLDocs comment within `CreatePoisonedStatusEffect`, but that can be disregarded if it's not consistent.

chrishayesmu (changes requested):

X2WOTCCommunityHighlander/Config/XComGameCore.ini
DaloLorn: > (...) unless there's a mod author who has already indicated (...)

... There's me. 😂 

I actually got into Highlander development to support an extensive TJ-based, LWR-inspired combat overhaul I've been thinking of. (As of yesterday, I've gotten it all to a point where I can playtest it!) I'm not actually sure whether I'll release it to the public yet - partly because it wants all my configs/events to make it into the Highlander in some form - and the closest thing I've got to a formal design doc is in a [media/rambling thread](https://discord.com/channels/590853492084572170/1439240720182546463/1450417338288574536) in the LWotC Discord, but here's the gist of it:

- Install Ablative Vests and Energy Shield Overhaul, because why adopt them when I can configure them?
- Make a derivative of Ablative Vests which does the same thing for armor. (I think all the other mods that already did that were doing other things I didn't agree with or want to include?)
- Make LW Plating items give armor instead of ablative HP.
- Make Psi, Bleeding, and Poison damage ignore all defenses: armor, cover, ablative HP, and shields are all useless, all the time. (Might want to compensate by making the DoTs tick on turn end to make Vipers a little less oppressive. *[This thought is the whole reason I made this proposal!]*) 
- Make ~~half of~~ Fire, Napalm, Cold, and Electric damage bypass ablative HP and go straight to HP. *[Partial shield bypassing across an entire damage type will require either extensive OPTCing or a far more invasive event than anything I've already had to make for #1540, so I've decided not to do it right now.]*
- Make cover work basically as in LWR: 33/50% DR, no effect at melee range (or versus certain special abilities like the Akimbo's Trick Shot ability). This includes removing the aim penalty and granting the +40% flanking crit chance, essentially dynamically changing the weapon to use the same rules as Trick Shot.
- Make armor/cover apply before ablative, and can reduce damage to 0. (This is just enabling a few Highlander features, I don't need to do anything besides continuing to use the mod I already use.)
- Adjust ADVENT units to no longer be giant blobs of HP, and have a percentage of their HP pool converted to ablative HP to represent their armor. Except Avatars and Ethereals, because those can be fluffed as psionic shielding or something.
- Make Psionic Reaper/Warp Rifle/Psionic Repeater do Psi damage now? *[I settled for OPTCing the XCOM weapons to use the same `Projectile_BeamAvatar` damage type as the Avatar weapons, and then setting that damage type to bypass defenses... though it only now occurs to me that I need to add ESO configs for the same effect.]*

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2StatusEffects.uc
```diff
@@ -1102,10 +1104,52 @@ static function X2Effect_PersistentStatChange CreatePoisonedStatusEffect()
...
+	///
+	/// `POISONED_TICKS_AFTER_TURN` will apply to all instances of the poisoned effect created using 
+	/// the `X2StatusEffect::CreatePoisonedStatusEffect()` helper method.
+	///
+	/// Similarly, `BLEEDING_TICKS_AFTER_TURN` will apply to all instances of the bleeding effect
+	/// created using the `X2StatusEffect::CreateBleedingStatusEffect()` helper method.
+	///
+	/// This should cover all instances of these effects in the base game, but mods can potentially
+	/// disregard these helper methods, or explicitly override the effect's `WatchRule` property
+	/// after calling the helper.
```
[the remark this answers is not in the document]
DaloLorn: Would that be in addition to the existing paragraph? If so, I might go with something more like this:

"However, it is recommended that mods which create DOT effects either use the appropriate helper methods, or access these configs to implement similar behavior. This improves compatibility between mods and provides more consistency for the player."

If it's replacing the paragraph, I might want to object, on the grounds that someone might have a legitimate interest in knowing how to bypass this feature.

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect_Burning.uc
```diff
@@ -52,8 +53,18 @@ simulated function SetBurnDamage(int Damage, int Spread, name DamageType)
 
 	ApplyOnTick.AddItem(BurnDamage);
 	`assert( ApplyOnTick.Length == 1 );
+
+	//Begin Issue #1539
+	// This isn't how I expected to do this, but I made one critical oversight
+	// in my initial research: SetBurnDamage isn't what calls BuildPersistentEffect. :o
```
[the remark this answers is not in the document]
DaloLorn: How does this sound?
```
  // While the helpers used to build poison/bleed DOTs do call BuildPersistentEffect,
  // it appears that burns don't have an equivalent shared behavior we could hook into.
  // Fortunately, burns typically call SetBurnDamage after BuildPersistentEffect,
  // so we can set the WatchRule directly, overriding whatever it was previously set to.
```

X2WOTCCommunityHighlander/Config/XComGameCore.ini
DaloLorn: Might be moot, depending on the outcome of our conversation in #1539, but I'll keep it in mind!

DaloLorn:
... Did I not post a comment replying to Ted earlier? 😕 

Anyway, took a while to retest the camera panning because *XCOM*'s units being on fire doesn't really come up organically in the early months of a campaign. Now that it finally happened? I couldn't reproduce the issue unless the burn caused a bleedout, and at that point I'm thinking it's either intended behavior or interference from my modlist.

DaloLorn:
Closing as inappropriate: It can be done with OPTC, is not as reliable as initially expected, and *has* since been done with OPTC in my mod.

PR #1548, opened 2025-12-31
Use env dmg of weapon when hitting tilefrac actors and add event

### LINKED ISSUE #1547: Projectile events deal massive environmental damage to walls
in `X2Ability::GenerateDamageEvents` if an ability generated projectile events then it deals environmental damage to XComTileFracLevelActors and other fragile actors

Issue is that it deals a flat 20 damage to walls instead of looking up the environmental damage value of a weapon

For context, a typical wall with a strong toughness archetype has 10 health, a conventional sword deals 1 environmental damage, but if this projectile event happened to hit a wall then it deals a massive 20 damage to it, which is why you might had seen a sword slice taking out a chunk of a wall with it

Environmental damage events are also created in `X2Effect_ApplyWeaponDamage` which use the proper environmental damage value of the weapon and the value set in the effect

furudee:
Fixes #1547


Tedster59 (changes requested):
For consistency with updated CHL code standards, we should move the new event trigger to a separate helper function, even if the original event was embedded in the main parent function.

Issue #1543, opened 2025-12-23
Enemies mind-controlled to XCOM side don't show their action points


furudee:
because when their unit flag is created, they're enemies so `ishuman` variable in actionscript is set to false and it isn't adjusted again
so their action points will not show up in their unit flags

<img width="651" height="683" alt="Image" src="https://github.com/user-attachments/assets/3da31f85-0cdd-455a-b662-26e92f265809" />

<img width="474" height="460" alt="Image" src="https://github.com/user-attachments/assets/e467d0d1-5db4-4787-867c-61cb5f84c04b" />

PR #1549, opened 2026-01-01
Reset all bonds for dead soldiers post-mission

Tedster59:
Fixes issue #1457 for units that are killed by calling the same `ResetAllBonds` that is used when a soldier is captured later in the same function, and also is used when a unit is dismissed in the avenger (https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/blob/856cebd123e5010da2395a99f1e8a53c73f6c79e/X2WOTCCommunityHighlander/Src/XComGame/Classes/XComGameStateContext_HeadquartersOrder.uc#L332).

This PR also comments out some now-unused variables from issue #1543


Tedster59:
Closing because this still doesn't actually hit the root cause

PR #1551, opened 2026-01-02
Fix patrol members getting stuck to each other

### LINKED ISSUE #1550: Patrol leaders get stuck on their own members
There is a function to unblock the tiles patrol members are on when initiating patrol movement, but that group may not ever initiate patrol movement because leader can be blocked from moving by their members during AI behavior tree evaluation, so if they're stuck they're just going to skip their turn from behavior tree

furudee:
Fixes #1550

Tested 9-man pod with leader in center able to initiate patrol movement and scamper

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIBehavior.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIBehavior.uc
@@ -4729,6 +4729,9 @@ function BT_DisableGroupMove()
 function bt_status BT_FindDestination(int MoveTypeIndex, bool bRestricted=false)
 {
 	local TTile Tile;
+
+	MaybeUnblockPatrolMemberTiles(); // Issue #1550
+
 	// Check if we need to reset our 
 	if (bRestricted != m_bUseMoveRestriction)
 	{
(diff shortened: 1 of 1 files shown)


X2WOTCCommunityHighlander/Src/XComGame/Classes/XGAIBehavior.uc
```diff
@@ -9668,6 +9673,24 @@ state XComMovement extends MoveState // Only accessed via specialized behavior t
 		return m_vBTDestination;
 	}
 }
+
+/// HL-Docs: ref:bugfixes; issue:1550
+/// Patrolling or scampering units should be able to path through other members of their group now
+// Start Issue #1550
+simulated function MaybeUnblockPatrolMemberTiles()
```
Tedster59: well the job is mad about the docs, so something is wrong here. I'm guessing capitalization stuff

Tedster59: yeah, it wants a capital **Bugfixes**

furudee: fixed

Issue #1559, opened 2026-01-21
Units surrounded by a Hazard refuse to move


whakomatic:
AI units surrounded by hazardous world effects (fire, poison, acid) become unable to move and will skip their movement

whakomatic:
# AI Hazard Escape Fix

AI units surrounded by hazardous world effects (fire, poison, acid) become unable to move and will skip their movement

### Root Cause

The native pathfinding system (`X2ReachableTilesCache.UpdateTileCacheIfNeeded()`) excludes tiles hazardous to the unit from the reachable tiles cache. AI shouldn't voluntarily walk into fire so this is correct behavior, however, when a unit is **surrounded by the hazard** the path to a destination will fail.

The reason for failure is that the pathfinding system integrates a Units immunities to hazards into the resulting destinations and unit pathing. An immune unit will path out just fine, but a non immune unit will not.

### Code Flow

```
1. BT_StartGetDestinations
  └─> [NATIVE] m_kReachableTilesCache.UpdateTileCacheIfNeeded()
      └─> Excludes hazardous tiles from cache
      └─> Result: Empty cache when unit is hazards

2. MoveToPoint
  └─> foreach m_kReachableTilesCache.CoverDestinations(Position)
      └─> Zero tiles to process when cache is empty
      └─> AI has no destination options
```

## Solution Overview

The fix detects when a unit is in a hazard with an empty movement cache and enables special "hazard escape" pathfinding that bypasses the native cache filtering.

### Key Components

1. **Detection Flag** (`function NeedsHazardEscape`): Set when cache is empty AND unit is in hazardous terrain
2. **Validation Bypass**: Skip `IsTileReachable` checks (which rely on the empty cache)
3. **Alternative Pathfinding**: Use `BuildNonUnitPath` instead of cache-based pathfinding
4. **Unit Capabilities**: Build traversal list from unit's character template (respects grapple, climb, etc.)

### Change 1: Detection Flag

Detect when hazard escape mode is needed when there are no cover destinations in movement range.
If there is no cover in movement range, then standing still is not really a disadvantage.

```unrealscript
function NeedsHazardEscape return m_kUnit.m_kReachableTilesCache.CoverDestinations.Length == 0 &&
                      class'XComPath'.static.TileContainsHazard(UnitState, UnitState.TileLocation));
```

**Note**: Uses `TileContainsHazard` (terrain hazards only) not `IsInDangerousArea` (includes flanking).

---

[...]

PR #1560, opened 2026-01-21
AI Hazard Escape

whakomatic:
fix for issue https://github.com/X2CommunityCore/X2WOTCCommunityHighlander/issues/1559#issuecomment-3776448170


BlackDog86:
Hello, please squash all commits into a single commit and force-push the squashed commits to your repo. CHL submissions must always be a single commit (unless adding additional files, in which case the new file(s) are added in commit 1 and all modifications in commit 2).

Tedster59:
I think you need triple slash for the comments in the UC file, now that I approved the docs checks to run, those fail.

PR #1568, opened 2026-04-12
VSCode improvements

DaloLorn:
Fulfilling Ted's request from #1541 at long last. That just took *forever* to get around to, didn't it...

- VSCode tasks should no longer misbehave when paths to the workspace, SDK, or game include spaces.
- Added VSCode workspaces to gitignore, on the grounds that a good X2 modding workspace contains user-specific information such as SDK paths and is unfortunately not generically applicable to other devs' environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 
 # Project specific user settings
 .vscode/settings.json
+*.code-workspace
 
 # Python
 *.pyc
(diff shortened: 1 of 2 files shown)


DaloLorn:
For... reasons... the commits in this PR are also reproduced across all the other PRs I'm going to be opening today. I can fix that later if we decide we don't want them, but right now I just wanted to get it all out for review.

Issue #1566, opened 2026-04-12
Add hook to edit armor mitigation flyovers


DaloLorn:
Branching EditMitigationMessages off from #1540 for HL-Docs.

PR #1569, opened 2026-04-12
Added EditMitigationMessages

DaloLorn:
Implements #1566.

<details>
<summary>Sample Screenshots</summary>
... This is apparently the only screenshot I have of the system working as intended. On an old demo implementation where I baked basically the entirety of my mod's featureset into the Highlander. With inappropriate flyover text. 😩 
(The current code *does* work, though, I've playtested it for a whole campaign by now. ... Well, the early months of a disaster, then a still-unfinished campaign going up to FL20, but that sums up to roughly a whole campaign!)
image
</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 
 # Project specific user settings
 .vscode/settings.json
+*.code-workspace
 
 # Python
 *.pyc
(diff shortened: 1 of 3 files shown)


Tedster59:
@DaloLorn Does this PR depend on the others? want to make sure the others are merged first if needed.

DaloLorn:
Yes and no. It is almost entirely meaningless without the others, but does not have any direct code dependencies.

PR #1577, opened 2026-04-30
Addressing issue 1575 - Adding functionality to allow mods to manage what X2Effects are allowed to be applied to an Unit

### LINKED ISSUE #1575: Add a hook to govern the application of effects in a similar manner to CanAbilityHitUnit
Goal: Provide widely applicable and easy to use tool to govern whether or not effects are allowed to be applied to a unit, providing a similar result to CanAbilityHitUnit, which directly governs what abilities are allowed to apply to a unit.

Something in the vein of: EffectShouldRedirect() as declared in X2Effect_Persistent.
Maybe something along the lines of adding the code as follows to ApplyEffect in X2Effect, and adding a function bool to X2Effect that enables this functionality:

simulated final function name X2Effect::ApplyEffect(/**/)
{
    if (TargetStateObject != none)
    {
        if (TargetStateObject.IsImmuneToEffect(self, SourceStateObject, AbilityStateObject))
        {
            return 'AA_UnitIsImmune';
        }
    }
}

And in XGS_Unit
function bool IsImmuneToEffect(X2Effect Effect, XComGameState_Unit SourceUnit, XComGameState_Ability AbilityState)
{
    local XComGameState_Effect EffectState;

    History = `XCOMHISTORY;

    foreach AffectedByEffects(EffectRef)
    {
        EffectState = XComGameState_Effect(History.GetGameStateForObjectID(EffectRef.ObjectID));
        if (EffectState.GetX2Effect().ProvidesImmunityToEffect(Effect, EffectState, SourceUnit, self, AbilityState))
        {
            return true;
        }
    }

    return false;
}

and in X2Effect_Persistent
function bool ProvidesImmunityToEffect(X2Effect Effect, XComGameState_Effect EffectState, XComGameState_Unit SourceUnit, XComGameState_Unit TargetUnit, XComGameState_Ability AbilityState) { return false; }

Thanks for the help with figuring out a better implementation for this, Merist o7

ReshiMael:
- Adding boolean X2Effect_Persistent.
- Adding boolean to XComGameState_Unit that checks for the presence of an effect that flags for immunity to an effect
- Modifying the ApplyEffect function in X2Effect to check for whether or not a unit is flagged as being immune to a specific effect.

This combination of features allows mod added persistent effects to arbitrarily exclude other effects from being allowed to be to applied to a unit, without having to rely on damage typing and other current systems that are heavily limited and may not be applicable or convenient for many use cases.

**_Example Given:_** We want to specifically ban a chosen from ever being **_Frozen_**, which is **_Frost_** damage typed. We do not want this chosen to be immune to damage from a gun another mod adds that has a **_Frost_** damage typing. This would enable the damage to be dealt to our chosen while also preventing our chosen from being **_Frozen_**.

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect.uc
@@ -1,0 +1,342 @@
+//---------------------------------------------------------------------------------------
+//  FILE:    X2Effect.uc
+//  AUTHOR:  Joshua Bouscher
+//           
+//---------------------------------------------------------------------------------------
+//  Copyright (c) 2016 Firaxis Games, Inc. All rights reserved.
+//---------------------------------------------------------------------------------------
+class X2Effect extends Object
+	abstract
+	native(Core)
+	dependson(XComGameStateContext_Ability)
+	editinlinenew
+	hidecategories(Object);
+
+struct native OverrideEffectInfo
+{
+	var int MaxNumberAllowed;
+	var int EffectIdxToApplyOnMaxExceeded;
+
+	structdefaultproperties
+	{
+		MaxNumberAllowed=0
+		EffectIdxToApplyOnMaxExceeded=INDEX_NONE
+	}
+};
+
+var array<X2Condition>  TargetConditions;
+var bool                bApplyOnHit;
+var bool                bApplyOnMiss;
+var bool                bApplyToWorldOnHit;
+var bool                bApplyToWorldOnMiss;
+var bool                bUseSourcePlayerState;             // If true, the source player's reference will be saved in PlayerStateObjectRef when the effect is applied
+var int                 ApplyChance;
+var delegate<ApplyChanceCheck> ApplyChanceFn;
+var int                 MinStatContestResult;              // The AbilityResultContext (in the EffectAppliedData) StatContestResult must be >= this number for the effect to apply.
+var int                 MaxStatContestResult;              // The AbilityResultContext (in the EffectAppliedData) StatContestResult must be <= this number for the effect to apply. Ignored if value < Min.
+var OverrideEffectInfo  MultiTargetStatContestInfo;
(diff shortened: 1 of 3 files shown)


BlackDog86:
Hi Reshi - the commit history in this submission needs cleaning up. All CHL PRs (unless the rules have changed under Tedster...) need to be made in a single commit, if adjusting files that are already part of Highlander. The only case where there should be more than one commit is if you're adding files TO CHL that aren't already present, so you'll probably need to squash & force push in Git (it can be a pain, depending which Git client you use).

I think in terms of the documentation as well, you should only add this section in one location 
```
/// HL-Docs: feature:EffectImmunity; issue:1575; tags:tactical
/// Long explainer about what this mod changes
```
Usually the docs are best done in the class with the most extensive changes. The other files can just have Start / End issue tags and a short explainer without the tags. Apologies if this is long-winded but otherwise someone else will just mention it down the line!

ReshiMael:
Commits are squashed for Clarity
2 Commits are necessary
1- Adding X2Effect to the CHL
2- Modifying X2Effect, X2Effect_Persistent, and XComGameState_Unit to enable the functionality required for 1575

BlackDog86:
Fixes #1575

Tedster59 (approved):

X2WOTCCommunityHighlander/Src/XComGame/Classes/X2Effect.uc
```diff
@@ -0,0 +1,363 @@
...
+
+	if (TargetStateObject != none && TargetStateObject.bRemovedFromPlay)
+	{
+		return 'AA_NotAUnit';
+	}
+
+	//Begin Issue #1575
+	if (TargetStateObject != none)
+	{
+		if (TargetStateObject.IsImmuneToEffect(self, SourceStateObject, AbilityStateObject))
```
BlackDog86, line 103: This line requires the AbilityStateObject but it's not assigned until line 110 & it creates a compiler warning when building CHL

PR #1583, opened 2026-05-21
Issue #1582 - Add WoundRecoveryTimeModifier event

### LINKED ISSUE #1582: Add hook to modify Wound Recovery rate
in XComGameState_HeadquartersProjectHealSoldier:CalculateWorkPerHour

Currently the only way to alter it is through a custom XGS_HQProject, but that isn't great for mod<>mod compatibility.

Styrium:
fixes #1582


Tedster59:
Do we need to handle `XComGameState_HeadquartersProjectHealSpark` as well? though that's an issue for the future DLC_3 Highlander

Styrium:
I didn't even think of that.. And ofc that class overrides `CalculateWorkPerHour` too. At least a mod overriding Spark projects is less of a compatibility nightmare than the general-purpose soldier one.

PR #1590, opened 2026-06-01
Change debug-promotion key in non-release CHL build from Tab to F11

### LINKED ISSUE #578: Add config var to disable the promote-on-tab feature
For those of use who have our own forked versions of the highlander, it's frustrating to accidentally tab on the new promotion screen (particularly as the old version tabbed between soldiers) and get a promotion. Building with `-final_release` seems an unnecessary burden just to disable this feature. So let's guard it behind a config var so those that want to disable the feature can.

BlackDog86:
Fixes #578 - Since this only affects debug / default CHL builds & community promo screen prevents it as well, may as well just change the key.


Tedster59:
Seems reasonable to me.

PR #1592, opened 2026-06-03
1591 stasis sustain bypass

### LINKED ISSUE #1591: Stasis / Sustain cannot be (easily) bypassed by Custom Abilities
The basic code path for Sustain is that units with the Sustain passive on their templates, trigger "default.SustainEvent" within the PreDeathCheck function. This event triggers an eventlistener on the SustainTriggered ability in X2Ability_PsiOperativeAbilitySet, which has the following self target effects:

```
	StasisEffect = new class'X2Effect_Stasis';
	StasisEffect.BuildPersistentEffect(1, false, false, false, eGameRule_PlayerTurnBegin);
	StasisEffect.bUseSourcePlayerState = true;
	StasisEffect.bRemoveWhenTargetDies = true;          //  probably shouldn't be possible for them to die while in stasis, but just in case
	StasisEffect.SetDisplayInfo(ePerkBuff_Penalty, Template.LocFriendlyName, Template.GetMyHelpText(), Template.IconImage);
	StasisEffect.StunStartAnim = 'HL_PsiSustainStart';
	StasisEffect.bSkipFlyover = true;
	Template.AddTargetEffect(StasisEffect);

```
So in effect, the sustained unit just has X2Effect_Stasis dumped on them - it's not a separate ability in that sense.

X2Effect_Stasis 'may be a bit unique' (and in this case is very much **not** on the level).

```
function bool ProvidesDamageImmunity(XComGameState_Effect EffectState, name DamageType)
{
	return true;
}

function bool CanAbilityHitUnit(name AbilityName) 
{
	return false;
}
```

In other words, it's hard-coded such that anything firing against a stasis-ed unit will have their shots forced to miss and do 0 damage. This means it's currently impossible / extremely difficult to make an ability that can kill a sustained unit. There is also another CHL fix (Issue #1318) which prevents skullmine from being activated against units in stasis / sustain, as it always fails (presumably for the same reasons). Potentially, mods could restore that behaviour as well, if we create some kind of config array of abilities that can bypass those checks.

Setting bBypassSustainEffects on ApplyWeaponDamage just prevents the unit from activating sustain in the first place, if the shot happens to kills the unit normally. It does not allow units to 'shoot-through' sustain. I propose some modification to these stasis functions to allow mods to create abilities that can fully bypass sustain and/or stasis if needed, or some kind of refactor which treats stasis and sustain as independent cases for the purposes of assessing how abilities can (or cannot) bypass them.

BlackDog86:
This is a draft PR for review & discussion - It's untested but tentatively fixes #1591

### DIFF
--- a/Components/DLC2CommunityHighlander/DLC2CommunityHighlander/Src/DLC_2/Classes/CHDLC2Version.uc
+++ b/Components/DLC2CommunityHighlander/DLC2CommunityHighlander/Src/DLC_2/Classes/CHDLC2Version.uc
@@ -11,6 +11,6 @@ defaultproperties
 {
 	MajorVersion = 1;
 	MinorVersion = 31;
-	PatchVersion = 0;
+	PatchVersion = 1;
 	Commit = "";
 }
--- a/VERSION.ps1
+++ b/VERSION.ps1
@@ -6,7 +6,7 @@ defaultproperties
 {
 	MajorVersion = 1;
 	MinorVersion = 31;
-	PatchVersion = 0;
+	PatchVersion = 1;
 	Commit = "%COMMIT%";
 }
 '@


BlackDog86:
Still needs modifications & further testing - do not merge

Tedster59:
Marked this as draft again for now, when it's ready we can submit it again

Issue #1605, opened 2026-08-15
UIStatMarkup doesn't allow for Unit-Specific logic


Styrium:
UIStatMarkups _do_ have a `ShouldStatDisplayFn`, but that function passes no arguments. This makes it impossible to display bonuses conditional to the Unit's equipment, other abilities, etc. I believe an Event or an X2DLCI hook could be added in order to allow for more accurately-displayed information, instead of the current system, that doesn't display conditional Stat Bonuses in the armory at all and relies on the player remembering the extra stats from certain abilities.

I believe an X2DLCI hook in `XGS_Unit::GetUIStatFromAbilities()` would be the less resource-intensive way to implement this. If an Event were to be used, it would require a trigger for each ability the soldier has or, alternatively, each Listener would need to gather the Unit's abilities via `XGS_Unit::GetEarnedSoldierAbilities()` for proper filtering, as we can't pass that information within a Tuple. A X2DLCI hook would allow the soldier's abilities to be passed as an argument alongside the UnitState.

PR #1606, opened 2026-08-16
fixes Issue #1605

Styrium:
fixes issue #1605


Styrium:
I would appreciate if someone else compile tested it before merge

Styrium:
compile-tested now

Tedster59 (approved):

PR #1611, opened 2026-08-22
#1379 - Add OverrideCovertActionRewardString event

### LINKED ISSUE #1379: Add hook to allow overriding Covert Action reward string
Allow mods to alter the reward string on Covert Action once a slot is interacted with.
Act in `UICovertActionStaffSlot::UpdateData` to modify the string where `RewardState.GetRewardPreviewString()` and `RewardState.GetRewardString()` are called.

boundir:
We trigger an event when `RewardState.GetRewardString()` is called, inside `UICovertActionStaffSlot::UpdateData`:

```unrealscript
/// EventID: OverrideCovertActionRewardString,
/// EventData: [
///     in XComGameState_Reward RewardState,
///     in XComGameState_StaffSlot StaffSlotState,
///     in bool bPreview,
///     inout string RewardString
/// ],
/// EventSource: UICovertActionStaffSlot,
/// NewGameState: none
```
`RewardState`: the reward granted by filling this slot.
`StaffSlotState`: the slot the reward belongs to.
`bPreview`: `true` for the preview string (`GetRewardPreviewString`) and `false` for the completion string (`GetRewardString`)
`RewardString`: the vanilla string. Setting it to "" suppresses the slot's reward line.

Fixes #1379

### DIFF
--- a/X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActionStaffSlot.uc
+++ b/X2WOTCCommunityHighlander/Src/XComGame/Classes/UICovertActionStaffSlot.uc
@@ -75,7 +75,7 @@ function UpdateData()
 
 	if( RewardState != None && !ActionState.bCompleted)
 	{
-		Value = RewardState.GetRewardPreviewString();
+		Value = TriggerOverrideCovertActionRewardString(RewardState, StaffSlotState, true); // Issue #1379
 		if( Value != "" && RewardState.GetMyTemplateName() != 'Reward_DecreaseRisk')
 		{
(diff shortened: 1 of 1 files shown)


Tedster59:
Is there an issue documenting what this is adding and the use case?

boundir:
Yes it's quite old it's issue #1379. As for the usage I made it back then to prevent stat farming and making the player aware the selected unit has reached the limit. You can see part of it [here](BoundirTweaks/Src/BoundirTweaks/Classes/X2EventListener_CovertActionStaffSlotFilled.uc).

Tedster59:
Interesting. I know LWOTC implemented a hard cap that simply prevents the covert from being launched if the unit is at the cap, this would make that clearer to the player.

Tedster59 (approved):

PR #1616, opened 2026-09-05
Add improved FreeKillOnDamage_CH()

### LINKED ISSUE #1615: Add improved FreeKillOnDamage() method
Base game `X2Effect_Persistent::FreeKillOnDamage()` is not being passed any values related to the damage that's being applied, making it impossible to depend on the calculated damage.

Solution: add a new `FreeKillOnDamage_CH()` method that would have additional arguments, allowing for more contextual logic.

M3r1st:
Adds an improved version of `X2Effect_Persistent::FreeKillOnDamage()` that provides additional arguments related to the calculated damage value.

Resolves #1615


Tedster59 (approved):

Tedster59:
Should we be worried about any mod that overrides `X2Effect_ApplyWeaponDamage` not implementing this change? though I guess this is an other mod problem and not CHL intentionally breaking anything.

